PR #5, opened 2019-03-08
Improve instructions for first time users.

deleugpn:
I followed the readme instructions and got an error on ConfigFile line 33 saying `Cannot find Psalm config`. This PR improves the instructions for other developers that, like me, never used psalm before and is trying to run it for the first time on a Laravel project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,20 @@ A Psalm plugin for Laravel
 
 ```
 composer require --dev psalm/plugin-laravel
+```
+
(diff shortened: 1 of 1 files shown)


muglug:
Hey, so this isn't how the plugin should be used - I've added better instructions in https://github.com/psalm/laravel-psalm-plugin/commit/4e9db5048dea01e101b845520958a6a71d425319 - let me know if that's not enough.

PR #6, opened 2019-03-08
Use Laravel's standard source code folder

deleugpn:
Laravel skeleton always defines `app` as the folder to store the source code. It makes sense for the plugin that targets Laravel users to provide a configuration that matches the standard Laravel configuration.


muglug:
See discussion in #5

Issue #24, opened 2020-03-04
 Class 'Illuminate\Support\Facades\Date' not found


leighman:
vimeo/psalm: 3.9.3
psalm/plugin-laravel: 1.1.0
laravel/framework: 5.6.39

 `Illuminate\Support\Facades\Date` seems to have only been introduced in laravel 6

pristinesource:
This is an issue in laravel-ide-helper.  I have opened an issue ticket there for it. 
https://github.com/barryvdh/laravel-ide-helper/issues/911

mr-feek:
I noticed this also @pristinesource . I think we might want to ignore those exceptions...

mr-feek:
This should be fixed now :)

PR #28, opened 2020-04-08
[wip] Tests

mr-feek:
This is based off of https://github.com/psalm/psalm-plugin-laravel/pull/27 , to add a test to it

When creating that PR, I wanted to test my implementation. I wasn't sure how, considering there weren't any tests in this repo. I've never used codeception before, but I saw that the other plugins in this org are using codeception, so I went ahead and took a shot at implementing it.

Do we want to introduce Codeception to this project?

If so, the remaining step would be to add it as a github workflow I would assume


composer.json
```diff
@@ -4,8 +4,10 @@
     "type": "psalm-plugin",
     "require": {
         "barryvdh/laravel-ide-helper": "^2.6",
+        "laravel/framework": "^5.5 || ^6.0",
```
mr-feek, line 7: I was surprised that `laravel/framework` wasn't an explicit dependency, considering this is a plugin for it. How do we feel about these version constraints?

composer.json
```diff
@@ -4,8 +4,10 @@
...
         "vimeo/psalm": "^3.8.2",
-        "orchestra/testbench": "^3.5 || ^4.0 || ^5.0"
+        "orchestra/testbench": "^3.5 || ^4.0 || ^5.0",
+        "psr/event-dispatcher": "^1.0"
```
mr-feek, line 10: When trying to run tests, the existing code was throwing this class not found exception in `FakeMetaCommand.php`. I'm not sure what is relying on this

src/AbstractPlugin.php
```diff
@@ -207,6 +209,6 @@ private function getViewFactory(
      */
     protected function getEnvironmentSetUp($app): void
     {
-        // ..
+        $app['config']->set('app.key', 'AckfSECXIvnK5r28GVIWUAxmbBSjTsmF');
```
mr-feek, line 212: We need a default app key set. Security doesn't matter and it can be exposed since it's only used for testing

composer.json
mr-feek: Ah, I see this package also supports Lumen. I guess we should switch to requiring `illuminate/*` specific packages?

composer.json
mr-feek: hm, this is probably related to https://github.com/psalm/psalm-plugin-laravel/issues/24

composer.json
mr-feek: I'm going to figure out these deps in a different pr

PR #29, opened 2020-04-08
chore: run github workflow on pr

mr-feek:
I noticed that this workflow wasn't kicked off when I made a PR introducing changes. I believe this will now kickoff the workflow whenever someone makes a PR


mr-feek:
@muglug shall I fix that error in master on this PR or would you prefer a separate one?

muglug:
Thanks!

Issue #23, opened 2020-03-01
Add stubs for Collection (template/generics)


vladyslavstartsev:
Since Laravel collections are quite popular and I'm not sure that it will be typed (I mean template annotations in Laravel code; see this https://github.com/laravel/framework/pull/31486#issuecomment-586609823), I would like to suggest adding stubs for Collection.

muglug:
Could you?

vladyslavstartsev:
probably yes, but not promising

muglug:
Also amused by @voku’s ill-fated attempt to impose any sort of external constraints on the code in Laravel core.

`laravel/framework` is an untamed horse galloping across the Great Plains, ya know... it’s this magnificent beast that knows no bondage, and you want to put a saddle on it? That’s just not right, man. Real PHP artisans would understand.

mr-feek:
@vladyslavstartsev @muglug I took a shot at this #32

PR #32, opened 2020-04-11
Generic collection

mr-feek:
should close https://github.com/psalm/psalm-plugin-laravel/issues/23

based on my other prs so i can have testing...  Happy to  rebase after others get merged


mr-feek:
This was based off of the [collection stub](https://github.com/nunomaduro/larastan/blob/master/stubs/Collection.stub) @canvural authored for larastan, but has a few additions. I suspect that this plugin for psalm is going to want a lot of the same stub files. Does it make sense  to maintain duplicates in each codebase? Is it worth extracting them out into their own codebase so they can be shared?

vladyslavstartsev:
First of all, thanks for the job done.
Secondly, if will be nice to have separate repo for this, but ask for it first in larastan repo.

muglug:
I think for the moment you could keep them here, until there’s more consolidation

muglug:
Thanks!

PR #34, opened 2020-04-12
Consolidate plugin files

mr-feek:
My main goal of this was to:
- reduce complexity of plugin files (no need for inheritance)
- reduce the size of the baseline file

Everything here should be functionally the same, it is just a refactor.


psalm.xml
```diff
@@ -11,11 +11,16 @@
...
             <directory name="tests" />
         </ignoreFiles>
     </projectFiles>
 
     <issueHandlers>
         <LessSpecificReturnType errorLevel="info" />
     </issueHandlers>
+    
+    <stubs>
+        <file name="src/Stubs/LumenApplication.stubphp"/>
```
mr-feek, line 24: I decided to add a stub for the lumen application file, instead  of requiring the lumen framework as a dependency. I assume most people using this package will be using laravel and not lumen, so I think this is acceptable

src/Plugin.php
```diff
@@ -5,34 +5,227 @@
...
+    {
+        if (file_exists($applicationPath = __DIR__.'/../../../../bootstrap/app.php')) { // Applications
+            $app = require $applicationPath;
+        } elseif (file_exists($applicationPath = getcwd().'/bootstrap/app.php')) { // Local Dev
+            $app = require $applicationPath;
+        } else { // Packages
+            $app = (new static)->createApplication();
+        }
+
+        if ($app instanceof \Illuminate\Contracts\Foundation\Application) {
```
mr-feek, line 221: I find having these two conditional statements here much more readable than splitting the logic via inheritance

src/Plugin.php
```diff
@@ -5,34 +5,227 @@
...
+            $app->make(\Illuminate\Contracts\Console\Kernel::class)->bootstrap();
+        } elseif ($app instanceof \Laravel\Lumen\Application) {
+            $app->boot();
+        }
+
+        $app->register(\Barryvdh\LaravelIdeHelper\IdeHelperServiceProvider::class);
```
mr-feek, line 227: it should be safe to call `register` on all the app instances. Register exists on both the lumen application as well as the laravel application. Both seem to short circuit if `getProvider` is true, so we don't need to explicitly call that ourselves.

muglug:
Mind rebasing this?

Issue #38, opened 2020-04-13
[enhancement] Template the container


mr-feek:
we need to add support for the likes of 

- `app()` and  `resolve()` global helpers to return instances of `Application` with no args, or else proxy  the method call to `Container::make`
- `Container::make` needs support to take a classname as first parameter, and returns an instance of that classname

PR #39, opened 2020-04-13
feature: add types to container

mr-feek:
This should close https://github.com/psalm/psalm-plugin-laravel/issues/38

### DIFF
--- a/src/ReturnTypeProvider/AppReturnTypeProvider.php
+++ b/src/ReturnTypeProvider/AppReturnTypeProvider.php
@@ -1,0 +1,37 @@
+<?php declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\ReturnTypeProvider;
+
+use Psalm\CodeLocation;
+use Psalm\Context;
+use Psalm\Plugin\Hook\FunctionReturnTypeProviderInterface;
+use Psalm\StatementsSource;
+use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Union;
+
+final class AppReturnTypeProvider implements FunctionReturnTypeProviderInterface
+{
+
+    /**
+     * @return array<string>
+     */
+    public static function getFunctionIds(): array
+    {
+        return ['app', 'resolve'];
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


src/ReturnTypeProvider/AppReturnTypeProvider.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public static function getFunctionReturnType(StatementsSource $statements_source, string $function_id, array $call_args, Context $context, CodeLocation $code_location)
+    {
+        if (!$call_args) {
+            return new Union([
+                new TNamedObject(\Illuminate\Contracts\Foundation\Application::class),
+            ]);
+        }
+
+        // @todo: this should really proxy to \Illuminate\Foundation\Application::make, but i was struggling with that
```
mr-feek, line 34: I was trying to do something like 

```
return $statements_source->getCodebase()->getMethodReturnType(
            \Illuminate\Foundation\Application::class . '::make',
            $context->self,
            $call_args
        );
```

however kept running into issues around specificity.

```
| MoreSpecificReturnType      | The declared return type 'Illuminate\Routing\Redirector' for Foo::appHelperResolvesTypes is more specific than the inferred return type 'T:fn-illuminate\foundation\application::make as object' |
| LessSpecificReturnStatement | The type 'T:fn-illuminate\foundation\application::make as object' is more general than the declared return type 'Illuminate\Routing\Redirector' for Foo::appHelperResolvesTypes
```

I must be missing something here, but both should end up with the same result regardless

muglug: > this should really proxy to \Illuminate\Foundation\Application::make

You can simulate that by having Psalm analyse a synthetic call: https://github.com/vimeo/psalm/blob/2fe8717639514834485ba354546bdc901f85aa8d/src/Psalm/Internal/Analyzer/Statements/Block/ForeachAnalyzer.php#L722-L749

mr-feek: oops, this implementation  was wrong anyway. I'll make a pr fixing

Issue #37, opened 2020-04-12
[testing] [enhancement] Better test matrix


mr-feek:
We should have tests run on various versions of our dependencies.

Ie we should be testing  on laravel 5.5, laravel 6, and laravel 7

PR #40, opened 2020-04-13
fix: stub out enumerable since it doesnt exist in 5.5

mr-feek:
When i was testing with a 5.5 app of mine, I ran into

```
MissingDependency - app/Tax/TaxService.php:442:20 - Illuminate\Database\Eloquent\Collection depends on class or interface illuminate\support\enumerable that does not exist (see https://psalm.dev/157)
        $profile = $user->billingProfiles()->find($paymentDetails['billingProfileId']);
```
That is because the enumerable interface was added in laravel 6.
This stub fixes that issue, and https://github.com/psalm/psalm-plugin-laravel/issues/37 would have caught it sooner

### DIFF
--- a/src/Stubs/Enumerable.stubphp
+++ b/src/Stubs/Enumerable.stubphp
@@ -1,0 +1,11 @@
+<?php
+
+namespace Illuminate\Support;
+
+/**
+ * @template TKey
+ * @template TValue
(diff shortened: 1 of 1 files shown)


muglug:
Thanks!

PR #41, opened 2020-04-14
Fix: we should be able to call class_exists

mr-feek:
https://github.com/psalm/psalm-plugin-laravel/issues/24#issuecomment-611184388 explains this pretty well, in conjunction with my comment.

Essentially the existing implementation makes it so that nothing else executed after it can call `class_exists`, even the application source code being tested


muglug:
Mind making phpcs happy?

mr-feek:
We could get fancier with this by solving https://github.com/barryvdh/laravel-ide-helper/issues/915 via our custom class here, but I think simply making a noop is fine for now

PR #44, opened 2020-04-18
Fix tests

mr-feek:
Sorry, I was all over the place with my initial implementation, and tests were failing. I'm new to psalm :)

I believe this will get the container working for the 90% use case where a class string is passed as the parameter to `app` and `resolve` helpers. In order to fix the  other  10% use case where the likes of the string `trans` is passed to `app`,[ I need to get the proxying to `Application::make` working](https://github.com/vimeo/psalm/issues/3173).

### DIFF
--- a/tests/acceptance/Container.feature
+++ b/tests/acceptance/Container.feature
@@ -54,9 +54,6 @@ Feature: Container
     """
     <?php
       class Foo {
-        public function resolveHelperGetContainer(): \Illuminate\Contracts\Foundation\Application {
-          return resolve();
-        }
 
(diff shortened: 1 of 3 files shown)


tests/acceptance/Container.feature
```diff
@@ -54,9 +54,6 @@ Feature: Container
     """
     <?php
       class Foo {
-        public function resolveHelperGetContainer(): \Illuminate\Contracts\Foundation\Application {
-          return resolve();
```
mr-feek, line 58: this is invalid code, `resolve` is not an alias to `app`

PR #48, opened 2020-04-18
exclude testing 5.5 on php 8

mr-feek:
Master is currently failing in this build matrix. 5.5 is the LTS, but only currently for security fixes. I don't think we need to bother resolving the dependency issue. Let's see if this gets tests green


mr-feek:
So turns out this isn't a php 8 issue. It's something with laravel 5.5 and testbench 3.5 causing composer to go into a permanent loop https://github.com/composer/composer/issues/7665

PR #55, opened 2020-04-29
chore: include example screenshot

mr-feek:
Always nice to see what a tool looks like. Thoughts? Generated with https://carbon.now.sh/

image


muglug:
Sure! What if the errors also included something that this plugin helps you spot?

mr-feek:
yeah, i think that'll be a better example. I'll do that when i have some time 😄

PR #84, opened 2020-06-10
chore: actually test other php versions

mr-feek:
I wasn't actually using the php version from the matrix before 😅

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,14 +11,26 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


mr-feek:
**php 7.1 support is going to be dropped in the next minor release**

Reasoning:
- `Laravel-ide-helper` has an important fix for us tagged in `2.7.0`. This version of ide-helper requires `"php": ">=7.2",`
- php 7.1 has reached end of life https://www.php.net/supported-versions.php
- Laravel 5.8 is approaching end of life, and Laravel 6 requires `PHP >= 7.2.0`

PR #85, opened 2020-06-10
chore: test with composer 2

mr-feek:
following along with https://github.com/laravel/framework/pull/33158


muglug:
Merging because Composer v1 can't really deal effectively with this plugin

Issue #59, opened 2020-05-04
Support for path helpers


mr-feek:
Currently, the likes of `base_path()` emits an `UnresolvableInclude`.

These are all of the path helpers to add support for 

```php
/**
     * Get the base path of the Laravel installation.
     *
     * @return string
     */
    public function basePath();

    /**
     * Get the path to the bootstrap directory.
     *
     * @param  string  $path Optionally, a path to append to the bootstrap path
     * @return string
     */
    public function bootstrapPath($path = '');

    /**
     * Get the path to the application configuration files.
     *
     * @param  string  $path Optionally, a path to append to the config path
     * @return string
     */
    public function configPath($path = '');

    /**
     * Get the path to the database directory.
     *
     * @param  string  $path Optionally, a path to append to the database path
     * @return string
     */
    public function databasePath($path = '');

    /**
     * Get the path to the environment file directory.
     *
     * @return string
     */
    public function environmentPath();

    /**
     * Get the path to the resources directory.
     *
     * @param  string  $path
     * @return string
     */
    public function resourcePath($path = '');

    /**
     * Get the path to the storage directory.
     *
     * @return string
     */
    public function storagePath();
```

mr-feek:
Hey @muglug , any insight for how to inform psalm of these paths via the plugin?

muglug:
A return type provider for that method that returns a literal string should work

mr-feek:
Ah cool, i can probably just swap out the typehinted contract with the actual instance that provides these methods, thanks!

mr-feek:
This should have been resolved in #99

PR #99, opened 2020-07-19
path helpers

mr-feek:
beginning work on https://github.com/psalm/psalm-plugin-laravel/issues/59


mr-feek:
I think this might be blocked on https://github.com/vimeo/psalm/issues/3850

PR #102, opened 2020-07-25
change: resolve abstracts from container. This will allow us to resolve named aliases

### LINKED ISSUE #96: Cannot resolve the type of named aliases from the container
**Describe the bug**
https://github.com/facade/ignition/pull/292#discussion_r454428312

ala

```php
app('log)`
```

**Impacted Versions**
all

**Additional context**
Laravel-ide-helper  already determines the type for each of these aliases by resolving them from the container. We can probably reuse that logic somehow, or else apply the same technique

mr-feek:
should close #96


mr-feek:
I think the psalm errors reported are a false positive: https://github.com/vimeo/psalm/issues/3894

PR #117, opened 2020-10-21
Fix anotation for first method of BelongsToMany

### LINKED ISSUE #116: `first` method of BelongsToMany has wrong return type
**Describe the bug**
the `first` method of BelongsToMany has wrong return type. It can return model or null, but described only model. Also the method has some code in body.
And on checking I get that error:
```
INFO: RedundantConditionGivenDocblockType - app\User.php:83:16 - Docblock-defined type TRelatedModel can never contain null (see https://psalm.dev/156)
            && null !== $this->roles()->where('name', $role)->first();
```
 
**Impacted Versions**
barryvdh/laravel-ide-helper           v2.8.0       Laravel IDE Helper, generates correct PHPDocs for all Facad...    
laravel/framework                     v5.8.38      The Laravel Framework.
laravel/tinker                        v1.0.10      Powerful REPL for the Laravel framework.
matt-allan/laravel-code-style         0.5.1        Code formatting for Laravel projects
psalm/plugin-laravel                  v1.4.0       A Laravel plugin for Psalm
vimeo/psalm                           3.17.2       A static analysis tool for finding errors in PHP applications

**Additional context**
I think problem is here:
https://github.com/psalm/psalm-plugin-laravel/blob/f31947807798182e9d7e14f28459608eb034e54b/src/Stubs/BelongsToMany.stubphp#L114
It should be 
```
     * @psalm-return TRelatedModel|null
```

pro2s:
This PR fixes #116.

Added correct return anotation for the `first` method of BelongsToMany relationship. Added test. Also I removed unessesary code in stub file.

### DIFF
--- a/src/Stubs/BelongsToMany.stubphp
+++ b/src/Stubs/BelongsToMany.stubphp
@@ -111,14 +111,9 @@ class BelongsToMany extends Relation
     /**
      * @param  array   $columns
      * @return mixed
-     * @psalm-return TRelatedModel
+     * @psalm-return TRelatedModel|null
      */
-    public function first($columns = ['*'])
-    {
-        $results = $this->take(1)->get($columns);
-
-        return count($results) > 0 ? $results->first() : null;
-    }
+    public function first($columns = ['*']) { }
 
     /**
      * Execute the query and get the first result or throw an exception.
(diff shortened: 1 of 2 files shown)


mr-feek:
Thanks a lot for your contribution and test @pro2s 🤝

PR #121, opened 2020-12-06
Fix for psalm 4

mr-feek:
Hoping this should get tests green


mr-feek:
I'm going to merge this. There are still some more issues to fix, however this solves the `offsetGet` issues

Issue #119, opened 2020-11-24
Laravel 8.3+: optional() in RouteServiceProvider causes MixedPropertyFetch + MixedArgument errors


caugner:
**Describe the bug**
As of Laravel 8.3.0 ([this commit](https://github.com/laravel/laravel/commit/bec982b0a3962c8a3e1f665e987360bb8c056298)), the `RouteServiceProvider` uses the helper `optional()`, which causes the following errors:

```
ERROR: MixedPropertyFetch - app/Providers/RouteServiceProvider.php:60:45 - Cannot fetch property on mixed var  (see https://psalm.dev/034)
            return Limit::perMinute(60)->by(optional($request->user())->id ?: $request->ip());
ERROR: MixedArgument - app/Providers/RouteServiceProvider.php:60:45 - Argument 1 of Illuminate\Cache\RateLimiting\Limit::by cannot be non-empty-mixed|null|string, expecting string (see https://psalm.dev/030)
            return Limit::perMinute(60)->by(optional($request->user())->id ?: $request->ip());
ERROR: PossiblyNullArgument - app/Providers/RouteServiceProvider.php:60:45 - Argument 1 of Illuminate\Cache\RateLimiting\Limit::by cannot be null, possibly null value provided (see https://psalm.dev/078)
            return Limit::perMinute(60)->by(optional($request->user())->id ?: $request->ip());
```

**Impacted Versions**
Affects Laravel 8.3.0 and later.

**Additional context**
I guess this could be fixed by adding stubs for both the `optional()` function and the `Optional` class:

1. `optional()` returns an `Optional` object if it's second `$callback` parameter is omitted.
2. `Optional::__get()` returns `null` unless the optional value is an object.

However, I'm not sure if it is possible to declare that `Optional::__get()` returns the value of the corresponding property (magic or not) on the value.

As for the `PossiblyNullArgument`, this would probably require to update the `Limit::by()` signature to accept `string|null`, since `Request::ip()` can return `null`.

PS: A workaround is to modify `RouteServiceProvider::configureRateLimiting()` as follows:
```
    /**
     * Configure the rate limiters for the application.
     *
     * @return void
     */
    protected function configureRateLimiting()
    {
        RateLimiter::for('api', function (Request $request) {
            /** @var User|null $user */
            $user = $request->user();

            /** @var \Illuminate\Support\Optional $optional */
            $optional = optional($user);

            /** @var string $key */
            $key = $optional->id ?: $request->ip();
[...]

PR #122, opened 2020-12-07
wip on optional

mr-feek:
Taking a stab at #119 , documenting where I left off


src/Plugin.php
```diff
@@ -106,38 +108,6 @@ private function ingestFacadeStubs(
         $registration->addStubFile($cache_dir . 'stubs.stubphp');
     }
 
-    /**
-     * @param \Illuminate\Foundation\Application|\Laravel\Lumen\Application $app
-     * @param \Illuminate\View\Factory $view_factory
-     */
-    private function ingestMetaStubs(
```
mr-feek: The meta files generated by ide-helper create a stub for `optional`, which prevents us from properly making our own stub. We can just remove the usage of this for now

src/Stubs/helpers.stubphp
```diff
@@ -57,3 +57,15 @@ function abort_if($boolean, $code, $message = '', array $headers = [])
 {
 
 }
+
+/**
+ * Provide access to optional objects.
+ *
+ * @template TValue
+ * @psalm-param TValue $value
+ * @psalm-return \Illuminate\Support\Optional<TValue>
```
mr-feek: This will need to change to account for the callback in the future

mr-feek:
`AtomicMethodCallAnalyzer` is a beast of a class, but looks like the template params aren't being passed properly

src/Stubs/helpers.stubphp
caugner: I guess this should work to account for the callback:

```suggestion
 * @template TValue
 * @template TResult
 * @template TCallback of null|callable(TValue): TResult
 * @psalm-param TValue $value
 * @psalm-param TCallback $callback
 * @psalm-return (TCallback is null ? \Illuminate\Support\Optional<TValue> : (TValue is null ? null : TResult))
```

PR #124, opened 2020-12-19
Various fixes

Ignas2526:
I've been investigating how Laravel models work and why typing doesn't seem to work.

Given the following code
```
class Article extends Model {
    public function comments()
    {
        return $this->hasMany(Article::class);
    }
}

class Comment extends Model {
    /**
     * @psalm-return \Illuminate\Database\Eloquent\Relations\BelongsTo<Article>
     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
     */
    public function article()
    {
        return $this->belongsTo(Article::class);
    }
}
```

The first thing is that relations retain their class and Builder class is never extended.
```
X instanceof \Illuminate\Database\Eloquent\Builder // returns false
get_class(X) // returns Illuminate\Database\Eloquent\Relations\BelongsTo
```
Where X is:
```
$comment->article();
$comment->article()->where('id', 2);
$comment->article()->where('id', 2)->whereDate('created_at', '>=', Carbon::now());
```
The correct type hint for these 3 cases should be:
\Illuminate\Database\Eloquent\Relations\BelongsTo<Article>

While in case of working directly with a model class, it will be part of the Builder:

```
Y instanceof \Illuminate\Database\Eloquent\Builder // returns true
get_class(Y) // returns Illuminate\Database\Eloquent\Builder
```
Where Y is:
```
Comment::query();
Comment::query()->where('type', 1);
Comment::query()->where('type', 1)->whereDate('created_at', '>=', Carbon::now());
```
The correct type should be:
\IIlluminate\Database\Eloquent\Builder<Article>

For this behavior, the methods in Eloquent\Builder must `@return static` and not `@self<TModel>`.

______

The next issue I found is related to `@mixin \Illuminate\Database\Eloquent\Builder<TRelatedModel>`.
Here's psalm ticket I wrote:
https://github.com/vimeo/psalm/issues/4863

Basically only Relationship should have this mixin. It does actually align with Laravel's code. Classes like BelongsTo do not have Eloquent\Builder mixin.

______

The last issue is numerous missing methods from the Query\Builder. I don't know why, but Psalm doesn't like when I add their definitions in Query\Builder. It only works when they are in Eloquent\Builder. The same is true for the `first` defined in the BuildsQueries.

______

Here's minimal from Laravel along with type hints and tests that illustrates working solution incorporating all the changes above.
https://psalm.dev/r/73f86f5fdd

### DIFF
--- a/src/Stubs/EloquentBuilder.stubphp
+++ b/src/Stubs/EloquentBuilder.stubphp
@@ -74,6 +74,15 @@ class Builder
     public function where($column, $operator = null, $value = null, $boolean = 'and') { }
 
     /**
+     * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $operator
+     * @param  mixed  $value
+     * @param  string  $boolean
+     * @return TModel|null
+     */
+    public function firstWhere($column, $operator = null, $value = null, $boolean = 'and') { }
+
+    /**
      * @param  \Closure|array|string  $column
      * @param  mixed  $operator
      * @param  mixed  $value
@@ -170,7 +179,7 @@ class Builder
      * @param  \Closure|null  $callback
      * @return TModel|mixed
      */
-    public function firstOr($columns = ['*'], Closure $callback = null) { }
+    public function firstOr($columns = ['*'], \Closure $callback = null) { }
 
     /**
      * @param  string  $column
@@ -296,7 +305,7 @@ class Builder
      * @param  \Closure  $callback
      * @return void
      */
-    public function onDelete(Closure $callback) { }
+    public function onDelete(\Closure $callback) { }
 
     /**
      * Call the given local model scopes.
@@ -341,6 +350,16 @@ class Builder
     public function setQuery($query) { }
 
     /**
+     * @return \Illuminate\Database\Query\Builder
+     */
+    public function toBase() { }
+
+    /**
+     * @return array
+     */
+    public function getEagerLoads() { }
+
+    /**
      * @param array $eagerLoad
      * @return static
      */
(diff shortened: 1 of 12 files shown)


src/Stubs/EloquentBuilder.stubphp
```diff
@@ -353,12 +372,662 @@ class Builder
 
     /**
      * @param TModel $model
-     * @return self<TModel>
```
mr-feek, line 356: I remember I had to do this hack a while back -- glad to see the core issue in psalm has been resolved!

src/Stubs/EloquentBuilder.stubphp
```diff
@@ -353,12 +372,662 @@ class Builder
...
+     * @return static
+     */
+    public function distinct() { }
+
+    /**
+     * @param  \Closure|\Illuminate\Database\Query\Builder|string  $table
+     * @param  string|null  $as
+     * @return static
+     */
+    public function from($table, $as = null) { }
```
mr-feek, line 439: Could you help me understand why all these methods are needed? Maybe a feature test if you wouldn't mind?

ignas2526: These methods are from Query/Builder. I tried to add them to QueryBuilder.stubphp but it doesn't work. If I don't specify them here, Psalm doesn't infer correct type. I believe its some obscure bug in Psalm or in stubs + Laravel files. I can't replicate it on psalm.dev :( 

Case 1
`Comment::query()->whereDate('created_at', '>=', Carbon::now())`

Without whereDate in any of the stubphp:
`\Illuminate\Database\Eloquent\Builder<Comment>|Illuminate\Database\Query\Builder`

With whereDate defined in the **Query**Builder.stubphp; OR with whereDate defined in the **Eloquent**Builder.stubphp:
`\Illuminate\Database\Eloquent\Builder<Comment>` - ok


Case 2
`$comment->article()->whereDate('created_at', '>=', Carbon::now())`

Without whereDate in any of the stubphp; OR with whereDate defined in the **Query**Builder.stubphp:
`mixed` and also error on the >= `Argument 2 of Illuminate\Database\Eloquent\Builder::__call expects array<array-key, mixed>, string(>=) providedPsalm(https://psalm.dev/004)`.

With whereDate defined in the **Eloquent**Builder.stubphp I get:
`\Illuminate\Database\Eloquent\Relations\BelongsTo<Article>` - ok


Okay, I can add some feature tests

mr-feek: Interesting -- definitely something going wrong here. Let me know if you would like a hand in implementing a fix or diagnosing further.

ignas2526: @mr-feek I tried to run the acceptance tests, but I could not get it working. I did find the test that should fail: "Eloquent Relation Types", the "Scenario: Relationships return themselves when the proxied method is a query builder method". 
I get these 3 errors with ErrorLevel 1:
on psalm-return in the comment:
```
Could not verify return type 'Illuminate\Database\Eloquent\Relations\HasOne<Tests\Psalm\LaravelPlugin\Models\Phone>' for App\Http\Controllers\Admin\AdvancementsController::testPsalm(https://psalm.dev/047)
```
on the code in return statement:
```Could not infer a return typePsalm(https://psalm.dev/138)```
And when I hover over the ASC:
```
Argument 2 of Illuminate\Database\Eloquent\Builder::__call expects array<array-key, mixed>, string(ASC) providedPsalm(https://psalm.dev/004)
```

I can fix it only by adding orderBy definition into EloquentBuilder stub.
Are you able to repeat this?

PR #126, opened 2021-01-29
DB facade alias support

mzur:
This failing test highlights the issue that the `DB` alias for `Illuminate\Support\Facades\DB` is not properly handled by this plugin.

It could be solved with a new `src/Stubs/DBFacadeAlias.stubphp`:

```php
<?php

class DB extends Illuminate\Support\Facades\DB
{}
```

However, I wanted to hear if there is a better solution.


gander:
Any progress?

mzur:
I'm still waiting on feedback from the maintainers.

mr-feek:
Sorry but I don't use  facades so I don't have much motivation to make them work with this plugin -- however I  will happily merge any PRs that fix the issue!

mzur:
IIRC in my initial test the additional stub fixed the issue. But now somehow it doesn't. I'll take this up again when I have more time. Meanwhile if anyone is interested in this, please feel free to sugesst a solution.

mzur:
I had another look: Does anything speak against making the facade stub function static? Since it's a facade, the function should be called statically anyway.

mr-feek:
thanks @mzur !

PR #133, opened 2021-03-13
This pr is to fix the build test issues

masterfermin02:
This pr is to fix the build issues

Summary changes:
- Edit doc to return array<array-key, lowercase-string> on  src/ReturnTypeProvider/AppReturnTypeProvider.php
- Edit doc to return array<lowercase-string> on src/ReturnTypeProvider/RedirectReturnTypeProvider.php


muglug:
Thanks!

PR #135, opened 2021-03-21
Make the Model template on the Builder covariant.

twistor:
Here's an example. Should be easier than explaining. 

Here's the issue:
https://psalm.dev/r/6c65f643a1

Here's the fix:
https://psalm.dev/r/ffc832b680


twistor:
Not sure if this is a workaround of if this is the right approach.

But this works without any of the proposed changes: https://psalm.dev/r/812a3e545c

mr-feek:
Thanks @twistor

PR #147, opened 2021-06-09
[2.x] Fix tests by dropping support for Laravel 7

caugner:
Fixes the "Install dependencies" step of the tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: [7.3, 7.4]
-        illuminate_version: [^7, ^8]
+        php: [7.3, 7.4, 8.0]
+        illuminate_version: [^8]
         composer_flags:  ['', '--prefer-lowest']
 
     name: PHP ${{ matrix.php }} | Illuminate ${{ matrix.illuminate_version }} | Composer Flags ${{ matrix.composer_flags }}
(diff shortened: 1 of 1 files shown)


caugner:
@mr-feek Eventually, I may have found the root cause for the failing tests: `barryvdh/laravel-ide-helper:2.9.2` and later require `laravel/framework:^8` which requires `illuminate/support:^8`. 

This indicates that your latest patch release (which surprised me a bit) included a breaking change (removing compatibility with Laravel 7). And since the Laravel 7 tests fail, we don't get to test the Laravel 8 side of the matrix.

I would suggest to revert the change and release another patch release (or re-release the previous patch release as a new patch release) , then remove Laravel 7 from the supported versions and bump the major version.

mr-feek:
it wasn't my intention to remove support for laravel 7 -- I did not realize that change also occured in ide-helper. Let me figure out how to undo this.

caugner:
> Let me figure out how to undo this.

@mr-feek Thanks for the swift response and the new patch release, it came just in time before our (Gitlab-)Dependabot noticed the release. :tada:

My suggestion would be to branch off `1` and to bump the major in `master`.

mr-feek:
I would be super happy to just ditch ide-helper all together

mr-feek:
I'm going to close this as I want to keep laravel 7 support for now. I took your require changes and put them in https://github.com/psalm/psalm-plugin-laravel/pull/151 .

thanks for noticing that!

PR #148, opened 2021-06-10
fix(lint): run phpcs with php 7.4

caugner:
Fixes the PHPCS check by running it directly.

This means we're losing the useful check annotations, but PHPCS had been failing with the following error:

```
Fatal error: Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.0.2". You are running 7.4.12. in /github/workspace/vendor/composer/platform_check.php on line 24
```

See: https://github.com/psalm/psalm-plugin-laravel/runs/2790324947

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,6 +13,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+
       - name: Install dependencies
         run: export COMPOSER_ROOT_VERSION=dev-master && composer install --prefer-dist --no-progress --no-suggest
 


caugner:
Maybe a better solution would be to use the [`shivammathur/setup-php@v2`](https://github.com/shivammathur/setup-php#basic-setup) action `with: php-version: '7.4'` before installing the Composer dependencies:

```
- name: Setup PHP
  uses: shivammathur/setup-php@v2
  with:
    php-version: '7.4'
```

caugner:
> 
> 
> Maybe a better solution would be to use the [`shivammathur/setup-php@v2`](https://github.com/shivammathur/setup-php#basic-setup) action `with: php-version: '7.4'` before installing the Composer dependencies:
> 
> ```
> - name: Setup PHP
>   uses: shivammathur/setup-php@v2
>   with:
>     php-version: '7.4'
> ```

Indeed that worked, the phpcs-action is running fine now!

cc @mr-feek

mr-feek:
thanks!

PR #150, opened 2021-06-13
Add missing schema declaration

eithed:
This PR will:
- add missing schema declarations taken from `/Illuminate/Database/Schema/Blueprint.php`, Laravel 8

Was not able to find tests checking `if migration with given method usage is found a stub is created with appropriate annotation` - happy to add unit tests if given guidance


mr-feek:
thanks!

PR #162, opened 2021-07-01
Isolate tools with composer-bin-plugin

caugner:
TIL from @weirdan that one can isolate tools with [composer-bin-plugin](https://github.com/bamarni/composer-bin-plugin), and I noticed that `psalm/plugin-laravel` breaks `beyondcode/laravel-dump-server` due to having `orchestra/testbench` in its dependencies (which depends on `spatie/laravel-ray`), see: https://github.com/beyondcode/laravel-dump-server/issues/78

Isolating all the tools that are required for the CI checks, but not required for users of the plugin, reduces conflicts and the overall size of the dependencies. At the end of the day, this also saves CO² 🌍.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 composer.phar
 /laravel/
 /vendor/
+/vendor-bin/**/vendor/
 composer.lock
 cache/*.stubphp
 .DS_Store
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,9 @@
         }
     },
     "scripts": {
+        "bin": "echo 'bin not installed'",
+        "post-install-cmd": ["@composer bin all install --ansi"],
+        "post-update-cmd": ["@composer bin all update --ansi"],
         "check": [
             "@analyze",
             "@lint",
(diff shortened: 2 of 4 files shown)


caugner:
Argh, `orchestra/testbench` is not only required for the CI checks, so this won't help that much. 😕

mr-feek:
This is interesting!

----

`orchestra/testbench` is really only required when psalm is being run on a _package_. I think I may be open to making it a suggested dependency -- in fact that is what larastan does: https://github.com/nunomaduro/larastan/blob/master/composer.json#L32-L34

caugner:
Closing this PR in favor of #168. This PR isn't that useful, because we would need to have `orchestra/testbench` both in `composer.json/require-dev` and `vendor-bin/codeception/composer.json`, which just increases disk usage.

We could still move `phpcs` to `vendor/bin/phpcs/composer.json` (or possibly migrate to `php-cs-fixer`), but it doesn't really make sense to introduce `composer-bin-plugin` *just* for that.

PR #164, opened 2021-07-05
chore(gitignore): add laravel folder

caugner:
Running the test locally initialises a `/laravel/` folder, but that folder is not ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 composer.phar
+/laravel/
 /vendor/
 composer.lock
 cache/*.stubphp


mr-feek:
thanks!

caugner:
I realise only now that this was actually not necessary. The issue is that the `tests/laravel-test.sh` and `tests/lumen-test.sh` depend on the current working directory to be the git repository root, and git repository's dirname to be "psalm-plugin-laravel". If these conditions are fulfilled, laravel and lumen are installed outside of the repository, so the gitignore isn't necessary.

On the other hand, it would probably be better to install both libraries inside the repository (or even inside the `tests` folder), but ignore lumen as well, if that has no negative side-effects.

/cc @mr-feek

mr-feek:
sounds fine to me

Issue #165, opened 2021-07-05
Composer test fails on Windows: "Failed to parse output: /usr/bin/env: ‘php\r’: No such file or directory"


caugner:
**Is your feature request related to a problem? Please describe.**
I tried to `composer install` and run `composer test`, but this doesn't seem to be sufficient for running the tests (as it doesn't install Laravel). The README doesn't mention how to run these tests either.

**Describe the solution you'd like**
The README should describe how to run the tests locally (or more generally, how to contribute).

**Describe alternatives you've considered**
(None.)

**Additional context**
(None.)

mr-feek:
Hey! I agree there could be some more documentation around contributing. There is a lot of overhead to be able to contribute to static analyzers. 

---

Composer install and composer test should be all that is required to run tests locally. Could you describe a little bit more about what doesn't work when you do that?

caugner:
> Composer install and composer test should be all that is required to run tests locally.

Ah, you're right. This is related to #166:

```console
$ composer test
(...)

59) url:
 Test  tests/acceptance/UrlReturnType.feature:
Failed to parse output: /usr/bin/env: ‘php\r’: No such file or directory
Error:Syntax error

Scenario Steps:

 9. $I->canSeeNoErrors()
 8. $I->runPsalm()
 7. $I->haveTheFollowingCode("<?php\n  class Foo {\n    public function getUrlGenerator(): \Illuminate\Contracts\Routing\UrlGenerator {\n      return url();\n    }\n  }")
 6. $I->haveTheFollowingConfig("<?xml version="1.0"?>\n<psalm totallyTyped="true">\n  <projectFiles>\n    <directory name="."/>\n    <ignoreFiles> <directory name="../../vendor"/> </ignoreFiles>\n  </projectFiles>\n  <plugins>\n    <pluginClass class="Psalm\LaravelPlugin\Plugin"/>\n  </p...")
 5. // The global url helper will return the correct type depending on args


---------
60) url:
 Test  tests/acceptance/UrlReturnType.feature:
Failed to parse output: /usr/bin/env: ‘php\r’: No such file or directory
Error:Syntax error

Scenario Steps:

 9. $I->canSeeNoErrors()
 8. $I->runPsalm()
 7. $I->haveTheFollowingCode("<?php\n  class Foo {\n    public function getUrl(): string {\n      return url('example.com');\n    }\n  }")
[...]

PR #167, opened 2021-07-05
feature: support head() and last()

caugner:
For me, after upgrading to `psalm-plugin-laravel` 1.4.8, Psalm was no longer able to determine the return type of [`head()`](https://laravel.com/docs/8.x/helpers#method-head), so I'm adding a test case to see if it's just me (unfortunately I wasn't able to run the acceptance tests locally: #165).

**edit**: It is actually not only broken for me.

### DIFF
--- a/tests/acceptance/Helpers.feature
+++ b/tests/acceptance/Helpers.feature
@@ -31,6 +31,38 @@ Feature: helpers
     When I run Psalm
     Then I see no errors
 
+  Scenario: head and last support
+    Given I have the following code
+    """
+        /**
+         * @return false
+         */
+        function empty_head()
+        {
+            return head([]);
+        }
+
+        /**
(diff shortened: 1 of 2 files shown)


mr-feek:
Awesome, thanks a lot!

tests/acceptance/Helpers.feature
```diff
@@ -31,14 +31,45 @@ Feature: helpers
...
+         * @return false
+         */
+        function empty_last()
+        {
+            return last([]);
+        }
+
+        function non_empty_head(): int
+        {
+            return last([1, 2, 3]);
```
canvural, line 55: Shouldn't it be `head` here?

caugner: Good catch, yes indeed.

PR #168, opened 2021-07-05
Move orchestra/testbench to require-dev

caugner:
Background: https://github.com/psalm/psalm-plugin-laravel/pull/162#issuecomment-873626249

~~⚠️ Blocked by #183 (refactors `SuppressHandler`, then we can ignore `PropertyNotSetInConstructor` of `Model` classes).~~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,6 @@
         "illuminate/support": "^6.0 || ^7.0 || ^8.0",
         "illuminate/view": "^6.0 || ^7.0 || ^8.0",
         "vimeo/psalm": "^4.8.1",
-        "orchestra/testbench": "^3.8 || ^4.0 || ^5.0 || ^6.0",
         "barryvdh/laravel-ide-helper": ">=2.8.0 <2.9.2"
     },
     "license": "MIT",
@@ -58,9 +57,13 @@
         "codeception/codeception": "^4.1.6",
         "codeception/module-phpbrowser": "^1.0.0",
         "codeception/module-asserts": "^1.0.0",
+        "orchestra/testbench": "^3.8 || ^4.0 || ^5.0 || ^6.0",
         "weirdan/codeception-psalm-module": "^0.13.1",
         "squizlabs/php_codesniffer": "*",
         "slevomat/coding-standard": "^6.2"
+    },
+    "suggest": {
+        "orchestra/testbench": "Running Psalm on a Laravel package requires Testbench"
     },
     "minimum-stability": "dev"
 }
--- a/src/Fakes/FakeCreatesApplication.php
+++ b/src/Fakes/FakeCreatesApplication.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Psalm\LaravelPlugin\Fakes;
+
+use Illuminate\Foundation\Application;
+use RuntimeException;
+
+trait FakeCreatesApplication
+{
+    /**
+     * @psalm-return never
+     */
+    public function createApplication()
+    {
+        throw new RuntimeException("Please install orchestra/testbench to analyze packages.");
+    }
+
+    /**
+     * Get package bootstrapper.
+     *
+     * @param  Application  $app
+     *
+     * @return array
+     */
+    protected function getPackageBootstrappers($app)
+    {
+        return [];
+    }
+}
--- a/src/Providers/ApplicationProvider.php
+++ b/src/Providers/ApplicationProvider.php
@@ -8,9 +8,18 @@ use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Foundation\Application as LaravelApplication;
 use Laravel\Lumen\Application as LumenApplication;
 use Orchestra\Testbench\Concerns\CreatesApplication;
+use Psalm\LaravelPlugin\Fakes\FakeCreatesApplication;
+use function class_alias;
 use function file_exists;
 use function get_class;
 use function getcwd;
+use function trait_exists;
+
+if (!trait_exists(CreatesApplication::class)) {
+    class_alias(CreatesApplication::class, __NAMESPACE__ . '\CreatesApplication');
+} else {
+    class_alias(FakeCreatesApplication::class, __NAMESPACE__ . '\CreatesApplication');
+}
 
 final class ApplicationProvider
 {


mr-feek:
I'd like to do a little bit of hand holding at runtime when someone is attempting to run this plugin during development of a library when they don't yet have `orchestra/testbench` installed. I'd prefer to show a clean error message like "you need to install orchestra/testbanch in order to analyze packages" rather than having a php error being thrown for the method not existing.

Maybe a `class_exists` check [here](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Providers/ApplicationProvider.php#L52)? What do you think?

mr-feek:
> There is a new class, Composer\InstalledVersions, which is autoloaded in every project and is available at runtime. It allows you to check which packages/versions are present at runtime of your own project.

https://blog.packagist.com/composer-2-0-is-now-available/

Perhaps using this `InstalledVersions` could be nifty too, instead of `class_exists`!

caugner:
Actually, it would also be nice to move `laravel/framework` to `suggest`.

In my project, I extracted `psalm` and `psalm-plugin-laravel` to `vendor-bin/psalm` (using `composer-bin-plugin`), and unfortunately that installs all Laravel dependencies again inside `vendor-bin/psalm/vendor`.

mr-feek:
@caugner interesting thought! however we still need to say "we support version(s) X,Y,Z of laravel". Maybe that can be accomplished with composer `conflict`?

caugner:
> Maybe that can be accomplished with composer `conflict`?

Yes, absolutely, that should be possible.

mr-feek:
Feel free to PR!

caugner:
> Maybe a `class_exists` check [here](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Providers/ApplicationProvider.php#L52)? What do you think?

The problem is that `ApplicationProvider` uses a trait from Testbench:

https://github.com/psalm/psalm-plugin-laravel/blob/935357a7c898e67966c8cb8a911ff9f66d83b978/src/Providers/ApplicationProvider.php#L6-L13

So we might need to have a `CreatesApplication` trait ourself that is conditionally defined to either extend Testbench's trait (if it is available) or a dummy trait alternatively.

> Actually, it would also be nice to move laravel/framework to suggest.

Well, as a matter of fact the plugin only requires those parts of `illuminate/*` that are explicitly used, so I was mistaken there and that's already the best we can do (I guess). Maybe running Psalm in a `composer-bin-plugin` environment isn't such a good idea, after all.

> Maybe that can be accomplished with composer conflict?

So that won't be necessary.

composer.json
```diff
@@ -5,13 +5,16 @@
     "require": {
         "php": "^7.3|^8",
         "ext-simplexml": "*",
+        "illuminate/config": "^6.0 || ^7.0 || ^8.0",
```
caugner: https://github.com/psalm/psalm-plugin-laravel/blob/28b697a39aa96986733aae79415bfd776306436f/src/Providers/FacadeStubProvider.php#L6

composer.json
```diff
@@ -5,13 +5,16 @@
...
         "illuminate/container": "^6.0 || ^7.0 || ^8.0",
         "illuminate/contracts": "^6.0 || ^7.0 || ^8.0",
         "illuminate/database": "^6.0 || ^7.0 || ^8.0",
+        "illuminate/events": "^6.0 || ^7.0 || ^8.0",
```
caugner: https://github.com/psalm/psalm-plugin-laravel/blob/28b697a39aa96986733aae79415bfd776306436f/src/Providers/ViewFactoryProvider.php#L31

composer.json
```diff
@@ -5,13 +5,16 @@
...
         "illuminate/http": "^6.0 || ^7.0 || ^8.0",
+        "illuminate/routing": "^6.0 || ^7.0 || ^8.0",
```
caugner: https://github.com/psalm/psalm-plugin-laravel/blob/28b697a39aa96986733aae79415bfd776306436f/src/Handlers/Helpers/RedirectHandler.php#L6

composer.json
```diff
@@ -5,13 +5,16 @@
...
         "illuminate/support": "^6.0 || ^7.0 || ^8.0",
+        "illuminate/view": "^6.0 || ^7.0 || ^8.0",
```
caugner, line 16: https://github.com/psalm/psalm-plugin-laravel/blob/28b697a39aa96986733aae79415bfd776306436f/src/Providers/ViewFactoryProvider.php#L6-L9

caugner:
@mr-feek Can I just suppress the following error? Or do you see a way to avoid it?

```
ERROR: DuplicateClass - src/Providers/CreatesApplication.php:18:11 - Class Psalm\LaravelPlugin\Providers\CreatesApplication has already been defined in /home/runner/work/psalm-plugin-laravel/psalm-plugin-laravel/src/Providers/CreatesApplication.php (see https://psalm.dev/071)
    trait CreatesApplication
```

caugner:
> @mr-feek Can I just suppress the following error? Or do you see a way to avoid it?

I found a solution in 7f06268.

However, now it looks like not having `orchestra/testbench` as a `require` dependency, Psalm reports the following errors on Laravel:

```
ERROR: UnresolvableInclude - app/Console/Kernel.php:39:9 - Cannot resolve the given expression to a file path (see https://psalm.dev/106)
        require base_path('routes/console.php');


ERROR: PropertyNotSetInConstructor - app/Models/User.php:10:7 - Property App\Models\User::$table is not defined in constructor of App\Models\User or in any methods called in the constructor (see https://psalm.dev/074)
class User extends Authenticatable


ERROR: PropertyNotSetInConstructor - app/Models/User.php:10:7 - Property App\Models\User::$dateFormat is not defined in constructor of App\Models\User or in any methods called in the constructor (see https://psalm.dev/074)
class User extends Authenticatable


ERROR: UnresolvableInclude - app/Providers/BroadcastServiceProvider.php:19:9 - Cannot resolve the given expression to a file path (see https://psalm.dev/106)
        require base_path('routes/channels.php');
```

caugner:
> Psalm reports the following errors on Laravel:

As for the `PropertyNotSetInConstructor` errors, I'm actually surprised that Psalm didn't catch that before. It probably makes sense to suppress those errors in `Model` subclasses as well, as part of #172.

caugner:
This is a very odd side-effect of moving `orchestr/testbench` to `require-dev`: 😵 

```
ERROR: UnusedClass - app/Console/Kernel.php:8:7 - Class App\Console\Kernel is never used (see https://psalm.dev/075)
class Kernel extends ConsoleKernel


ERROR: UnresolvableInclude - app/Console/Kernel.php:39:9 - Cannot resolve the given expression to a file path (see https://psalm.dev/106)
        require base_path('routes/console.php');


ERROR: UnusedClass - app/Exceptions/Handler.php:8:7 - Class App\Exceptions\Handler is never used (see https://psalm.dev/075)
class Handler extends ExceptionHandler


ERROR: UnusedClass - app/Http/Controllers/Controller.php:10:7 - Class App\Http\Controllers\Controller is never used (see https://psalm.dev/075)
class Controller extends BaseController


ERROR: UnusedClass - app/Http/Kernel.php:7:7 - Class App\Http\Kernel is never used (see https://psalm.dev/075)
class Kernel extends HttpKernel


ERROR: PossiblyUnusedMethod - app/Http/Middleware/RedirectIfAuthenticated.php:20:21 - Cannot find any calls to method App\Http\Middleware\RedirectIfAuthenticated::handle (see https://psalm.dev/087)
    public function handle(Request $request, Closure $next, ...$guards)


ERROR: UnusedClass - app/Http/Middleware/TrustHosts.php:7:7 - Class App\Http\Middleware\TrustHosts is never used (see https://psalm.dev/075)
class TrustHosts extends Middleware


ERROR: PropertyNotSetInConstructor - app/Models/User.php:10:7 - Property App\Models\User::$table is not defined in constructor of App\Models\User or in any methods called in the constructor (see https://psalm.dev/074)
class User extends Authenticatable


ERROR: PropertyNotSetInConstructor - app/Models/User.php:10:7 - Property App\Models\User::$dateFormat is not defined in constructor of App\Models\User or in any methods called in the constructor (see https://psalm.dev/074)
class User extends Authenticatable


ERROR: UnusedClass - app/Models/User.php:10:7 - Class App\Models\User is never used (see https://psalm.dev/075)
class User extends Authenticatable


ERROR: UnusedClass - app/Providers/AppServiceProvider.php:7:7 - Class App\Providers\AppServiceProvider is never used (see https://psalm.dev/075)
class AppServiceProvider extends ServiceProvider


ERROR: UnusedClass - app/Providers/AuthServiceProvider.php:8:7 - Class App\Providers\AuthServiceProvider is never used (see https://psalm.dev/075)
class AuthServiceProvider extends ServiceProvider


ERROR: UnusedClass - app/Providers/BroadcastServiceProvider.php:8:7 - Class App\Providers\BroadcastServiceProvider is never used (see https://psalm.dev/075)
class BroadcastServiceProvider extends ServiceProvider


ERROR: UnresolvableInclude - app/Providers/BroadcastServiceProvider.php:19:9 - Cannot resolve the given expression to a file path (see https://psalm.dev/106)
        require base_path('routes/channels.php');


ERROR: UnusedClass - app/Providers/EventServiceProvider.php:10:7 - Class App\Providers\EventServiceProvider is never used (see https://psalm.dev/075)
class EventServiceProvider extends ServiceProvider
```

caugner:
Closing this PR, because I have left the company that was using this package and for which I was doing these contributions as part of my daily work. :wave: Apologies to anybody who may be disappointed, because they counted on this PR getting merged someday. :pray: 

Obviously, if anybody wants to base their contribution on this PR, feel free to do so! No attribution required.

PR #170, opened 2021-07-06
fix: Support @property on Model with imported type

caugner:
Originally reported in https://github.com/vimeo/psalm/issues/6024.

### DIFF
--- a/src/Providers/ModelStubProvider.php
+++ b/src/Providers/ModelStubProvider.php
@@ -50,7 +50,8 @@ final class ModelStubProvider implements GeneratesStubs
 
         $models_generator_command->run(
             new ArrayInput([
-                '--nowrite' => true
+                '--nowrite' => true,
+                '--reset' => true,
             ]),
             new NullOutput()
         );
--- a/tests/Models/AbstractUuidModel.php
+++ b/tests/Models/AbstractUuidModel.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Tests\Psalm\LaravelPlugin\Models;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Str;
+use Ramsey\Uuid\UuidInterface;
+
+/**
+ * @property-read UuidInterface $uuid
+ */
+abstract class AbstractUuidModel extends Model
+{
+    protected static function boot()
+    {
+        parent::boot();
+
+        static::creating(function (Model $model) {
+            $model->setAttribute('uuid', Str::uuid());
+        });
+    }
+}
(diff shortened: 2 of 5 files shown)


caugner:
@mr-feek The root of the problem is probably located in `ModelPropertyAccessorHandler::getPropertyType()`:

https://github.com/psalm/psalm-plugin-laravel/blob/7f15023e45ef086b58264641eae65403af3bda5f/src/Handlers/Eloquent/ModelPropertyAccessorHandler.php#L56-L70

caugner:
@mr-feek I have tried to debug with `PSALM_ALLOW_XDEBUG=1`, as described [here](https://github.com/psalm/codeception-psalm-module/issues/17#issuecomment-660685568), but this didn't seem to work either.

mr-feek:
@caugner yeah, it's a really frustrating experience. What I have started doing recently is cloning a fresh laravel install and adding the code I want to analyze.

Then I require this plugin via composer, however [use a symlink to load it in locally ](https://getcomposer.org/doc/05-repositories.md#path).

Then I'm able to get xdebug working. this is the command I use: 
`PSALM_ALLOW_XDEBUG=1 php -dxdebug.mode=debug -dxdebug.start_with_request=yes ./vendor/bin/psalm`

caugner:
So the reason why the `@property CarbonInterface` annotation causes trouble on the Model is that it seems to be copied directly to `vendor/psalm/plugin-laravel/src/cache/models.stubphp`, but without the import.

Now the question is whether this is a bug in `laravel-ide-helper` or in the Psalm plugin. 😕

src/Providers/ModelStubProvider.php
```diff
@@ -47,7 +47,8 @@ public static function generateStubFile(): void
 
         $models_generator_command->run(
             new \Symfony\Component\Console\Input\ArrayInput([
-                '--nowrite' => true
+                '--nowrite' => true,
+                '--reset' => true,
```
mr-feek, line 54: would you mind explaining what this is fixing/accomplishing please? Maybe an example of a bad docblock that this is fixing?

caugner: Already included in the MR: https://github.com/psalm/psalm-plugin-laravel/blob/7174fd50e758d899e04220ff6feb5a208a03d0bd/tests/Models/User.php#L13

Corresponding scenario:
https://github.com/psalm/psalm-plugin-laravel/blob/7174fd50e758d899e04220ff6feb5a208a03d0bd/tests/acceptance/EloquentModelPropertyTypes.feature#L38-L47

caugner: You'll also find an explanation in the commit message, see: ~~7174fd50e758d899e04220ff6feb5a208a03d0bd~~ 025a4476699f05133121139f42eccf4180ae7163

PR #174, opened 2021-07-06
chore: optimize + add imports

caugner:
I noticed that there are quite some unused imports, so I thought it would be a good idea to optimize them once in a while.


src/Handlers/Application/ContainerHandler.php
```diff
@@ -2,8 +2,6 @@
 
 namespace Psalm\LaravelPlugin\Handlers\Application;
 
-use Illuminate\Foundation\AliasLoader;
-use Illuminate\Support\Facades\RateLimiter;
```
caugner, line 6: @mr-feek When I noticed this unused import, I wondered if this line is/was causing #134. 🤔

mr-feek:
Thanks for the help here! I've been lazy :)

PR #178, opened 2021-07-07
Migrate from Psalm Hooks to EventHandlers

caugner:
See: https://github.com/vimeo/psalm/pull/5898


mr-feek:
This was on my long list of things to do... thanks so much for refactoring these!

PR #181, opened 2021-07-09
test(Notification): run only for Laravel 6.0.4+

caugner:
`\Illuminate\Bus\Queuable` has invalid docblock in Laravel 6.0.3 and earlier, which [was fixed in 6.0.4](https://github.com/laravel/framework/commit/35572b39059d64bd2119fbf28638654b2a28ee61).

Therefore we should ignore the new Notification acceptance test in Laravel 6.0.3 and before, using the syntax pointed out [here](https://github.com/psalm/psalm-plugin-laravel/pull/157#issuecomment-877350836).

PR #157, opened 2021-06-23
Ensure firstOrNew/Create can be called without parameters in Laravel 8+

caugner:
Fixes #156.

🟢 ~~Includes 🍒 from #181.~~ ~~Blocked by #182.~~

PR #182, opened 2021-07-09
Ensure factory() helper is gone in Laravel 8+

caugner:
~~⚠️ Includes 🍒 from #181.~~


caugner:
@mr-feek This PR is ready to be merged now.

Afterwards I will rebase #157, because that PR also contains `feature: support stubs per major version`. This PR has a smaller scope, so I think it's better to add that feature here (#182).

mr-feek:
thanks!

PR #183, opened 2021-07-13
Suppress UnusedClass/PossiblyUnusedMethod on well-known classes/methods of fresh Laravel/Lumen projects

### LINKED ISSUE #53: This plugin ought to suppress PossiblyUnusedClass/Method on common Laravel idoms
This is mainly a reference to this, but more broadly for the whole of Laravel: https://github.com/vimeo/psalm/issues/2489

List of issues to handle: https://github.com/vimeo/psalm/issues/2489#issuecomment-568410933

Currently getting 695 extra errors when I add `--find-unused-code` 😬 (on a different code base than above)

One of these days I'll put aside some time to fix this.

caugner:
Fixes #53.

### DIFF
--- a/src/Handlers/SuppressHandler.php
+++ b/src/Handlers/SuppressHandler.php
@@ -2,73 +2,108 @@
 
 namespace Psalm\LaravelPlugin\Handlers;
 
-use Illuminate\Console\Command;
-use Illuminate\Foundation\Http\FormRequest;
-use Illuminate\Notifications\Notification;
 use Psalm\Plugin\EventHandler\AfterClassLikeVisitInterface;
 use Psalm\Plugin\EventHandler\Event\AfterClassLikeVisitEvent;
+use Psalm\Storage\ClassLikeStorage;
 use Psalm\Storage\MethodStorage;
-use function array_key_exists;
+use Psalm\Storage\PropertyStorage;
+use function array_intersect;
 use function in_array;
 
 class SuppressHandler implements AfterClassLikeVisitInterface
 {
-    private const UNUSED_CLASSES = [
-        "App\Console\Kernel",
-        "App\Exceptions\Handler",
-        "App\Http\Controllers\Controller",
-        "App\Http\Kernel",
-        "App\Http\Middleware\Authenticate",
-        "App\Http\Middleware\TrustHosts",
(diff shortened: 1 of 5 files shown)


src/Handlers/SuppressHandler.php
```diff
@@ -2,39 +2,108 @@
...
+use function array_intersect;
 use function in_array;
 
 class SuppressHandler implements AfterClassLikeVisitInterface
 {
+    /**
+     * @var array<string, list<class-string>>
+     */
+    private const BY_CLASS = [
+        'UnusedClass' => [
```
mr-feek, line 19: I wonder if these are still emitted if the `bootstrap/app.php` file is analyzed? Thinking out loud..

mr-feek:
This is _awesome_! Thanks a lot for your help

src/Handlers/SuppressHandler.php
caugner: Good point, `bootstrap/app.php` *does* use `App\Console\Kernel`, `App\Exceptions\Handler` and `App\Http\Kernel`. And some of the other classes *are* used in `config/app.php`.

So maybe it is worth adding those directories to the `tests/(laravel|lumen)-test-psalm.xml` configs. 🤔 

In my project, I have configured `<projectFiles>` likes this, probably due to too many errors (and some of this could explain some of the weird errors I have experienced):

```xml
    <projectFiles>
        <directory name="."/>
        <ignoreFiles allowMissingFiles="true">
            <directory name="app/Http/Middleware"/>
            <directory name="bootstrap/cache"/>
            <directory name="cache"/>
            <directory name="config"/>
            <directory name="database"/>
            <directory name="node_modules"/>
            <directory name="routes"/>
            <directory name="storage"/>
            <directory name="tests"/>
            <directory name="vendor"/>
            <directory name="vendor-bin"/>
            <file name="bootstrap/app.php"/>
            <file name="public/index.php"/>
            <file name="server.php"/>
        </ignoreFiles>
    </projectFiles>
```

**edit**: Ideally, we would support as much files as possible, and make a recommendation in the `README`.

mr-feek: yep, didn't view that as a blocker for this PR, but something we should consider!

PR #194, opened 2021-07-29
Added missing Builder methods - count() and whereDate()

shoxy:
Also changed the return type for `DB::raw()`


mr-feek:
Thanks for the PR! Any idea why  the unit tests are failing? doesn't seem related to your changes..

```
1) factory(): can use factory helper in Laravel 6.x and 7.x
```

🤔 💭

mr-feek:
Leaving a note here, it seems that something in upstream psalm might have broken our tests. Pin psalm to "vimeo/psalm:dev-master#0317f1d"  and the tests pass

PR #198, opened 2021-09-06
Fix fake meta command

krasnikov87:


mr-feek:
@krasnikov87 you can't hardcode a native return type of `callable` and not return a callable here

PR #209, opened 2022-01-16
feat: support foreach with Support\Collection

tjmmm:
Currently, Psalm cannot infer iterated value type from `\IteratorAggregate<TKey, TValue>`.

see also: https://github.com/vimeo/psalm/issues/7187


mr-feek:
thanks!

PR #214, opened 2022-01-19
chore: test supported PHP version only

tjmmm:
This PR prevents CI failure caused by unsupported version combinations.

Note:
Test CI stil failing if `--prefer-lowest` flag used with PHP 8.1.
We need bump up `orchestra/testbench` to fix Test CI.

> Added draft support for PHP 8.1.

https://github.com/orchestral/testbench/releases/tag/v6.22.0


mr-feek:
Thanks!

PR #217, opened 2022-01-28
Fixes #184 add UserHandler

### LINKED ISSUE #184: Use config('auth.providers.user') for return type of Auth::user() and Request::user()
**Is your feature request related to a problem? Please describe.**

According to PHPDoc:

* `Illuminate\Support\Facades\Auth::user()` returns `\Illuminate\Contracts\Auth\Authenticatable|null`
* `Illuminate\Http\Request::user()` returns `mixed`

However, if `providers.users.driver` is set to `eloquent` in `config/auth.php`, they return the model configured in `providers.users.model`, which [defaults to](https://github.com/laravel/laravel/blob/c636fd0f678008caf30683833b740fb6f31d9de6/config/auth.php#L68-L72):

```php
    'providers' => [
        'users' => [
            'driver' => 'eloquent',
            'model' => App\Models\User::class,
        ],
```

**Describe the solution you'd like**
The Psalm plugin should use that config to determine (and narrow) the return type of `Auth::user()` and `$request->user()`.

**Describe alternatives you've considered**
(None.)

**Additional context**
(None.)

tm1000:
This fixes #184. I dont think you actually need to look at the config object because in Laravel 8.x documentation it's stated that:

> By default, Laravel includes an App\Models\User Eloquent model in your app/Models directory. 

https://laravel.com/docs/8.x/authentication#introduction-database-considerations

### DIFF
--- a/src/Handlers/UserHandler.php
+++ b/src/Handlers/UserHandler.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Psalm\LaravelPlugin\Handlers;
+
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
+use Psalm\Type;
+
+final class UserHandler implements MethodReturnTypeProviderInterface
+{
+    public static function getClassLikeNames(): array
+    {
+        return [
+            \Illuminate\Support\Facades\Auth::class,
+            'Auth',
+            \Illuminate\Http\Request::class
+        ];
+    }
+
+    public static function getMethodReturnType(
+        MethodReturnTypeProviderEvent $event
+    ): ?Type\Union {
+        $method_name_lowercase = $event->getMethodNameLowercase();
+        $class_name = $event->getFqClasslikeName();
+
+        if (
+            $class_name === \Illuminate\Support\Facades\Auth::class ||
+            $class_name === 'Auth'
+        ) {
+            if ($method_name_lowercase === 'user') {
+                return new Type\Union([
+                    new Type\Atomic\TNamedObject('App\Models\User'),
+                    new Type\Atomic\TNull(),
+                ]);
+            }
+
+            if ($method_name_lowercase === 'loginusingid') {
+                return new Type\Union([
+                    new Type\Atomic\TNamedObject('App\Models\User'),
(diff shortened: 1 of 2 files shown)


mr-feek:
Would you be able to add a corresponding test for this by chance?

src/Handlers/UserHandler.php
```diff
@@ -0,0 +1,56 @@
...
+        $method_name_lowercase = $event->getMethodNameLowercase();
+        $class_name = $event->getFqClasslikeName();
+
+        if (
+            $class_name === \Illuminate\Support\Facades\Auth::class ||
+            $class_name === 'Auth'
+        ) {
+            if ($method_name_lowercase === 'user') {
+                return new Type\Union([
+                    new Type\Atomic\TNamedObject('App\Models\User'),
```
mr-feek, line 32: This does work for the _default_ laravel install, i wonder if we could dynamically look up what object to use 🤔 

I have lots of projects that have a different namespace etc. thinking out loud

tm1000: Well two things I suppose (and overall I agree with you).

- In #184 you can declare the model but that wouldn't work in my case (where in my code I've added my own authprovider)
- If there is a place to set config (I suppose Psalm.xml) for this that would be the best place to set it I think as this plugin does not read anything for itself from Laravel's config subsystem

tm1000:
@mr-feek yes I can do tests for this. Just need some time so if you want you can draft this PR

src/Handlers/UserHandler.php
mr-feek: > If there is a place to set config (I suppose Psalm.xml) for this that would be the best place to set it I think as this plugin does not read anything for itself from Laravel's config subsystem

I wouldn't be _opposed_ to doing this -- we rely on dynamic analysis in order to fetch container dependencies etc. Whether it's possible or not to determine the correct provider to use, i dunno off the top of my head

lptn:
FYI: I'm working on alternative, more comprehensive solution

tm1000:
@lptn you can close this then no worries. Looking forward to your solution!

lptn:
@tm1000 
I just created a draft PR, feel free to review and leave some feedback: https://github.com/psalm/psalm-plugin-laravel/pull/326

PR #221, opened 2022-02-21
Fix Psalm v4.21 compatibility

mzur:
`totallyTyped` was deprecated in [Psalm v4.21.0](https://github.com/vimeo/psalm/releases/tag/4.21.0).

Drom the [docs](https://psalm.dev/docs/running_psalm/configuration/#totallytyped): `totallyTyped="true"` is equivalent to `errorLevel="1"` and `totallyTyped="false"` is equivalent to `errorLevel="2"` plus `reportMixedIssues="false"`.

PR #223, opened 2022-02-21
Fix lint action

mzur:
This fixes the lint action which seemed to have an improperly configured `actions/checkout@v2`.

See #221 for the failing tests. Both PRs should make CI green again.


mr-feek:
thanks!

PR #224, opened 2022-03-01
Add support for anonymous migrations

ekvedaras:
Adds support for anonymous migrations as at the moment it just skips them.

I couldn't find any tests in regard to this piece of functionality so was not sure where to add any 🤷


mr-feek:
thanks for the PR!

PR #228, opened 2022-06-09
Catch even Throwable when trying to create class

Legion112:
If class require connection to database for creation for example getting state from there and saving to property.
This break psalm anaylis currently. 
This case happen when CI environment does not have connection to database.


Legion112:
@mr-feek could you take a look?

mr-feek:
seems fine to me, just gonna let CI run then ill merge!

mr-feek:
thanks for the pr!

PR #229, opened 2022-06-17
feature: tighten return types of config helper by using dynamic analysis

mr-feek:
This PR should tighten the return types of the `config` helper to be the actual resolved type at runtime, rather than `Repository|mixed`. 

This should result in proper types being returned, and reducing the number of `PossiblyInvalidCast` that psalm emits after using the `config` helper


mr-feek:
Going to leave this open while i ponder how types might change depending on environment configuration

mr-feek:
note: always add the type of the default value to the return union type

Issue #247, opened 2023-01-11
Model stubs are not generated


dkuzmenchuk:
**Describe the bug**
Both fake model commands ([FakeModelsCommand210.php](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Fakes/FakeModelsCommand210.php) and [FakeModelsCommand291.php](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Fakes/FakeModelsCommand291.php)) override method `getPropertiesFromTable` and contain the single call of `$this->getProperties($model)` in the method body.
The problem is that method `getProperties` does not exist anywhere in the current plugin or Psalm sources. As I see by the branch history, this method existed in the trait [FakeModelsCommandLogic.php](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Fakes/FakeModelsCommandLogic.php). But it was renamed to `getPropertiesFromTable` in the commit [0591299](https://github.com/psalm/psalm-plugin-laravel/commit/05912991852d967921a70bc69f3d4aafd888aace#diff-ae2fe16b0f8772abe2493429bdcd3a0a6cfba3d7b393d995028e9eb93c51597aR31).
As a result, the plugin cannot generate the stubs for the Eloquent models, file `models.stubphp` is empty. So Psalm report contains a lot of errors related to the Eloquent "magic".

Please let me know if you want me to create PR to rename the method `getPropertiesFromTable` back to `getProperties`.

**Impacted Versions**
```
barryvdh/laravel-ide-helper           v2.12.3
beyondcode/laravel-dump-server        1.8.0
bugsnag/bugsnag-laravel               v2.24.0
fruitcake/laravel-cors                v3.0.0
genealabs/laravel-model-caching       0.12.5
genealabs/laravel-pivot-events        9.0.4
jackwh/laravel-new-relic              v1.0.2
laravel/cashier                       v12.17.2
laravel/framework                     v9.29.0
laravel/serializable-closure          v1.2.2
laravel/socialite                     v5.5.5
laravel/tinker                        v2.7.2
orklah/psalm-insane-comparison        v2.2.0
psalm/plugin-laravel                  v2.0.2
psalm/plugin-phpunit                  0.18.4
spatie/laravel-newsletter             4.11.0
spatie/laravel-ray                    1.30.0
staudenmeir/laravel-cte               v1.6.2
vimeo/psalm                           5.x-dev 4e177bf
```

**Additional context**
PHP 8.1.9
[...]

PR #248, opened 2023-01-12
Rename method to get model's properties

dkuzmenchuk:
Fix for issue #247

### DIFF
--- a/src/Fakes/FakeModelsCommandLogic.php
+++ b/src/Fakes/FakeModelsCommandLogic.php
@@ -26,7 +26,7 @@ trait FakeModelsCommandLogic
      *
      * @param Model $model
      */
-    public function getPropertiesFromTable($model): void
+    public function getProperties($model): void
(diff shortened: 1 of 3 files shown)


lptn:
@dkuzmenchuk 
Thanks a lot!
How it will be difficult for you to write a test that will fail on previous code and passes on the new one? Thanks a aging!

dkuzmenchuk:
@lptn I didn't think of anything better than to check if file `models.stubphp` contains strings like `class <ModelName> extends \Eloquent` for every non-abstract model. Please let me know if you have a better idea.

lptn (approved):
Hey @dkuzmenchuk 

Thanks a lot for your work, it looks great! I also now sure what is the best approach to test stubs (I'm pretty new to this repo), let's learn and improve it down the road. Great start, I hope we will collaborate more on this repo and will make it almost Larastan-level :)

PR #252, opened 2023-01-13
Update dependencies, remove unused ones

lptn:
- bump dependencies
 - remove unused dependencies (`phpoption/phpoption` dev dependency, added by 906d80fa684cd6777d6f300044a6e4dd27edd31f)
 - [normalize](https://github.com/ergebnis/composer-normalize) composer.json

I found it was a difficult decision to drop old Laravel versions support. I just joined the team and trying to prepare this repo for Laravel 10, make this plugin compatible with Larastan. It seems like the repo don't have active and stable contributors, in such situations IMO it's better to drop old versions and simplify life for new potential contributors by reducing number of supported versions. This PR is just a small step on doing it.

What do you think about this direction? E.g. I don't see any reasons of supporting old Psalm versions as they may produce a bit different error messages what is bad for testing.The same for Laravel versions - if we support 9.x, I think we can bump minor Laravel versions e.g. every 1-2 months to support versions not older than 6 months (Laravel minor updates are safe nowadays, they fully comply SerVer 2.0). As result - we can utilize more `@template` and other advanced annotations that cormmunitity added to Laravel code. I theory these updates should simplify maintenance and life of contributors.


mr-feek:
I'm up for limiting support by major release versions of this package, if you think that will make maintenance easier! My thought process before was to have `master` support all of Laravel's supported releases -- but I'm open to changing that.

lptn:
@mr-feek 
we can also support one-two previous Laravel versions, but not on master branch

E.g. [this is](https://laravel.com/docs/master/releases#support-policy) the current list of supported Laravel versions (supported by Laravel itself):
image


According this list, we can support:
 - Laravel 9 on master
 - Laravel 8 on special working branch (or even without branches)
 - Laravel 10 experimental support on 3.x branch (currently blocked by ide-helper dependency :( )

mr-feek:
> According this list, we can support:
> 
> Laravel 9 on master
> Laravel 8 on special working branch (or even without branches)
> Laravel 10 experimental support on 3.x branch (currently blocked by ide-helper dependency :( )
> I hope with such approach we can focus o the main what is very important in case of very limited human resources

Sounds good to me 👍 

Laravel 8 is losing support in a few days, so we can probably skip worrying about that.

mr-feek (approved):

PR #257, opened 2023-01-14
Improve testing experience

lptn:
As a new contributor and MacOS user I found few issues:
 - codecept are very slow (**I have no idea how to solve it ATM, do you have any ideas?**)
 - codecept failed with timeout (fixed by https://github.com/psalm/psalm-plugin-laravel/commit/a8d192dab5e77f41fea208d60460e3cfd395f775)
 - tests/laravel-test.sh and tests/lumen-test.sh created `{repo}/laravel` and `{repo}/lumen` dirs and didn't clean up after themself. There are actually 2 issues (both fixed by https://github.com/psalm/psalm-plugin-laravel/commit/772c81ad981eb1ecce55491473deb998cb82eeb8):
      1. these scripts were sensitive on the current dir when where you run them
      2. these dirs left in the filesystem without even a message that you can remove them 
 - tests/laravel-test.sh and tests/lumen-test.sh simply doesn't work on MaxOS: `sed: -I or -i may not be used with stdin`, fixed by https://github.com/psalm/psalm-plugin-laravel/commit/b3603decc2f72199a977cb85086c496d067a9d63


weirdan:
> tests/laravel-test.sh and tests/lumen-test.sh simply doesn't work on MaxOS: `sed: -I or -i may not be used with stdin`, fixed by https://github.com/psalm/psalm-plugin-laravel/commit/b3603decc2f72199a977cb85086c496d067a9d63

That's weird. Looking at the script, there are no attempts to use STDIN. And `-i` is actually needed, otherwise `sed` won't change the original file.

weirdan:
> codecept are very slow (I have no idea how to solve it ATM, do you have any ideas?)

Unfortunately, there's no easy way around this. These are e2e tests that run `psalm` cli for every test case. Psalm scans dependencies every time, so time per case is roughly proportional to the size of `vendor` folder.

lptn:
@weirdan 
You are right, it doesn't modify `composer.json`. composer support cli command like:
```
composer config repositories.psalm-plugin-laravel '{"type": "path", "url": "../psalm-plugin-laravel"}'
```

and it works:
image


not sure why it haven't used at the original commit.

I will try to find a time to debug it and fix tomorrow

mr-feek (approved):

PR #272, opened 2023-01-18
[3.x] Eloquent Collection with templated keys

lptn:
Use Larastan compatible syntax for **Eloquent** Collections:
instead of a single param syntax:
```php
/** @return \Illuminate\Database\Eloquent\Collection<\App\Models\User> */
```
use two params syntax:
```php
/** @return \Illuminate\Database\Eloquent\Collection<int, \App\Models\User> */
```

This will finally allow to use Psalm and Larastan at the same project without a lot of noise


lptn:
Hey @mr-feek 

This is the first PR for 3.x branch. I propose you to tag few alpha/beta/RC before the final release, but firstly let's support Laravel 10 on this branch, it will be my test task for the 3.x. One of ideas - use new Laravel's bin to generate Facade helpers - https://laravel-news.com/automatic-docblock-generation-on-facades. Stay tuned! :)

Issue #271, opened 2023-01-18
InvalidArgumentException Could not get class storage for eloquent


eusonlito:
**Describe the bug**

```
InvalidArgumentException Could not get class storage for eloquent
Emitted in /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:45
Stack trace in the forked worker:
#0 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ClassAnalyzer.php(783): Psalm\Internal\Provider\ClassLikeStorageProvider->get()
#1 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ClassAnalyzer.php(416): Psalm\Internal\Analyzer\ClassAnalyzer::addContextProperties()
#2 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/FileAnalyzer.php(203): Psalm\Internal\Analyzer\ClassAnalyzer->analyze()
#3 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(1581): Psalm\Internal\Analyzer\FileAnalyzer->analyze()
#4 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Fork/Pool.php(209): Psalm\Internal\Codebase\Analyzer->analysisWorker()
#5 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(373): Psalm\Internal\Fork\Pool->__construct()
#6 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Analyzer.php(272): Psalm\Internal\Codebase\Analyzer->doAnalysis()
#7 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(651): Psalm\Internal\Codebase\Analyzer->analyzeFiles()
#8 /var/www/vendor/vimeo/psalm/src/Psalm/Internal/Cli/Psalm.php(380): Psalm\Internal\Analyzer\ProjectAnalyzer->check()
#9 /var/www/vendor/vimeo/psalm/psalm(9): Psalm\Internal\Cli\Psalm::run()
#10 /var/www/vendor/bin/psalm(120): include('...')
#11 {main}

  at vendor/vimeo/psalm/src/Psalm/Internal/Fork/Pool.php:397
[...]

PR #273, opened 2023-01-18
Added Eloquent Model alias at ApplicationProvider

eusonlito:
Related with issue https://github.com/psalm/psalm-plugin-laravel/issues/271


eusonlito:
@lptn The failed tests are my bad?

lptn:
@eusonlito 
>The failed tests are my bad?

I don't think so. Codeception tests failed due to Psalm changes (a new output), Laravel projects tests failed due to a flag to find unused code. I'll try to fix both tomorrow (if there are no other volunteers), but it will be nice to add tests for your changes.

eusonlito:
@lptn 

> but it will be nice to add tests for your changes.

I don't know how codeception works 😥

mr-feek:
Thanks for the PR! 🤝 

This seems reasonable to me. 

However, I did just think of a slight edge case. Consider that `Eloquent` isn't aliased in the application at runtime in actual code, however the runtime code actually references `Eloqouent` as an alias -- psalm would think that `Eloquent` is aliased and the code will work properly, rather than resulting in an error reported by psalm. 

I really think this should be fixed upstream in ide-helper, but I think I'm okay with merging this as-is for now.

alies-dev:
Hey @eusonlito 
Do you still have this issue with `\Eloquent`? Do we need to work on this PR?

Is this PR related to the issue https://github.com/barryvdh/laravel-ide-helper/pull/1352 ?

Issue #274, opened 2023-01-18
Migrations that make columns non-null are mistaken as nullable


dpash:
**Describe the bug**

If you have a migration that changes a column to non-null:

    $table->string('column')->nullable(false)->change();

the migration parser ignores the false parameter. I imagine it would also be confused with non-changing migrations.

**Impacted Versions**
```
laravel-ide-helper/tree/v2.12.3barryvdh/laravel-ide-helper           v2.12.3           Laravel IDE Helper,...
laravel/frameworklaravel/framework                     v9.48.0           The Laravel Framework.
laravel/horizon/tree/v5.13.0laravel/horizon                       v5.13.0           Dashboard and code-...
laravel/pintlaravel/pint                          v1.4.0            An opinionated code...
laravel/saillaravel/sail                          v1.18.1           Docker files for ru...
laravel/sanctumlaravel/sanctum                       v3.2.1            Laravel Sanctum pro...
laravel/serializable-closurelaravel/serializable-closure          v1.2.2            Laravel Serializabl...
laravel/telescope/tree/v4.12.0laravel/telescope                     v4.12.0           An elegant debug as...
laravel/tinker/tree/v2.8.0laravel/tinker                        v2.8.0            Powerful REPL for t...
psalm/psalm-plugin-laravel/tree/v2.2.0psalm/plugin-laravel                  v2.2.0            A Laravel plugin fo...
laravel/tree/3.1.3sentry/sentry-laravel                 3.1.3             Laravel SDK for Sen...
laravel-health/tree/1.18.1spatie/laravel-health                 1.18.1            Monitor the health ...
laravel-ignitionspatie/laravel-ignition               1.6.4             A beautiful error p...
laravel-package-tools/tree/1.14.0spatie/laravel-package-tools          1.14.0            Tools for creating ...
laravel-permission/tree/5.8.0spatie/laravel-permission             5.8.0             Permission handling...
laravel-ray/tree/1.32.0spatie/laravel-ray                    1.32.0            Easily debug Larave...
psalm/tree/5.4.0vimeo/psalm                           5.4.0             A static analysis t...
```

**Additional context**
This can be fixed with the following code at SchemaAggregator.php:191

```php
                        $noParameter = count($root_var->args) === 0;
                        $trueParameter =
                            array_key_exists(0, $root_var->args)
                        && $root_var->args[0]->value instanceof PhpParser\Node\Expr\ConstFetch
[...]

PR #275, opened 2023-01-19
Check for a parameter with nullable() in migrations

dpash:
If you pass `false` into nullable(), for example when turning a column from nullable to non-null, the migration parsing code would ignore this and assume the column was nullable, causing false "potentially null value" errors when the column was used.

This commit adds a check for either no parameter or a true parameter before making the column as nullable.

Fixes psalm/psalm-plugin-laravel#274

### DIFF
--- a/src/Handlers/Eloquent/Schema/SchemaAggregator.php
+++ b/src/Handlers/Eloquent/Schema/SchemaAggregator.php
@@ -6,6 +6,7 @@ use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
 use PhpParser;
 
+use function array_key_exists;
 use function count;
 use function is_string;
 use function strtolower;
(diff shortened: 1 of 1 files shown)


src/Handlers/Eloquent/Schema/SchemaAggregator.php
```diff
@@ -189,7 +190,12 @@ private function processColumnUpdates(string $table_name, string $call_arg_name,
                         $root_var->name instanceof PhpParser\Node\Identifier
                         && $root_var->name->name === 'nullable'
                     ) {
-                        $nullable = true;
+                        $noParameter = count($root_var->args) === 0;
```
alies-dev, line 193: naming suggestion: how about "argument" instead "parameter"? see https://stackoverflow.com/a/59928588/4343530

lptn (approved):
Hey @dpash

Amazing 🚀!

This is really important fix, as false-positive nullables are and false-negative non-nullable are always a headache and a point of bugs. Approving, but can you please took a look to my comment below? 

👍

mr-feek (approved):
thanks for the PR!

PR #284, opened 2023-01-22
Use stubs instead handlers

lptn:
Hey @mr-feek 

I was working on improving stubs for helpers, adding missing and then realised we already have custom handlers for some helper functions. I found that you are an author of these handlers. I really respect your work and contribution to the project, this is why it's totally up to you what to use: handlers or stubs. Of course, handlers are more powerful and customisable, but we don't always need it. Some benefits of stubs:
 - more friendly for new contributors
 - single place of changes
 - less code to maintain (good in case of Psalm API change; good for new comers)

Additionally:
 - I've grouped tests by helper, so CI wii run Psalm only once per helper that should have a positive effect performance-wise.
 - Added more constraints to to status codes


mr-feek:
If something can be fully expressed in stubs, let's do it! Less code to maintain, the better 👍

mr-feek (approved):

PR #313, opened 2023-01-31
Add assertions to Collections

### LINKED ISSUE #239: Collection isNotNull isNull methods
Should be something like this
```php
    /**
     * @psalm-assert-if-true TValue !empty($this->items)
     * @psalm-assert-if-true TValue $this->first()
     * @psalm-mutation-free
     */
    public function isNotEmpty():bool {}
```

Also this option have to be set to support this check https://psalm.dev/docs/running_psalm/configuration/#memoizemethodcallresults
https://psalm.dev/docs/annotating_code/adding_assertions/    

Example from documentation:
image

lptn:
Partially fixes #239 

Hey @mr-feek 
Do you have any other ideas what we can improve in collections using assertions?


lptn:
@mr-feek 
I will merge this one  - as it's better than nothing. But I'n pen for dialog on this topic ;)

PR #335, opened 2023-02-23
Support more helpers (+18 more)

lptn:
Supported new helpers:
 - blank
 - filled
 - transform
 - class_basename
 - class_uses_recursive
 - trait_uses_recursive
 - str
 - abort_unless
 - action
 - back
 - bcrypt
 - broadcast
 - dispatch_sync
 - event
 - data_fill
 - data_set
 - precognitive
 - object_get

+ improved some existing helpers
________

BTW, this is the final bunch of stubs for handlers, it seems like all of them are covered, my favourite one is transform, see  6875b3b :)

### DIFF
--- a/stubs/Foundation/helpers.stubphp
+++ b/stubs/Foundation/helpers.stubphp
@@ -10,7 +10,7 @@
  * @param  bool  $boolean
  * @psalm-assert falsy $boolean
  * @psalm-return ($boolean is false ? never-return : void )
- * @param  int  $code
+ * @param  \Symfony\Component\HttpFoundation\Response|\Illuminate\Contracts\Support\Responsable|int<400, 511>  $code
  * @param  string  $message
  * @param  array  $headers
  * @return void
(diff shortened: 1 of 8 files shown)


stubs/Foundation/helpers.stubphp
```diff
@@ -177,9 +325,11 @@ function url($path = null, $parameters = [], $secure = null) {}
...
 
 /**
+ * Get the evaluated view contents for the given view.
+ *
  * @param  string|null  $view
- * @param  \Illuminate\Contracts\Support\Arrayable<array-key, mixed>|array<string, mixed>  $data
+ * @param  \Illuminate\Contracts\Support\Arrayable<string, mixed>|array<string, mixed>  $data
  * @param  array<string, mixed>  $mergeData
- * @return ($view is null ? \Illuminate\Contracts\View\Factory : \Illuminate\Contracts\View\View)
+ * @return ($view is null ? \Illuminate\Contracts\View\Factory : (\Illuminate\View\View|\Illuminate\Contracts\View\View))
```
aivchen, line 333: Why do we have `\Illuminate\View\View` in `@return`?

alies-dev: Good question. I briefly answered [here](https://github.com/psalm/psalm-plugin-laravel/pull/335/commits/3c6485c1ffa8083059737b3846adc44a3f2c43fc) and [here](https://github.com/psalm/psalm-plugin-laravel/pull/335/commits/2a8c68d2cc15c5e95f03158673f53aa60bc8ad1e) (pls see commit body), but in a nutshell - this helps bypass issues with similar annotations in Laravel and Larastan: there is no consistency on using this contract vs implementation. AFAIK, Laravel team wanted to remove this contract as probably nobody created and use alternative implementations, but for some reason they haven't done it in Laravel 10. As result - Laravel still somewhere uses Contract, somewhere implementation and Psalm emit MoreSpecificReturnType or LessSpecificReturnType :). Ideally it should be fixed in Laravel, but this package supports 2 major Laravel versions and big range of minor versions - so we still need to have this hack in place, at least for a while right after fixes.

PR #355, opened 2023-11-06
Run test suite with PHP 8.3

Nielsvanpach:
**Changes**

- Run test suite with 8.3, which is released in 2 weeks
- Run psalm action with 8.2 instead of 8.1 (8.1 is skipped in tests.yml)
- Run test when config file has changes

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.2, 8.0]
+        php: [8.3, 8.2, 8.0]
         laravel: [^10.0, ^9.0]
         dependencies: [lowest, highest]
         exclude:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,7 @@ on:
       - '**.feature'
       - '**.yml'
       - 'composer.json'
+      - '.github/workflows/test.yml'
   pull_request:
     paths:
       - '**.php'
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.1'
+          php-version: '8.2'
           coverage: none
 
       - name: Install composer dependencies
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -49,7 +50,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.2, 8.0]
+        php: [8.3, 8.2, 8.0]
```
alies-dev, line 53: **question:** How about dropping 8.2 version in order to speedup GH checks?

lptn (approved):
great, thank you!

.github/workflows/test.yml
Nielsvanpach: Skipping PHP versions might not be the way to go if you want to speed up or reduce the number of CI checks. PHP 8.2 is still widely used and is likely the most essential version to ensure its correct functionality.

In my opinion, a more impactful strategy for reducing the number of checks would be to consider removing the cron job for running daily actions. A commit or merge already trigger all actions to run, which should be sufficient, especially since the test suite on the main branch frequently colours red.

lptn: Ok, let's test over all supported versions. What I meant is that a contributor should not wait a lot of time to get feedback from the CI. Cron jobs on CI don't block developers in this sense. But I agree about testing using 8.2, of course, it's better to optimize the test suite, but I don't see an easy way of doing it ATM. 

Thank you for your contribution! 💚

alies-dev: Ok, let's test over all supported versions. What I meant is that a contributor should not wait a lot of time to get feedback from the CI. Cron jobs on CI don't block developers in this sense. But I agree about testing using 8.2, of course, it's better to optimize the test suite, but I don't see an easy way of doing it ATM.

Thank you for your contribution! 💚

alies-dev:
@Nielsvanpach 
BTW, there is one failing test, it's not connected to changes you made, so you can ignore it. I'm struggling to fix it for a long time, I will appreciate any help with this. This test (and incorrect types returned by the plugic) blocks the whole release (which should be great and work in some cases better than LaraStan 🚀 )

alies-dev:
it's finally releases https://github.com/psalm/psalm-plugin-laravel/releases/tag/v2.9.0 🎉

PR #359, opened 2023-11-16
Use `orchestra/testbench-core` as `testbench` should only be used for Package development

crynobone:
In package development, `orchestra/testbench` should be added explicitly which then requires `orchestra/workbench`, `orchestra/canvas` etc which isn't useful for this package usage when used in Laravel project.


alies-dev:
thank you, @crynobone ! I'll take a look into it soon, but firstly would like to understand why orchestra packages are not in dev requirements. It will take some time, but I'll do my best, thanks

crynobone:
> but firstly would like to understand why orchestra packages are not in dev requirements.

It is being used directly via the following class:

https://github.com/psalm/psalm-plugin-laravel/blob/edc45c0cd193c3e82c3503154a56cb581e4b1e20/src/Providers/ApplicationProvider.php#L11-L23

So changing to `orchestra/testbench-core` will be the most minimal changes to avoid loading requirements available via `orchestra/testbench` when using this package.

composer.json
```diff
@@ -24,7 +25,7 @@
         "illuminate/support": "^9.48 || ^10.0",
         "illuminate/view": "^9.48 || ^10.0",
         "nikic/php-parser": "^4.13",
-        "orchestra/testbench": "^7.19 || ^8.0",
+        "orchestra/testbench-core": "^7.35 || ^8.15",
```
alies-dev: >^7.35 || ^8.15"

Do these versions contain for required functionality, or are they just to enforce new versions? 

Note, this plugin built to support old versions

composer.json
```diff
@@ -14,6 +14,7 @@
         "php": "^8.0.2",
         "ext-simplexml": "*",
         "barryvdh/laravel-ide-helper": "^2.13",
+        "fakerphp/faker": "^1.21",
```
alies-dev: I couldn't find places where we use Faker on this package. Do you know them?

composer.json
```diff
@@ -34,6 +35,7 @@
         "codeception/module-cli": "^2.0",
         "codeception/module-filesystem": "^3.0",
         "codeception/module-phpbrowser": "^3.0",
+        "laravel/framework": "^9.48 || ^10.0",
```
alies-dev, line 37: this package is made to support both Laravel and Lumen. I'm not sure that require Laravel is a good idea.

alies-dev (commented):
I have some questions regarding these changes in dependencies, can you please take a look?

composer.json
crynobone: Both versions contain conflict restrictions (using composer.json) to ensure Testbench Core 7 will only be used with Laravel 9 and Testbench Core 8 only used with Laravel 10.

composer.json
crynobone: This is only require-dev deps. Previously was provided in `require` via `orchestra/testbench`. 

So this actually improves support for Lumen.

composer.json
crynobone: https://github.com/psalm/psalm-plugin-laravel/blob/edc45c0cd193c3e82c3503154a56cb581e4b1e20/stubs/legacy-factories/FactoryBuilder.stubphp#L5

alies-dev:
Great, thank you, @crynobone 
I see some issues caused updates on the trunk branch, but I'll fix them on my own.

PR #490, opened 2026-03-22
Add taint source for session/flash data retrieval

### LINKED ISSUE #479: Add taint source for session/flash data — stored taint propagation
## Summary

Session data retrieved via \`session()->get()\`, \`Session::get()\`, \`\$request->session()->get()\`, and the \`session()\` helper are not marked as taint sources. Session/flash data often originates from user input stored in a previous request, creating a stored taint vector that the current analysis misses entirely.

**OWASP:** A03:2021 — Injection (Stored XSS, Stored SQL Injection)

## The Problem

```php
// Request 1: User input stored in session
public function search(Request $request)
{
    session()->put('last_query', $request->input('q'));
    // or: session()->flash('search_term', $request->input('q'));
    return redirect()->route('results');
}

// Request 2: Session data used unsafely — NO taint warning
public function results()
{
    $query = session('last_query'); // This is user input! But Psalm sees it as clean.
    
    return response("<h1>Results for: $query</h1>"); // XSS — undetected
    // or: DB::select("SELECT * FROM items WHERE name LIKE '%$query%'"); // SQLi — undetected
}
```

### Current state

- \`\$request->input()\` is correctly marked as \`@psalm-taint-source input\`
- \`session()->get()\` returns \`mixed\` with **no taint source** annotation
- The \`old()\` helper IS marked as a taint source (in \`InteractsWithInput.stubphp\`) — this is the right precedent, since \`old()\` reads flash session data
- The \`session()\` helper with a key argument is NOT a taint source
- \`Session::get()\`, \`Session::pull()\`, \`Session::all()\` — not taint sources

## Proposed Solution

```php
// Session helper function (already has a stub in helpers.stubphp)
/**
 * @psalm-taint-source input
 */
function session($key = null, $default = null) {}

// Session Store
namespace Illuminate\Session;

class Store
{
    /** @psalm-taint-source input */
    public function get($key, $default = null) {}

    /** @psalm-taint-source input */
    public function pull($key, $default = null) {}

    /** @psalm-taint-source input */
    public function all() {}

    /** @psalm-taint-source input */
    public function only(array $keys) {}

    /** @psalm-taint-source input */
    public function remember($key, Closure $callback) {}
}
```

## Why this is important

Stored XSS and stored SQL injection are among the most dangerous vulnerability classes because:
1. They affect ALL users who view the tainted data, not just the attacker
2. They persist across requests and sessions
3. They bypass CSP nonce-based protections (the injected script appears to be first-party)
4. Session data is often treated as "internal" and not sanitized

The \`old()\` taint source is precedent that session-derived data should be treated as tainted. The same logic applies to all session retrieval methods.

## False positive considerations

This will produce some false positives when developers store system-generated data in sessions. However:
- The fix is simple: pass session data through appropriate escaping before use
- False positives here are **safer** than false negatives (missing real vulnerabilities)
- Users can suppress individual cases with inline annotations if needed

## References

- [OWASP Stored XSS](https://owasp.org/www-community/attacks/xss/#stored-xss-attacks)
- [CWE-79: Cross-site Scripting](https://cwe.mitre.org/data/definitions/79.html)

alies-dev:
## Summary

Add `@psalm-taint-source input` annotations for session data retrieval methods to detect stored XSS and stored SQL injection vulnerabilities.

**Changes:**
- `session()` helper: mark return as taint source
- New taint stub for `Illuminate\Session\Store`: `get()`, `pull()`, `all()`, `only()`, `except()`, `remove()`, `getOldInput()`, `remember()`
- Type tests for `Store` methods + taint analysis tests

## Data flow

Session data is a stored taint vector. User input enters the application in Request A, gets stored in the session, and is retrieved in Request B — where it may be used unsafely:

```
Request A (input stored):
  $_POST['q']
    → $request->input('q')           ← already a taint source
    → session()->put('q', ...)        ← data enters session storage
    → SessionHandler::write()         ← persisted to file/DB/Redis

Request B (input retrieved):
  SessionHandler::read()              ← loaded from storage
    → Store::$attributes              ← deserialized into memory
    → session('q') / $session->get()  ← ⚠️ was NOT a taint source
    → echo / DB::raw() / ...          ← reaches a taint sink
```

The gap: Psalm tracked taint in Request A but lost it across the request boundary. By marking session retrieval as a taint source, we close this gap.

## Vulnerability example

```php
// Request 1: Store user input in session
public function search(Request $request)
{
    session()->put('last_query', $request->input('q'));
    return redirect()->route('results');
}

// Request 2: Use session data unsafely
public function results()
{
    $query = session('last_query');

    // Stored XSS — attacker sets q=<script>alert(1)</script>
    return response("<h1>Results for: $query</h1>");

    // Stored SQLi — attacker sets q=' OR 1=1 --
    // DB::select("SELECT * FROM items WHERE name LIKE '%$query%'");
}
```

Without this PR, Psalm produces **no warnings** for the code above. With this PR, it correctly reports `TaintedHtml` / `TaintedSql`.

## Precedent

The `old()` helper and `InteractsWithInput::old()` are already taint sources — they read flashed form data from the session (`_old_input`). This PR extends the same logic to all session retrieval methods.

## Known limitation

`Session::get()` facade static calls may not propagate taint due to the `__callStatic` magic method losing taint context (pre-existing limitation affecting all facades). Use `session()->get()` or `$request->session()->get()` instead.

Closes #479

## Test plan

- [x] `composer test` passes (lint + psalm + unit + type tests)
- [x] Taint test: `echo session('key')` flagged as `TaintedHtml`
- [x] Taint test: `echo $session->get('key')` flagged as `TaintedHtml`
- [x] Taint test: `echo $session->pull('key')` flagged as `TaintedHtml`
- [x] Taint test: `echo $session->all()['key']` flagged as `TaintedHtml`
- [x] Taint test: `echo $session->getOldInput('key')` flagged as `TaintedHtml`
- [x] Type test: all `Store` methods return correct types
- [x] All stub signatures verified against Laravel source

### DIFF
--- a/tests/Type/tests/SessionTest.phpt
+++ b/tests/Type/tests/SessionTest.phpt
@@ -51,4 +51,4 @@ function session_store_remember(Store $session): mixed
     return $session->remember('key', fn () => 'default');
 }
 ?>
---EXPECTF--
+--EXPECT--
--- a/tests/Type/tests/TaintAnalysis/TaintedHtmlSessionStoreAll.phpt
+++ b/tests/Type/tests/TaintAnalysis/TaintedHtmlSessionStoreAll.phpt
@@ -1,0 +1,14 @@
+--ARGS--
+--no-progress --no-diff --config=./tests/Type/psalm.xml --taint-analysis
+--FILE--
+<?php declare(strict_types=1);
+
+function renderAllSessionData(\Illuminate\Session\Store $session) {
+    $data = $session->all();
+
+    echo $data['name'];
+}
+?>
+--EXPECTF--
+%ATaintedHtml on line %d: Detected tainted HTML
+TaintedTextWithQuotes on line %d: Detected tainted text with possible quotes
--- a/tests/Type/tests/TaintAnalysis/TaintedHtmlSessionStoreOldInput.phpt
+++ b/tests/Type/tests/TaintAnalysis/TaintedHtmlSessionStoreOldInput.phpt
@@ -1,0 +1,12 @@
+--ARGS--
+--no-progress --no-diff --config=./tests/Type/psalm.xml --taint-analysis
+--FILE--
+<?php declare(strict_types=1);
+
+function renderOldInput(\Illuminate\Session\Store $session) {
+    echo $session->getOldInput('email');
+}
+?>
+--EXPECTF--
+%ATaintedHtml on line %d: Detected tainted HTML
+TaintedTextWithQuotes on line %d: Detected tainted text with possible quotes
--- a/tests/Type/tests/TaintAnalysis/TaintedHtmlSessionStorePull.phpt
+++ b/tests/Type/tests/TaintAnalysis/TaintedHtmlSessionStorePull.phpt
@@ -1,0 +1,12 @@
+--ARGS--
+--no-progress --no-diff --config=./tests/Type/psalm.xml --taint-analysis
+--FILE--
+<?php declare(strict_types=1);
+
+function renderPulledSessionData(\Illuminate\Session\Store $session) {
+    echo $session->pull('user_input');
+}
+?>
+--EXPECTF--
+%ATaintedHtml on line %d: Detected tainted HTML
+TaintedTextWithQuotes on line %d: Detected tainted text with possible quotes


alies-dev:
Superseded by #557

PR #523, opened 2026-03-24
Resolve `foreignIdFor()` column type from referenced model's primary key

### LINKED ISSUE #521: feat: resolve foreignIdFor() column type from referenced model's primary key
## Description

`foreignIdFor(SomeModel::class)` always produces `int` type, even when the referenced model uses a UUID or ULID primary key. This causes false positives when the FK column is actually a `string`.

## Current behavior

```php
$table->foreignIdFor(UuidModel::class); // → SchemaColumn('uuid_model_id', 'int')
```

## Expected behavior

```php
$table->foreignIdFor(UuidModel::class); // → SchemaColumn('uuid_model_id', 'string') if PK is uuid/ulid
```

## Proposed solution

After resolving the column name via `getForeignKey()`, look up the referenced model's primary key type in the already-parsed schema:

1. Get the model's table name via `getTable()`
2. Look up the PK column (usually `id`) in `$this->tables[$tableName]`
3. Use the PK column's type for the FK column

Larastan implements this in `SchemaAggregator::getModelReferenceType()` — instantiates the model, gets its table, looks up the `id` column type from the parsed schema.

## References

- Larastan implementation: `SchemaAggregator.php` lines 412-436

alies-dev:
## Summary

Closes #521

- `foreignIdFor(UuidModel::class)` previously always produced an `unsigned int` column, causing false positives when the referenced model uses a UUID or ULID primary key
- Now looks up the referenced model's primary key column type in the already-parsed schema: if the PK is `string` (uuid/ulid), the FK column is also `string`; otherwise defaults to `unsigned int`
- Removed `foreignidfor` from `UNSIGNED_INT_METHODS` constant (was dead code after this change)

### Approach

Uses `$instance->getTable()` and `$instance->getKeyName()` to find the referenced model's PK column in `$this->tables`, then checks its type. This avoids calling `getKeyType()` which doesn't work with `newInstanceWithoutConstructor()` (trait initialization is skipped, so `HasUuids`/`HasUlids` don't set `$usesUniqueIds`).

**Known limitation**: If the referenced model's table migration hasn't been parsed yet (unusual migration ordering), the FK falls back to unsigned int. This is documented in a code comment and covered by a test.

## Test plan

- [x] Unit test: UUID model FK produces `string` column
- [x] Unit test: ULID model FK produces `string` column
- [x] Unit test: Standard int PK model FK produces `unsigned int` column
- [x] Unit test: Custom column name with UUID model produces `string`
- [x] Unit test: Nullable FK with UUID model produces nullable `string`
- [x] Unit test: Custom `$primaryKey` with UUID model produces `string`
- [x] Unit test: Missing referenced table falls back to `unsigned int`
- [x] Existing `ForeignIdForTest` still passes (custom column name)
- [x] Full test suite: lint, Psalm self-analysis, 166 unit tests, 74 type tests

### DIFF
--- a/src/Handlers/Eloquent/Schema/SchemaAggregator.php
+++ b/src/Handlers/Eloquent/Schema/SchemaAggregator.php
@@ -474,7 +474,13 @@ final class SchemaAggregator
                     break;
 
                 case 'foreignidfor':
-                    // foreignIdFor with a string column name — can't resolve model PK type, default to int
+                    // foreignIdFor() accepts a class-string as first arg (e.g. foreignIdFor('App\Models\User')).
+                    // If the string contains a namespace separator, it's a class name, not a column name —
+                    // skip it to avoid registering a bogus column. We can only resolve class-const refs above.
+                    if (\str_contains($column_name, '\\')) {
+                        break;
+                    }
+
                     $table->setColumn(new SchemaColumn($column_name, 'int', $nullable, default: $default, unsigned: true));
                     break;
 


alies-dev:
Addressed Copilot's comment in 772adac — when foreignIdFor receives a string that looks like a class-string (contains \\), we now skip column registration to avoid bogus entries.

PR #531, opened 2026-03-24
Handle `Blueprint::datetimes()` and fix `ulid()` default column name

### LINKED ISSUE #529: Schema aggregator: $table->datetimes() not handled
## Description

`Blueprint::datetimes()` creates nullable `created_at` and `updated_at` columns (datetime type), but the schema aggregator silently ignores it.

`datetimes()` is not in `METHODS_USE_HARDCODED_COLUMN_NAME` (which handles `timestamps`, `timestampstz`, etc.), so when called without arguments it hits the `continue` at line 446 ("unknown type [method call without args]") and the columns are never registered.

## Laravel source (present in both Laravel 12 and 13)

```php
// Blueprint.php
public function datetimes($precision = null)
{
    $this->datetime('created_at', $precision)->nullable();
    $this->datetime('updated_at', $precision)->nullable();
}
```

## Impact

Migrations using `$table->datetimes()` instead of `$table->timestamps()` will not have `created_at` / `updated_at` recognized by the schema aggregator, causing false positives on property access.

## Suggested fix

Add `'datetimes'` to `METHODS_USE_HARDCODED_COLUMN_NAME` and handle it in the no-args switch (lines 416-438) alongside the existing `timestamps` case:

```php
case 'datetimes':
case 'timestamps':
case 'timestampstz':
case 'nullabletimestamps':
case 'nullabletimestampstz':
    $table->setColumn(new SchemaColumn('created_at', 'string', true));
    $table->setColumn(new SchemaColumn('updated_at', 'string', true));
    break;
```

### LINKED ISSUE #530: Schema aggregator: $table->ulid() default column name maps to uuid instead of ulid
## Description

`METHODS_HAVE_DEFAULT_COLUMN_NAME` maps `'ulid' => 'uuid'` (line 43 of SchemaAggregator.php), but Laravel's `Blueprint::ulid()` defaults to column name `'ulid'`, not `'uuid'`.

## Laravel source (present in both Laravel 12 and 13)

```php
// Blueprint.php
public function ulid($column = 'ulid', $length = 26)
{
    return $this->char($column, $length);
}
```

The default column name is `'ulid'`, but the plugin registers it as `'uuid'`.

## Impact

`$table->ulid()` (without arguments) registers a column named `uuid` instead of `ulid`. Models with a `ulid` column will have the wrong column name in the schema, potentially causing false positives or false negatives.

## Suggested fix

```diff
- 'ulid' => 'uuid',
+ 'ulid' => 'ulid',
```

alies-dev:
## Summary

Fixes two schema aggregator bugs:

- **#529** — `$table->datetimes()` was silently ignored because `datetimes` was missing from `METHODS_USE_HARDCODED_COLUMN_NAME`. Now registers nullable `created_at`/`updated_at` columns, matching `timestamps()` behavior.
- **#530** — `$table->ulid()` without arguments registered a column named `uuid` instead of `ulid` due to a copy-paste error in `METHODS_HAVE_DEFAULT_COLUMN_NAME`.

**Bonus fix**: Moved the hardcoded-column-name dispatch before the `args === []` gate, so `timestamps($precision)`, `datetimes($precision)`, and similar calls with precision arguments are no longer silently ignored. This was a pre-existing bug affecting all timestamp-family methods.

## Test plan

- [x] New `DatetimesTest` — verifies `datetimes()` creates nullable columns, works with precision arg, behaves like `timestamps()`
- [x] New `UlidDefaultColumnNameTest` — verifies `ulid()` defaults to column name `ulid`, custom names work, `uuid()` default unchanged
- [x] Full test suite passes (lint + Psalm + 205 unit tests + 74 type tests)

Closes #529, closes #530

### DIFF
--- a/src/Handlers/Eloquent/Schema/SchemaAggregator.php
+++ b/src/Handlers/Eloquent/Schema/SchemaAggregator.php
@@ -17,6 +17,7 @@ final class SchemaAggregator
      * @see \Illuminate\Database\Schema\Blueprint
      */
     private const METHODS_USE_HARDCODED_COLUMN_NAME = [
+        'datetimes',
         'timestamps',
         'timestampstz',
         'nullabletimestamps',
@@ -40,7 +41,7 @@ final class SchemaAggregator
         'softdeletesdatetime' => 'deleted_at',
         'softdeletestz' => 'deleted_at',
         'uuid' => 'uuid',
-        'ulid' => 'uuid',
+        'ulid' => 'ulid',
         'ipaddress' => 'ip_address',
         'macaddress' => 'mac_address',
     ];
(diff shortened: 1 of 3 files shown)


src/Handlers/Eloquent/Schema/SchemaAggregator.php
```diff
@@ -411,35 +412,38 @@ private function processColumnUpdates(string $table_name, string $call_arg_name,
...
+                    case 'droptimestamps':
+                    case 'droptimestampstz':
+                        $table->dropColumn('created_at');
+                        $table->dropColumn('updated_at');
+                        break;
 
-                    continue; // foreach
+                    case 'remembertoken':
+                        $table->setColumn(new SchemaColumn('remember_token', 'string', $nullable));
+                        break;
```
alies-dev, line 427: Re: Copilot's comments on `rememberToken`/`dropRememberToken` — not applicable.

Both methods accept **zero parameters** in Laravel:

```php
public function rememberToken()
{
    return $this->string('remember_token', 100)->nullable();
}

public function dropRememberToken()
{
    $this->dropColumn('remember_token');
}
```

There is no custom column name support, so treating them as hardcoded is correct.

PR #689, opened 2026-04-11
fix(taint): remove false-positive TaintedHtml sinks from Blade view data

### LINKED ISSUE #684: view() helper annotated as html sink causes false-positive TaintedHtml for all Blade views
`view()` and `ResponseFactory::view()` are annotated with `@psalm-taint-sink html \$data`, which means any user input passed to a Blade view triggers `TaintedHtml`. But Blade auto-escapes all variables rendered with `{{ \$var }}` via `htmlspecialchars` — so the annotation is wrong for the common case.

This is the single biggest source of false positives in taint analysis. Every controller that does `return view('some.view', ['q' => $request->input('q')])` fires a warning, even when the template uses `{{ \$q }}`.

Real XSS via Blade requires `{!! \$var !!}` — which Psalm can't inspect because it doesn't parse `.blade.php` files.

**Possible fixes:**
- Remove the `@psalm-taint-sink html` from `view()` entirely — accepts we can't catch XSS through Blade
- Keep it but document it as intentionally noisy
- Add `@psalm-taint-escape html` to `view()` to neutralise the taint, and rely on blade-specific analysis for real XSS detection

Observed in pixelfed (10 of 30 taint issues trace back to this).

alies-dev:
## Issue to Solve

`view()` and `ResponseFactory::view()` were annotated with `@psalm-taint-sink html $data`, causing `TaintedHtml` for every controller that passes user input to a Blade view — even when the template safely uses `{{ $var }}`.

## Related

Closes #684

## Solution Description

Blade auto-escapes all `{{ $var }}` output via `htmlspecialchars()`. Real XSS requires `{!! $var !!}` (raw echo), which Psalm cannot detect since it doesn't parse `.blade.php` files. The html sink annotations created false positives with no detection benefit.

Removed `@psalm-taint-sink html` from all Blade view data parameters:
- `view()` helper — `$data`, `$mergeData`
- `Factory::make()`, `file()`, `first()`, `renderWhen()`, `renderUnless()`, `renderEach()`, `share()`
- `View::with()`
- `ResponseFactory::view()` — `$data`

File/include sinks on `Factory::file($path)` and html sinks on `ResponseFactory::make()`, `json()`, `jsonp()` are kept — those produce raw HTML responses.

## Checklist
- [x] Tests cover the change (3 new `Safe*` type tests replacing 16 false-positive tests)


alies-dev:
Closing to reopen with a clean diff against current master.

PR #708, opened 2026-04-11
stub `Conditionable::when()`/`unless()` and `Tappable::tap()` to fix `mixed` return on fluent chains

### LINKED ISSUE #704: feat: stub Conditionable::when()/unless() and Tappable::tap() to fix mixed return on fluent chains
## Problem

`->when()`, `->unless()`, and `->tap()` return `mixed` in static analysis, breaking fluent chains on builders, collections, Stringable, and paginators:

```php
User::query()
    ->when($search, fn($q, $v) => $q->where('name', 'like', "%$v%"))
    ->where('active', true)   // ERROR: method does not exist on mixed
    ->paginate();

Str::of($input)
    ->when($upper, fn($s) => $s->upper())
    ->value();                // ERROR: method does not exist on mixed
```

## Root cause

Both `Conditionable::when()` and `Tappable::tap()` return `mixed` because their source bodies call `$callback($this, ...)` — a callable invocation that Psalm can't narrow. The Laravel PHPDoc (`@return $this|TWhenReturnType`) doesn't help because template inference from callable return types produces `mixed` in Psalm 7.

## Solution

Add trait stubs that override the return type:

**`stubs/common/Support/Traits/Conditionable.stubphp`**
```php
<?php
namespace Illuminate\Support\Traits;

trait Conditionable {
    /** @return $this */
    public function when($value = null, ?callable $callback = null, ?callable $default = null);

    /** @return $this */
    public function unless($value = null, ?callable $callback = null, ?callable $default = null);
}
```

**`stubs/common/Support/Traits/Tappable.stubphp`**
```php
<?php
namespace Illuminate\Support\Traits;

trait Tappable {
    /**
     * @return ($callback is null ? \Illuminate\Support\HigherOrderTapProxy : $this)
     */
    public function tap($callback = null);
}
```

## Confirmed working

The `Conditionable` stub was verified in this codebase: with the stub loaded, `->when()` returns `Builder<Model>&static` (Psalm's `$this` representation) instead of `mixed`. Fluent chains type-check correctly.

Affects all classes using these traits:
- `Eloquent\Builder`, `Query\Builder` (via `BuildsQueries`)
- `Support\Collection`, `Support\Stringable`
- `Pagination\AbstractPaginator`, `AbstractCursorPaginator`
- `Mail\Mailable`, `Foundation\Http\FormRequest`, `Pipeline\Pipeline`

Borrowed from Larastan's stub approach. Tracked as part of the Larastan stub sync work in #688.

alies-dev:
## Issue to Solve

`when()`, `unless()`, and `tap()` return `mixed` in static analysis, breaking fluent chains on builders, collections, `Stringable`, and paginators.

## Related

Fixes #704

## Solution Description

Add trait stubs for `Conditionable` and `Tappable` that override the return type to `$this`.

Psalm can't narrow `$this|TWhenReturnType` when the callback return type is a template — inference collapses to `mixed`. Returning `$this` unconditionally keeps fluent chains type-safe for the common case (callbacks returning void or the same instance).

Verified with type tests covering all major trait consumers: `Eloquent\Builder`, `Query\Builder`, `Collection`, and `Stringable`.

## Checklist
- [x] Tests cover the change (type test in `tests/Type/tests/ConditionableTappableTest.phpt`)

### DIFF
--- a/stubs/common/Support/Traits/Conditionable.stubphp
+++ b/stubs/common/Support/Traits/Conditionable.stubphp
@@ -1,0 +1,27 @@
+<?php
+
+namespace Illuminate\Support\Traits;
+
+trait Conditionable
+{
+    /**
+     * Apply the callback if the given "value" is (or resolves to) truthy.
+     *
+     * Psalm can't narrow `$this|TWhenReturnType` when the callback return type
+     * is a template — inference collapses to `mixed`. Returning `$this` unconditionally
+     * keeps fluent chains type-safe for the common case where the callback returns void
+     * or the same builder/collection instance.
+     *
+     * @return $this
+     */
+    public function when($value = null, ?callable $callback = null, ?callable $default = null) {}
+
+    /**
+     * Apply the callback if the given "value" is (or resolves to) falsy.
+     *
+     * Same rationale as `when()` — `$this` keeps fluent chains type-safe.
+     *
+     * @return $this
+     */
+    public function unless($value = null, ?callable $callback = null, ?callable $default = null) {}
+}
--- a/stubs/common/Support/Traits/Tappable.stubphp
+++ b/stubs/common/Support/Traits/Tappable.stubphp
@@ -1,0 +1,18 @@
+<?php
+
+namespace Illuminate\Support\Traits;
+
+trait Tappable
+{
+    /**
+     * Call the given Closure with this instance then return the instance.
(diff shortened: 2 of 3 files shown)


alies-dev:
Closing as duplicate — superseded by #710, which incorporates these test cases.

Issue #849, opened 2026-04-26
Refine `Request::route()` taint source by route parameter constraints


alies-dev:
## Context

`Illuminate\Http\Request::route($param)` is currently annotated as `@psalm-taint-source input` in `stubs/common/Http/Request.stubphp:38`. That is correct in the general case (route parameters come from the URL and are user-controlled), but it produces noisy/overstated findings when the route definition already constrains the parameter to a safe character class.

Example finding (from a real app):

```php
Route::get('/authorize-dropbox/{key}', AuthorizeDropboxController::class)
    ->where('key', '[A-Za-z0-9]+');

// In controller:
$appKey = $request->route('key');
return redirect()->away("https://www.dropbox.com/oauth2/authorize?client_id={$appKey} & ...");
```

Psalm reports `TaintedHeader` here, but with `where('key', '[A-Za-z0-9]+')` the value cannot contain CRLF, `&`, `?`, `#`, `/`, etc., so neither header injection nor query-parameter injection is reachable. The taint source is, in practice, neutralized at the routing layer.

## Proposal

Enhance the plugin so that `Request::route($name)` is treated as tainted only when the route parameter lacks a constraint that would render the relevant sinks unreachable.

Concretely:

1. During analysis, collect route parameter constraints from `Route::...->where(...)`, `whereNumber`, `whereAlpha`, `whereAlphaNumeric`, `whereUuid`, `whereUlid`, `whereIn`, and pattern arguments in route definition files (`routes/*.php`).
2. For each `(routeName|uri, paramName)` pair, store the regex / known-safe class.
3. In a `MethodReturnTypeProvider` (or a dedicated taint provider) for `Request::route()`, when the literal `$param` argument is a string and a constraint is known for the current route, emit a return type without the `taint-source input` annotation, or emit it with `@psalm-taint-escape` for sink categories the constraint provably defeats (`html`, `header`, `sql`, `shell`, etc.).
4. Fall back to the current behavior (taint source) when:
   - the parameter name is dynamic,
   - no constraint is registered,
   - the constraint is too permissive to rule out the sink (e.g. `.*`).

## Open questions

- How to map a controller method back to the route(s) that reach it. Action-class routes (`Route::get(..., Controller::class)` and `[Controller::class, 'method']`) are tractable; closure-based and dynamically-registered routes are not, and should keep the current conservative behavior.
[...]

Issue #803, opened 2026-04-19
`Route::bind()` closure return type reported as `Model`, not the concrete bound class


alies-dev:
## Summary

`Route::bind('name', fn ($value) => Model::query()->firstOrFail())` triggers `MoreSpecificReturnType` + `LessSpecificReturnStatement` on the closure when its return annotation is the concrete model type — because `Builder::firstOrFail()` (like `firstOrCreate()`, see sibling issue) returns `Model` not `static`.

The sibling issue (static returns) would fix this as a side-effect, but `Route::bind` is worth calling out separately because:

1. Route model binding is **the** canonical way to register custom binders; nearly every non-trivial Laravel app has at least one.
2. Even after fixing static returns, the plugin could additionally introspect `Route::bind` / `Route::model` registrations to narrow `$request->route('name')` at consumer sites (see sibling issue about `Request::route`).

## Environment

- `psalm/plugin-laravel` dev-master @ `8017a1d`
- Psalm 7.0.0-beta19
- Laravel 12

## Reproducer

```php
<?php
use Illuminate\Database\Eloquent\Model;
use Illuminate\Support\Facades\Route;

class Genre extends Model {}

Route::bind('genre', static function (string $value): ?Genre {
    return Genre::query()->where('public_id', $value)->firstOrFail();
    // MoreSpecificReturnType (070): declared Genre|null, inferred Model
    // LessSpecificReturnStatement (129)
});
```

Koel exact diagnostics (`app/Providers/AppServiceProvider.php:82-87`):

```
MoreSpecificReturnType - The declared return type 'App\Models\Genre|null' for
  /path/app/providers/appserviceprovider.php:82:...-:closure is more specific than
  the inferred return type 'Illuminate\Database\Eloquent\Model|null'
LessSpecificReturnStatement - The type 'Illuminate\Database\Eloquent\Model' is more
  general than the declared return type 'App\Models\Genre|null'
```

## Two fixes, composable

1. **Primary (covered by sibling issue):** `Builder::firstOrFail()` / `firstWhere()` / `firstOr()` should return `TModel`. This alone unblocks the closure.

2. **Secondary, higher-value (new in this issue):** the plugin could keep a registry of bindings declared via `Route::bind('name', fn ($v) => ...)`, mapping `'name' => TModel`, so that downstream `$request->route('name')` can narrow. This is the same idea as sibling issue "Request::route typed binding" but keyed off the binder closure's inferred return type rather than route-file introspection.

[...]

PR #857, opened 2026-04-28
feat: narrow `Request::route()` by route bindings and constraints

### LINKED ISSUE #801: `$request->route('param')` returns `string|null`, discarding route-model binding
## Summary

`Illuminate\Http\Request::route($param)` is stubbed as returning `TDefault|string|null`. In routes that declare model binding (either implicit type-hinting or explicit `Route::model()` / `Route::bind()`), the runtime value is the **bound model instance** — not a string.

Consumers that chain property access — `$request->route('station')->id` — get `PossiblyNullPropertyFetch` + `InvalidPropertyFetch` (fetching on a `string`).

## Environment

- `psalm/plugin-laravel` dev-master @ `8017a1d`
- Psalm 7.0.0-beta19
- Laravel 12

## Current stub

`stubs/common/Http/Request.stubphp`:

```php
/**
 * @template TDefault of object
 *
 * @param string|null $param
 * @param TDefault $default
 * @psalm-return ($param is null ? \Illuminate\Routing\Route : TDefault|string|null)
 *
 * @psalm-taint-source input
 */
public function route($param = null, $default = null) {}
```

## Reproducer

```php
<?php
use Illuminate\Foundation\Http\FormRequest;

class RadioStation extends Model {}

class RadioStationUpdateRequest extends FormRequest
{
    public function unique(): mixed
    {
        return $this->route('station')->id;
        //                              ^^
        // PossiblyNullPropertyFetch (psalm/082) - on possibly null variable
        // of type null|string
        // InvalidPropertyFetch (psalm/029) - Cannot fetch property on non-object
        // of type string
    }
}
```

Exact diagnostics (koel `app/Http/Requests/API/Radio/RadioStationUpdateRequest.php:32`):

```
PossiblyNullPropertyFetch - Cannot get property on possibly null variable
  of type null|string
InvalidPropertyFetch - Cannot fetch property on non-object of type string
```

## Expected

When the route parameter is bound to a model (either via implicit binding or explicit `Route::bind`), `$request->route('name')` should type as `TModel|null`.

## Suggested fix

Two complementary layers:

1. **Stub return union:** add `Model` to the union in the `@psalm-return`:
   ```php
   @psalm-return ($param is null ? Route : TDefault|string|Model|null)
   ```
   This at least eliminates the "non-object" branch, letting `->id` access through (with `PossiblyNull` but no `InvalidPropertyFetch`).

2. **Plugin handler (harder, higher-value):** implement an `AfterClassLikeVisit` / route-table introspector that reads `routes/api.php` and `routes/web.php`, resolves `Route::bind('station', ...)` + implicit parameter hints, and narrows `$request->route('station')` to the exact bound model type in each route's scope.

For most users, (1) is a huge ergonomic win and would reduce koel's error count by ~10.

## Census (koel)

6 occurrences across:
- `app/Http/Resources/SongResource.php:112`
- `app/Http/Requests/API/Radio/RadioStationUpdateRequest.php:32`
- `app/Http/Controllers/AuthorizeDropboxController.php:13`

## Related

- #539 (closed) — `Route::parameter()` taint source (the sibling method)
- #758 (open) — macro discovery (route binding introspection would be a similar effort)

alies-dev:
## Summary

Closes #801, #803, #849.

Adds runtime introspection of the booted Laravel router to drive two new Psalm hooks on `Request::route('name')`:

- A `MethodReturnTypeProvider` narrows the return type to the concrete bound class when `Route::bind()` / `Route::model()` registers a Model subclass, replacing the stub's `string|null`. Eliminates the `InvalidPropertyFetch` koel saw on `$request->route('genre')->id`.
- An `AddTaintsInterface` handler refines the stub's `@psalm-taint-source input` based on the route's regex constraint. When the constraint is on a conservative whitelist (`\d+`, `[A-Za-z0-9]+`, `whereUuid`, `whereUlid`, `whereIn` of safe literals, etc.), structural-injection sinks (HTML, SQL, headers, cookies, LDAP, file paths, SSRF, XPath, sleep, has-quotes, unserialize) drop the source. Identifier-eating sinks (callable, include, eval, extract, shell) and LLM prompts keep it — `\w+` is a valid PHP function name, so `call_user_func($req->route('handler'))` must not be silently de-tainted.

The stub fallback now also includes `BackedEnum` from Laravel's implicit enum binding (`ImplicitRouteBinding::resolveBackedEnumsForRoute`), so unresolved cases still type-check.

Per-route `where()` constraints are aggregated conservatively: a name appearing on any unconstrained route is rejected (no taint suppression), and all collected regexes for the name must individually be safe.

## What's new

- `src/Handlers/Routing/RouteParameterRegistry` — read-only singleton populated once at plugin boot.
- `src/Handlers/Routing/RouteParameterRegistryBuilder` — boot entry point; warns on resolve/scan failures, falls back to an empty registry.
- `src/Handlers/Routing/RouteScanner` — walks `app('router')->getRoutes()`, uses reflection on the protected `Router::$binders` to enumerate binders, reads return types via `ReflectionFunction::getReturnType()` and falls back to captured static variables for `Route::model()`.
- `src/Handlers/Routing/SafeRoutePattern` — verbatim string-match against a small whitelist mirroring Laravel's `whereAlpha` / `whereAlphaNumeric` / `whereNumber` / `whereUuid` / `whereUlid` / `whereIn` shorthand emit. Anchors are stripped; alternations of safe literals are accepted.
- `src/Handlers/Routing/RequestRouteReturnTypeProvider` and `RequestRouteTaintHandler` — coordinated override + partial re-add.
- `src/Util/MethodCallerResolver` — promoted from `Handlers\Validation\ValidationCallerResolver` once a second handler started consuming it. `InlineValidateRulesCollector` and `ValidationTaintHandler` updated to the new path.

## Out of scope

Per-route resolution (mapping a controller method to its route(s)) and implicit binding from typed controller parameters are not covered. Both would need additional infrastructure and were left for a follow-up.

### DIFF
--- a/src/Handlers/Routing/RequestRouteReturnTypeProvider.php
+++ b/src/Handlers/Routing/RequestRouteReturnTypeProvider.php
@@ -1,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Routing;
+
+use Illuminate\Http\Request;
+use Psalm\Internal\Analyzer\StatementsAnalyzer;
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
+use Psalm\Type;
+use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Atomic\TNull;
+use Psalm\Type\Union;
+
+/**
+ * Narrows {@see Request::route()} calls when the parameter name is a literal
+ * string and {@see RouteParameterRegistry} has information about it.
+ *
+ * Two narrowing modes:
+ *
+ *  - Binding known (Route::bind / Route::model with a Model subclass) → the
+ *    return is `BoundClass|null` for the resolved subclass, replacing the
+ *    stub's `TDefault|string|Model|BackedEnum|null`. Eliminates
+ *    `InvalidPropertyFetch` on `$request->route('genre')->id` and gives the
+ *    caller the exact bound type to chain Eloquent attribute access on.
+ *    Issues #801, #803.
+ *
+ *  - No binding but a safe regex constraint is registered for the name →
+ *    the return narrows to `string|null`. The narrowing itself is what
+ *    causes Psalm to drop the stub's `@psalm-taint-source` annotation, which
+ *    is the whole point: when the route's regex defeats the relevant sinks
+ *    the value should not propagate as tainted input. The companion
+ *    {@see RequestRouteTaintHandler} keeps the source removed for sinks the
+ *    constraint provably defeats and partially re-adds it for sinks that
+ *    accept alphanumeric identifiers (callable/include/eval/extract/shell).
+ *    Issue #849.
+ *
+ * Calls without a literal first argument and calls outside the registry's
+ * coverage are left alone, so the stub's `($param is null ? Route :
+ * TDefault|string|Model|BackedEnum|null)` falls through with its taint source
+ * intact.
+ */
+final class RequestRouteReturnTypeProvider implements MethodReturnTypeProviderInterface
+{
+    /**
+     * @return list<string>
+     * @psalm-pure
+     */
+    #[\Override]
+    public static function getClassLikeNames(): array
+    {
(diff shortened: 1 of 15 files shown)


alies-dev:
closing: too small RIO

PR #870, opened 2026-05-04
Bind `TDeclaringModel` to `static` on `$this` relation dispatch

alies-dev:
## Issue to Solve

`ModelRelationReturnTypeHandler` emitted plain `bindingClass` for the `TDeclaringModel` template slot regardless of whether the receiver was `$this` or a concrete instance. Once Psalm parses `$this` in template position (vimeo/psalm#11768), user docblocks like `@return BelongsToMany<User, $this, Pivot, 'membership'>` resolve to `BelongsToMany<User, Class&static, ...>`. A method body that re-enters the handler via `$this->relation()->where(...)` then mismatched the declared `Class&static` and triggered `MoreSpecificReturnType` + `LessSpecificReturnStatement`.

Surfaced by the v4.10.0 app benchmark on solidtime's `Organization::realUsers` pattern.

## Related

Depends on vimeo/psalm#11768 for the user-visible benefit. Backward compatible with stock Psalm 7 (where `$this` does not parse in template position and the fix is a no-op).

## Solution Description

Two complementary changes:

1. **`ModelRelationReturnTypeHandler`** detects `$this` dispatch via `$event->getStmt()`. On `$this` calls it emits `TNamedObject($bindingClass, is_static: true)` so the inferred type carries the same `&static` constraint as `$this` in user docblocks. External dispatch like `(new Model())->m()` keeps the plain concrete-class emission, so existing assertions like `Rel<X, ConcreteClass>` (no `&static`) still hold. Cache key was extended with `isThisDispatch` because the same `(declaring, binding)` pair now produces two different `Union`s depending on the receiver.

2. **Relation stubs (14 files)** mark `TDeclaringModel` as `@template-covariant`. Lets the handler-emitted `Rel<X, Concrete&static>` satisfy declarations using a plain concrete class (`@return Rel<X, Campaign>`). Without covariance, change 1 would introduce `InvalidReturnType` errors on those.

### Verification

* 374 type tests pass.
* 566 unit tests pass.
* Self-analysis clean.
* App benchmark on 15 real-world apps versus v4.9.1: net **-1,614** issues, no regressions. solidtime's 2 `Organization::realUsers` errors removed (1001 -> 999). ixdf-web, which uses Psalm without vimeo/psalm#11768, goes from 7,772 to 7,759 with no new errors, confirming the fix is backward compatible with stock Psalm 7.

## Checklist

- [x] Tests cover the change. Existing relation tests in `tests/Type/tests/Relation/` exercise external dispatch and continue to pass. The `$this`-dispatch path is validated end-to-end via the solidtime app benchmark. A dedicated PHPT regression test was deliberately skipped because Psalm 7.0.0-beta19 does not parse `$this` in template position at all, so any PHPT exercising the bug pattern would simply fall through to upper bounds and pass trivially regardless of the fix. A regression test should be added once vimeo/psalm#11768 ships in a Psalm release.

### DIFF
--- a/src/Handlers/Eloquent/ModelRelationReturnTypeHandler.php
+++ b/src/Handlers/Eloquent/ModelRelationReturnTypeHandler.php
@@ -11,6 +11,8 @@ use Illuminate\Database\Eloquent\Relations\MorphPivot;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Database\Eloquent\Relations\MorphToMany;
 use Illuminate\Database\Eloquent\Relations\Pivot;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\Variable;
 use Psalm\Codebase;
 use Psalm\Internal\Analyzer\StatementsAnalyzer;
 use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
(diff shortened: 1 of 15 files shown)


alies-dev:
Superseded by #1055, which fixes the same declaring-model template collapse at the stub layer (`$this` -> `static` in HasRelationships.phpstub) without depending on the unreleased vimeo/psalm#11768, and carries the same `@template-covariant TDeclaringModel` second half. Good ideas from here (receiver-awareness, backward-compat framing) noted on #1055. Closing.

PR #880, opened 2026-05-05
Resolve Factory `for*()`/`has*()` magic methods from model relations

### LINKED ISSUE #696: Resolve `Factory` `for*()/has*()` magic methods from relationship names
Laravel factories expose magic `for*()` and `has*()` methods derived from the associated model's relationship names:

```php
Post::factory()->forUser()->hasComments(3)->create();
// forUser()      → sets BelongsTo relationship with User
// hasComments(3) → creates 3 related Comment models
```

Currently Psalm reports `UndefinedMagicMethod` for these calls.

**Implementation:** `MethodExistenceProviderInterface` + `MethodReturnTypeProviderInterface` on `Factory` subclasses — resolve relationship names from the associated model and expose `for{Relation}()`/`has{Relation}()` methods returning `static`.

alies-dev:
## Issue to Solve

Laravel factories expose magic `for{Relation}()` / `has{Relation}()` methods derived from the associated model's relationship methods, dispatched at runtime by `Factory::__call`. Psalm previously reported `UndefinedMagicMethod` for these chained calls.

```php
Post::factory()->forUser()->hasComments(3)->create();
// forUser()      → BelongsTo to User, dispatched via Factory::__call
// hasComments(3) → 3 related Comment factories, dispatched via Factory::__call
```

## Related

Closes #696

## Solution Description

`FactoryRegistrationHandler` runs in `AfterCodebasePopulated`, discovers every concrete `Factory` subclass, and resolves its target Model via two paths:

1. **Reflection** — instantiate the factory and call `modelName()`, which internally handles `#[UseModel]`, `protected $model`, and Laravel's naming-convention resolver in one shot.
2. **Static fallback** — read `TModel` from `@extends Factory<TModel>` via `ClassLikeStorage::$template_extended_params`, for factories that are not autoloadable in the analysis environment (e.g. inline declarations in PHPT type-tests).

For each resolved (factory, model) pair, `FactoryMagicMethodHandler` is registered as a per-class existence/visibility/params/return-type provider. The handler:

- Recognizes `for{Relation}()` / `has{Relation}()` whose tail names a `Relation`-returning method on the model (mirrors the runtime `Str::camel(Str::substr($method, 3))` lookup).
- Declines (returns `null`) when the method exists natively on the factory class. Psalm's `MethodCallReturnTypeFetcher` consults registered providers for *every* method call once any closure is registered, so without an explicit guard our variadic-mixed params would shadow real Factory methods like `has()`, `for()`, `hasAttached()`, `forEachSequence()`, and any user-declared override on the subclass.
- Returns the called class (each subclass registers its own handler, so chaining preserves the most-derived type without an `is_static` flag).

The native-method probe uses `use_method_existence_provider: false` so the existence_provider does not recurse back into this handler.

### Test coverage

Five PHPT tests exercise the new functionality:

- `MagicMethodsTest.phpt` — happy path: `BelongsTo`, `HasOne`, `BelongsToMany`, `MorphMany`, chains, state arrays, sequence-of-arrays form (autoloaded factory).
- `MagicMethodsTemplateExtendsTest.phpt` — `@extends Factory<TModel>` static fallback (inline non-autoloadable factory).
- `MagicMethodsNativeShadowingTest.phpt` — verifies real Factory methods (`has`, `for`, `hasAttached`) keep their real arities.
- `MagicMethodsUnknownRelationTest.phpt` — verifies `UndefinedMagicMethod` still surfaces for non-existent relations.
- `MagicMethodsUserOverrideTest.phpt` — verifies a user-declared `forVehicle(): int` wins over magic resolution.

## Checklist

- [x] Tests cover the change (5 type tests in `tests/Type/tests/Factory/`)

### DIFF
--- a/src/Handlers/Eloquent/FactoryMagicMethodHandler.php
+++ b/src/Handlers/Eloquent/FactoryMagicMethodHandler.php
@@ -1,0 +1,286 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Eloquent;
+
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\Relation;
+use Psalm\Codebase;
+use Psalm\Plugin\EventHandler\Event\MethodExistenceProviderEvent;
+use Psalm\Plugin\EventHandler\Event\MethodParamsProviderEvent;
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\Event\MethodVisibilityProviderEvent;
+use Psalm\StatementsSource;
+use Psalm\Storage\FunctionLikeParameter;
+use Psalm\Type;
+use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Union;
+
+/**
+ * Resolves Laravel Factory magic methods `for{Relation}()` and `has{Relation}()`
+ * that map to relationship methods on the associated Model.
+ *
+ *   Post::factory()->forUser()->hasComments(3)->create();
+ *
+ * Laravel's Factory::__call dispatches these dynamically: it strips the for/has prefix,
+ * camelCases the remainder, then calls `$this->newModel()->{$relationship}()` to look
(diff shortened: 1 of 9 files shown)


alies-dev:
closing as non-priority for now

Issue #903, opened 2026-05-10
Collection::random() conditional return type `($number is null ? TValue : static<int, TValue>)` collapses to TValue regardless of argument


alies-dev:
## Summary

`Illuminate\Support\Collection::random()` is stubbed with a conditional return type:

```php
/**
 * @param  (callable(self<TKey, TValue>): int)|int|null  $number
 * @return ($number is null ? TValue : static<int, TValue>)
 */
public function random($number = null, $preserveKeys = false) {}
```

When called on a real `Illuminate\Support\Collection<int, T>`, the conditional resolves to `TValue` even when the argument is clearly non-null (`->random(3)`, `->random(fn ($items) => ...)`). The condition behaves as if `$number` were always `null`.

## Reproduction

```php
/** @param Collection<int, string> $c */
function call(Collection $c): void {
    $a = $c->random();   // expected: string         actual: string           OK
    $b = $c->random(3);  // expected: Collection<...>  actual: string         BROKEN
    /** @psalm-check-type-exact $b = Illuminate\Support\Collection<int, string> */
}
```

Psalm reports:
```
CheckType: Checked variable $b = Illuminate\Support\Collection<int, string> does not match $b = string
```

## Bisecting

I isolated the bug by re-creating the exact same shape (class-level `@template`, conditional return, callable union, same default) in a synthetic class:

- Synthetic class `App\C<TKey, TValue>` with the identical signature: **conditional works** (`->f(3)` returns `App\C<...>`).
- Synthetic class `Illuminate\Support\Coll<TKey, TValue>` (same namespace but a different class name, so no parent interface): **conditional works**.
- Real `Illuminate\Support\Collection<TKey, TValue>` (implements `Enumerable`, has Laravel source visible in `vendor/`): **conditional collapses to TValue**.

The only structural difference is the presence of `interface Enumerable` (`vendor/laravel/framework/src/Illuminate/Collections/Enumerable.php`) which declares:

```php
/**
 * @param  int|null  $number
 * @return static<int, TValue>|TValue
 */
public function random($number = null);
```

I tried mirroring `Collection::random()`'s conditional onto the Enumerable interface stub (`stubs/common/Support/Enumerable.phpstub`). That **did not** fix the conditional resolution, so the interface-override hypothesis alone is not enough — there is some other interaction at play. Possibly:

- Psalm merges the source docblock with the stub instead of replacing it.
[...]

PR #904, opened 2026-05-10
Align `Collection::random()` with Laravel's actual signature

alies-dev:
## Issue to Solve

The plugin's stub for `Illuminate\Support\Collection::random()` does not match Laravel's actual signature in two ways:

1. **Callable parameter:** stubbed as `callable(TValue): int` (closure receives a single element), but Laravel calls the closure with the whole collection at `Illuminate\Collections\Collection::random` (`$number($this)`). The correct shape is `callable(self<TKey, TValue>): int`.
2. **Missing second parameter:** the real signature is `random($number = null, $preserveKeys = false)`. The stub omitted `$preserveKeys`.

This emitted spurious `InvalidArgument` (and `TooManyArguments` when `$preserveKeys` was passed) on caller code that matched Laravel's real API, e.g. monicahq/monica `app/Domains/Vault/.../VaultIndexViewHelper.php:135`:

```php
$vault->contacts
    ->random(fn (Collection $items): int => min(5, count($items)))
    ->map(fn (Contact $c) => ['id' => $c->id, ...]);
```

## Related

Filed during a benchmark audit of `UndefinedMagicMethod` findings across the v4.10.0 corpus.

This PR fixes the `InvalidArgument` half of the surface. A separate (and harder) bug remains: the conditional return type `($number is null ? TValue : static<int, TValue>)` collapses to `TValue` even for non-null arguments on the concrete `Illuminate\Support\Collection`. That interacts with the parent `Enumerable` interface's looser annotation. Filed and scoped out of this PR: #903.

## Solution Description

- `stubs/common/Support/Collection.phpstub:138-147`
  - `@param` callable arm updated from `callable(TValue): int` to `callable(self<TKey, TValue>): int`, matching Laravel's `Illuminate\Collections\Collection.php:1127` docblock and the runtime behaviour (`$number($this)`).
  - Method signature gains `$preserveKeys = false` (Laravel's second parameter).
- `tests/Type/tests/Collection/CollectionRandomTest.phpt`
  - Regression test for the closure-typed-as-`Support\Collection` pattern (monica).
  - Regression test for `random(int, bool)` not emitting `TooManyArguments`.
  - Documents what is intentionally out of scope (the conditional bug) with a link to #903 so future readers know why the test does not assert the int / callable branches of the return type.

Verified: `composer test:type -- --filter=Collection` passes (20/20). Self-analysis (`composer psalm`) clean.

## Checklist

### DIFF
--- a/stubs/common/Support/Collection.phpstub
+++ b/stubs/common/Support/Collection.phpstub
@@ -138,12 +138,13 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * @psalm-mutation-free
      *
-     * @param  (callable(TValue): int)|int|null  $number
+     * @param  (callable(self<TKey, TValue>): int)|int|null  $number
+     * @param  bool  $preserveKeys
      * @return ($number is null ? TValue : static<int, TValue>)
      *
      * @throws \InvalidArgumentException
      */
-    public function random($number = null) {}
+    public function random($number = null, $preserveKeys = false) {}
 
     /**
      * @psalm-mutation-free
--- a/tests/Type/tests/Collection/CollectionRandomTest.phpt
+++ b/tests/Type/tests/Collection/CollectionRandomTest.phpt
@@ -1,0 +1,68 @@
+--FILE--
+<?php declare(strict_types=1);
+
+use App\Models\Customer;
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
+use Illuminate\Support\Collection as BaseCollection;
+
+/**
+ * Collection::random() stub alignment with Laravel's actual signature.
+ *
+ * Covers:
+ *  - `random()` with no argument returns a single TValue (works correctly today).
+ *  - `random(callable)` accepts a closure receiving the whole collection
+ *    (callable(self<TKey, TValue>): int), matching Laravel's
+ *    `Illuminate\Collections\Collection::random` where `$number($this)` is called.
+ *    Before this stub change the param was `callable(TValue): int`, which produced
+ *    a spurious `InvalidArgument` on callers like
+ *    `$collection->random(fn (Collection $items): int => ...)`.
+ *  - `random(int, bool)` compiles without TooManyArguments — the `$preserveKeys`
+ *    second parameter was missing from the stub.
+ *
+ * NOT covered here: the conditional return type `($number is null ? TValue :
+ * static<int, TValue>)` does not currently narrow correctly when `$number` is
+ * non-null. That is tracked separately in
+ * https://github.com/psalm/psalm-plugin-laravel/issues/903 and is out of scope
+ * for this stub-signature fix.
+ */
+final class CollectionRandomTest
+{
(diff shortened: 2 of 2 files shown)


alies-dev:
## Benchmark delta (PR vs `v4.10.0`)

Ran `/psalm-app-benchmark fix-collection-random-stub-callable-signature as v4.10.0-pr904` on the corpus (excluding `ixdf-web` per request). JSON results not retained.

### Per-app totals

| app | v4.10.0 | this PR | Δ |
|---|---:|---:|---:|
| aureuserp | 149 | 149 | 0 |
| bagisto | 35 | 35 | 0 |
| bookstack | 3,939 | 3,939 | 0 |
| coolify | 22,161 | 22,161 | 0 |
| corcel | 374 | 374 | 0 |
| filament | 14,060 | 14,060 | 0 |
| laravel-excel | 1,213 | 1,213 | 0 |
| **monica** | **6,654** | **6,650** | **-4** |
| pixelfed | 25,617 | 25,617 | 0 |
| solidtime | 1,704 | 1,704 | 0 |
| spatie-dashboard | 223 | 223 | 0 |
| tastyigniter | 58 | 58 | 0 |
| unit3d | 12,063 | 12,063 | 0 |
| vito | 3,406 | 3,406 | 0 |
| **TOTAL** | **91,656** | **91,652** | **-4** |

### What changed

All four removed issues are in monica's `app/Domains/Vault/ManageVault/Web/ViewHelpers/VaultIndexViewHelper.php` — the exact file used as the motivating example in the PR description.

| location | issue type | before | after |
|---|---|---|---|
| L131 | `MissingPureAnnotation` | 1 | 0 |
| L133 | `MixedReturnStatement` | 1 | 0 |
| L134 | `InvalidArgument` (the directly-targeted false positive) | 1 | 0 |
| L135 | `UndefinedMagicMethod` (`Contact::map`) | 1 | 0 |

L134 is the direct effect of the stub change. L131/L133/L135 are downstream: once Psalm accepts the callable argument as valid, it can narrow the `random(closure)` return type and the chained `->map(...)` resolves cleanly, which in turn makes the surrounding method's purity/return analysis succeed.

### Analytics

- **No regressions.** 13 of 14 apps unchanged at 0 delta.
- **No type-distribution shifts** anywhere else — the conditional return type bug ([#903](https://github.com/psalm/psalm-plugin-laravel/issues/903)) caps the upside; monica's `random(callable)` path happens to escape it, but any `random(int, ...)` caller in the corpus would still be broken (none found).
- **Effort:efficiency:** small, surgical stub fix (3 lines + 1 new test); -4 real issues across the corpus; 0 false positives introduced. No noticeable perf change.
- **Why so concentrated?** `Collection::random()` with a closure receiving the collection is rare in real apps — most callers pass an int or nothing. So the addressable population is small, but where it does hit, it cascades.

If/when [#903](https://github.com/psalm/psalm-plugin-laravel/issues/903) (conditional return type collapses to `TValue`) lands, the same fix will yield further deltas on apps that use `random(int)` and `random(callable)` and chain on the result.

PR #909, opened 2026-05-11
CI: verify the plugin works under psalm.phar

### LINKED ISSUE #895: CI: verify the plugin works under psalm.phar (not just vendor/bin/psalm)
## Problem

`Plugin::registerHandlers()` carries 41 paired `require_once __DIR__ . '/Handlers/...'` + `$registration->registerHooksFromClass(...)` lines. Sister plugins differ:

- `psalm-plugin-symfony` keeps the same `require_once` pattern.
- `psalm-plugin-phpunit` and `Lctrs/psalm-psr-container-plugin` use `class_exists(\$fqcn, true)` instead (each has only one handler though).

The defense exists because Psalm itself (`vendor/vimeo/psalm/src/Psalm/Config.php::loadPlugin`) sidesteps `spl_autoload_register` and direct-`require_once`s the plugin entry-point file. The maintainers' own comment in that loader: _"It may seem that the last step will always fail, but it's only true if project uses Composer autoloader."_ That is, Psalm does not guarantee an active project autoloader at plugin-invocation time.

In practice, the plugin also boots Testbench → requires `Illuminate\…` autoloading → so an environment where our own classes are NOT autoloadable is hard to construct. But \"hard to construct\" is not \"impossible.\" The current CI uses `vendor/bin/psalm` exclusively, so we have zero evidence that the plugin works under `psalm.phar`.

This issue blocks a maintainability cleanup in `Plugin.php` (currently ~793 lines, of which ~155 are the handler registration block). With empirical evidence that `.phar` works, the block can be collapsed to a `private const HANDLERS = [...]` + `foreach` loop without the per-handler `require_once`.

## Proposed work

Add a CI job (e.g., `test-laravel-app-phar.yml` or a job in the existing `test-laravel-app.yml`) that:

1. Installs `psalm.phar` from a pinned Psalm release (`https://github.com/vimeo/psalm/releases/download/<version>/psalm.phar`) — NOT via Composer.
2. Sets up a fresh Laravel project (mirror what `tests/Application/laravel-test.sh` does today).
3. `composer require --dev psalm/plugin-laravel` to install the plugin.
4. Runs `php psalm.phar --no-cache` against the project (instead of `vendor/bin/psalm`).
5. Asserts:
   - The plugin initializes without warnings (\"Laravel plugin error on initialisation\" must NOT appear in output).
   - At least one expected plugin diagnostic is emitted on a known-bad fixture (e.g., a `MissingTranslation` or `NoEnvOutsideConfig` issue) — proves handlers were actually registered.
6. Runs across the same PHP × Laravel matrix as the existing `test-laravel-app.yml`.

## Acceptance criteria

- New CI job runs on PRs and pushes to master.
- Job passes consistently for at least one full week of green runs (no flake).
- Job fails clearly if a handler silently doesn't register (e.g., by checking the diagnostic count on a fixture).
- `docs/contributing/decisions.md` updated: replace \"issue link TBD\" with this issue's URL, and add a follow-up note about whether the constraint can be lifted.

## Out of scope (separate work, gated by this)

- Refactoring `Plugin::registerHandlers()` to drop `require_once`.
- Refactoring the entry point to use a centralized `private const HANDLERS = [...]` list.
- Other Plugin.php cleanups (stub-finder extraction, schema-builder extraction, init-helper relocation) — those are independent and don't depend on `.phar` validation.

## Context

- Affected file: `src/Plugin.php` (lines ~205-356, the `registerHandlers()` method).
- Existing CI: `.github/workflows/test-laravel-app.yml` runs the plugin under `vendor/bin/psalm` only.
- Decision record: `docs/contributing/decisions.md` § \"Class Loading and Discovery\" → \"Handler files are loaded with explicit `require_once`, not via Composer autoload\".
- Internal docs (`CLAUDE.local.md`, `AGENTS.md`) reference this issue under \"Plugin Initialization\".

alies-dev:
Closes #895. **First run revealed a category-wide latent bug; see #910.**

## Summary

Adds `.github/workflows/test-laravel-app-phar.yml`. The job downloads the Psalm version composer resolves in a freshly-built Laravel project, runs the plugin under `psalm.phar` (not `vendor/bin/psalm`), and asserts that `NoEnvOutsideConfig` fires on a tiny bad-env fixture. The diagnostic firing is positive evidence that handlers actually registered, not just absence of errors.

This is the empirical gate that was meant to unblock collapsing `Plugin::registerHandlers()`'s 41 `require_once` + `registerHooksFromClass` pairs into a `foreach` loop over a `HANDLERS` constant (see `docs/contributing/decisions.md` → "Handler files are loaded with explicit `require_once`"). **That cleanup is now gated on #910 instead.**

## Empirical finding (first run)

I triggered the new workflow on this PR via a throwaway `pull_request` trigger (since `gh workflow run` cannot dispatch a workflow that does not yet exist on master). Every matrix cell failed with the same fatal:

```
PHP Fatal error: Cannot redeclare class Psalm\Internal\ErrorHandler
  (previously declared in phar:///.../psalm.phar/src/Psalm/Internal/ErrorHandler.php:20)
  in /.../vendor/vimeo/psalm/src/Psalm/Internal/ErrorHandler.php on line 24
```

Full logs: https://github.com/psalm/psalm-plugin-laravel/actions/runs/25658024029

This is a **dual-Psalm collision**, not the autoloader-not-registered scenario the `require_once` defense was guarding against. When the plugin is composer-installed, `vimeo/psalm` comes along as a hard dependency (declared in `require`). Then when `psalm.phar` runs, the phar's bundled Psalm and the composer-installed Psalm fight in PHP's class table. Sister plugin `psalm-plugin-symfony` has the identical composer.json shape and only tests via `vendor/bin/psalm` in its CI; the bug is presumably category-wide. Filed as #910.

The throwaway trigger commit was then dropped (force-push-with-lease), so the diff in this PR matches the original spec: triggers are `workflow_dispatch` + `release: published` + weekly Friday cron.

## Triggers (per maintainer spec)

- `workflow_dispatch` (manual)
- `release: types: [published]` (post-tag smoke test)
- `schedule: "0 6 * * 5"` (weekly Friday, matches `test-laravel-app.yml`)

Intentionally NOT on PR/push: the matrix is heavy (7 cells) and the failure modes here are slow-moving.

## Job structure

1. **Create Laravel app + install plugin** — reuses `tests/Application/laravel-test.sh` with the new `--setup-only` flag (exits before Psalm runs).
2. **Download psalm.phar** matching the composer-resolved Psalm version. Includes a semver guard that rejects `dev-master` and Composer path-repo aliases (`9999999-dev`), plus an integrity check via `php psalm.phar --version | grep -q '^Psalm'` to catch truncated downloads.
3. **Baseline check** — runs `psalm.phar` against the existing `laravel-test-psalm.xml` config.
4. **Positive assertion** — drops `app/PharCheck/PharCheckEnvFixture.php` (calls `env()` outside `config/`) into the test app, runs `psalm.phar` with a stripped-down config, and three-way-gates on (a) non-zero psalm exit, (b) no init-failure warning in output, (c) `NoEnvOutsideConfig` appears in output. Distinct error messages per failure mode so a crash before output is not misreported as "handler failed to register".

## Companion changes

- `tests/Application/laravel-test.sh`: adds `PSALM_BINARY` env var (default `./vendor/bin/psalm`) and `--setup-only` flag. Existing `composer test:app` path is unchanged.
- `docs/contributing/decisions.md`: updates the "How to remove this constraint" paragraph to note the new CI job exists, with the caveat that it does not run on every PR.

## Recommended path forward

**Land this PR as proof-of-discovery.** The workflow correctly identifies that the plugin is broken under `psalm.phar` (first run, every cell). That's exactly what #895 asked for. Until #910 is fixed:

- The post-merge `gh workflow run test-laravel-app-phar.yml --ref master` will fail; the failure is the deliverable.
- The weekly Friday cron and release-trigger runs will keep failing as a visible reminder.
- The `require_once` cleanup contemplated in `docs/contributing/decisions.md` stays gated.

If you'd rather not merge a workflow that fails on first run, the alternative is to keep this PR open until #910 is fixed and the workflow can produce a green signal.

## Validation

- `actionlint` passes on the new workflow.
- `bash -n` passes on the modified script.
- Two rounds of `/psalm-review` (code, architecture, docs, tests, errors agents) before push.
- Workflow ran end-to-end on the PR via a throwaway `pull_request` trigger (since reverted) and behaved as designed: the three-way gate correctly identifies the dual-Psalm collision as a non-init-warning, psalm-exited-non-zero, no-NoEnvOutsideConfig-in-output failure mode and surfaces it cleanly.

### DIFF
--- a/.github/workflows/test-laravel-app-phar.yml
+++ b/.github/workflows/test-laravel-app-phar.yml
@@ -1,0 +1,193 @@
+name: Test Laravel app (psalm.phar)
+
+# Validates the plugin works when Psalm is run as a standalone phar (not
+# vendor/bin/psalm). This is the "phar class-loading" gate referenced in
+# docs/contributing/decisions.md → "Handler files are loaded with explicit
+# require_once". A passing run here is the empirical evidence that unblocks
+# collapsing src/Plugin.php::registerHandlers() into a foreach loop.
+# See https://github.com/psalm/psalm-plugin-laravel/issues/895.
+#
+# Triggers are deliberately narrow: manual + on a new release (post-tag smoke
+# test) + a weekly schedule (catches drift between releases). We do NOT run
+# this on every PR/push because the matrix is heavy and the failure modes are
+# slow-moving (Psalm's plugin loader and our require_once block both change
+# rarely).
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+  schedule:
+    # Weekly on Friday at 06:00 UTC, matching test-laravel-app.yml.
+    - cron: "0 6 * * 5"
+
+# Cancel superseded runs on the same ref (mostly relevant for manual dispatch).
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Test Laravel ${{ matrix.laravel-installer-version }} under psalm.phar (PHP ${{ matrix.php-version }})
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - '8.2'
+          - '8.3'
+          - '8.4'
+          - '8.5'
+        laravel-installer-version:
+          - '12.12.2'
+          - 'dev-master'
+        exclude:
+          - php-version: '8.2'
+            laravel-installer-version: 'dev-master'
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          coverage: none
+          tools: composer:v2
+
+      - name: Create Laravel app and install plugin (no Psalm run)
+        run: |
+          LARAVEL_INSTALLER_VERSION=${{ matrix.laravel-installer-version }} \
+            ./tests/Application/laravel-test.sh --verbose --setup-only
+
+      - name: Download psalm.phar matching the resolved Psalm version
+        working-directory: tests-app/laravel-example
+        run: |
+          # Read the exact Psalm version composer resolved in the test project.
+          # jq is preinstalled on GitHub-hosted ubuntu runners. Parsing JSON is
+          # more robust than text-parsing `composer show`.
+          PSALM_VERSION=$(composer show -f json vimeo/psalm | jq -r '.versions[0]')
+          if [ -z "$PSALM_VERSION" ] || [ "$PSALM_VERSION" = "null" ]; then
+            echo "::error::Could not resolve installed vimeo/psalm version"
+            composer show vimeo/psalm
+            exit 1
+          fi
+          echo "Resolved Psalm version: $PSALM_VERSION"
+          # Restrict to a proper semver tag. This rejects both `dev-master`
+          # (no phar published) and Composer's path-repo aliases like
+          # `9999999-dev`, where building a GitHub release URL would 404 with
+          # a less actionable message.
+          if [[ ! "$PSALM_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+            echo "::error::Resolved Psalm version '$PSALM_VERSION' is not a tagged semver release; no phar artifact is published for dev refs."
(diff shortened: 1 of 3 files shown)


alies-dev:
low priority as for now

PR #917, opened 2026-05-11
Use `static` instead of `$this` in relation factory return templates

### LINKED ISSUE #913: MixedMethodCall when chaining any method off a relation factory (`$this->belongsTo()/hasMany()/...->...`) inside relation methods
## Summary

Psalm reports `MixedMethodCall` when chaining `withoutGlobalScopes()` directly onto a `belongsTo()` call inside a relationship definition. The receiver of `withoutGlobalScopes()` (the `BelongsTo` returned by `belongsTo()`) is treated as `mixed`, breaking the fluent chain.

## Versions

- `psalm/plugin-laravel`: `dev-master` @ `9304208713a57542c08b14ed12e073cddbef978f` (composer constraint `^4.0@RC`)
- `psalm/plugin-phpunit`: `0.20.1`
- `laravel/framework`: v12
- PHP 8.5

## Reproducer

```php
namespace Domain\Cart\Models;

use Domain\Customer\Models\Customer;
use Illuminate\Database\Eloquent\Relations\BelongsTo;

final class AbandonedCart extends Model
{
    /** @return BelongsTo<Customer, self> */
    public function customer(): BelongsTo
    {
        return $this->belongsTo(Customer::class)->withoutGlobalScopes();
    }
}
```

Reported error:

```
ERROR: MixedMethodCall
at src/Cart/Models/AbandonedCart.php:40:51
Cannot determine the type of the object on the left hand side of this expression (see https://psalm.dev/015)
        return $this->belongsTo(Customer::class)->withoutGlobalScopes();
```

The column points at `->withoutGlobalScopes()`. Psalm has lost the type of the `$this->belongsTo(Customer::class)` receiver.

## Expected behaviour

`belongsTo()` is stubbed (in `stubs/common/Database/Eloquent/Concerns/HasRelationships.phpstub`) to return `\Illuminate\Database\Eloquent\Relations\BelongsTo<TRelatedModel, $this>`, and `withoutGlobalScopes(?array $scopes = null): $this` is declared in the Builder stub. Via the Relation's `@mixin Builder<TRelatedModel>`, the chain should resolve to `BelongsTo<Customer, AbandonedCart>` so the method's `@return BelongsTo<Customer, self>` is satisfied without a Mixed cascade.

## Notes / context

- Likely the same family as #190 (`Support hasMany(...)->orderBy()`), which was fixed in #642 by adding `MethodForwardingHandler`. That handler covers QueryBuilder-only methods that come in through `__call` (Path 2: `orderBy`, `limit`, `groupBy`, etc.) and Builder-mixin methods (Path 1). `withoutGlobalScopes()` is declared directly in the Builder stub with `@return $this`, so it's a Path 1 / mixin-interception candidate — but the chain still falls back to Mixed at the immediate `belongsTo()` receiver, not just at later links.
- Workaround: assign the relation to a typed local first, then call the chained method:
  ```php
  /** @var BelongsTo<Customer, self> $relation */
  $relation = $this->belongsTo(Customer::class);
  return $relation->withoutGlobalScopes();
  ```
  Forcing the type with the intermediate variable resolves the error, which confirms the regression is in how the chained call infers the receiver type rather than in the Builder stub's `withoutGlobalScopes()` signature itself.

I haven't dug further into whether the gap is in `MethodForwardingHandler`'s mixin-interception path, in the stubbed `belongsTo()` `@return ... <TRelatedModel, $this>` template resolution under chaining, or in core Psalm 7's handling of `$this` through `@mixin`. Happy to provide more debug info if helpful.

alies-dev:
## Summary

`\$this->belongsTo(X::class)->withoutGlobalScopes()` (and similar chained calls on relation factories inside Model methods) raised `MixedMethodCall` because the intermediate factory call collapsed to `mixed`. The root cause sits in the stub: `@return Relation<TRelated, \$this>` in `HasRelationships` left the TDeclaringModel slot unsubstituted under Psalm 7. Swapping `\$this` for `static` substitutes correctly and produces the right concrete generic type.

The receiver of any relation factory is always the model itself, so `static` and `\$this` are semantically equivalent at the TDeclaringModel slot. This is a stub divergence from Laravel source documented inline in the trait.

Applies to all 11 factory methods in `HasRelationships`: `hasOne`, `hasOneThrough`, `morphOne`, `belongsTo`, `morphTo`, `hasMany`, `hasManyThrough`, `morphMany`, `belongsToMany`, `morphToMany`, `morphedByMany`.

## Notes for reviewers

- The stub change is documented at the trait level so future stub-sync passes will preserve the divergence (see #913 for context).
- The regression test asserts both the intermediate factory type (a load-bearing `@psalm-check-type-exact`, since a future regression that re-collapses the intermediate to `mixed` would still satisfy a declared return type) and the full chained-return shape for the four structurally distinct relation shapes:
  - 2-template (`belongsTo`, `hasOne`, `hasMany`)
  - 3-template through (`hasManyThrough`)
  - 4-template pivot-bearing (`belongsToMany`)
  - No-TRelatedModel (`morphTo`)
- The remaining methods (`morphOne`, `morphMany`, `hasOneThrough`, `morphToMany`, `morphedByMany`) share the same template shapes and `static`-substitution path with covered methods.
- Possible follow-up worth a separate pass: `ModelRelationReturnTypeHandler` was introduced specifically to work around the same `\$this`-in-template Psalm limitation for user-defined relation methods. Now that the factory stubs work directly, that handler might have some redundant logic to simplify.

Fixes #913


alies-dev:
Superseded by #1055, which implements both coupled halves of the #913 fix (`$this`→`static` **and** `@template-covariant TDeclaringModel`). This PR did only the first half and regresses on non-final models. Ideas worth keeping carried over in https://github.com/psalm/psalm-plugin-laravel/pull/1055#issuecomment-4723909698.

PR #921, opened 2026-05-15
Infer `list<T>` from `Collection::values()->all()` chain

alies-dev:
## Issue to Solve

`Collection::values()->all()` infers `array<int, TValue>` instead of `list<TValue>`. Callers that need a list (JSON serialization of an array of objects, `array_is_list()` guards, code typed against `list<T>`) have to cast or annotate the result manually, even though `values()` provably produces sequentially indexed keys.

## Solution Description

Adds `CollectionValuesAllHandler`, a `MethodReturnTypeProvider` registered on `Support\Collection`, `Support\LazyCollection`, and `Eloquent\Collection`. It narrows `all()` to `list<TValue>` only when the call is chained immediately after `values()` (i.e. `$c->values()->all()`), keyed on the AST shape of the receiver.

### Why a handler, not a stub conditional

A pure-stub approach was attempted first: `values()` narrowed to `static<int<0, max>, TValue>` plus `all()` keyed on `(TKey is int<0, max> ? list<TValue> : array<TKey, TValue>)`. **This was unsound.** `int<0, max>` requires only non-negativity, not contiguity from 0: a `Collection<int<0, max>, T>` with keys `[1 => ..., 3 => ...]` satisfies the bound but is not list-shaped, so the conditional would have emitted false `list<T>` for any such collection. AST chain detection avoids the issue entirely by only firing on the literal call shape where listness is guaranteed.

### Documented limitations (both purely syntactic)

- Variable-bound chains `$v = $c->values(); $v->all()` keep the upstream `array<TKey, TValue>` shape. The receiver of `all()` is a `Variable` AST node, not an immediate `values()` `MethodCall`.
- Nullsafe operators `$c?->values()?->all()` keep the upstream shape. Psalm's `NullsafeAnalyzer` desugars `?->all()` to a `VirtualMethodCall` whose receiver is a synthesized temp variable, not the inner `values()` call.

Both limitations are pinned by dedicated test cases so future AST broadening is observable.

### Test coverage

`tests/Type/tests/Collection/CollectionValuesAllTest.phpt` covers nine cases:

- Chain produces `list<T>` on `Support\Collection`, `LazyCollection`, and `Eloquent\Collection`.
- Plain `all()` keeps `array<TKey, TValue>` on int-keyed and string-keyed inputs (no regression for collections keyed by model id, etc.).
- Enumerable interface chain keeps `array` (handler registered on concrete classes only).
- The two documented limitations (variable-bound, nullsafe) pinned to current behavior.

Full type suite (394/394) passes; `composer psalm` and `composer cs` clean.

### DIFF
--- a/src/Handlers/Collections/CollectionValuesAllHandler.php
+++ b/src/Handlers/Collections/CollectionValuesAllHandler.php
@@ -1,0 +1,101 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Collections;
+
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
+use Illuminate\Support\Collection;
+use Illuminate\Support\LazyCollection;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Identifier;
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
+use Psalm\Type;
+use Psalm\Type\Union;
+
+/**
+ * Narrows Collection::all() to list<TValue> when the call is chained
+ * immediately after values() (i.e. `$c->values()->all()`).
+ *
+ * The chain is the only call shape where the result is guaranteed to be a
+ * list — values() calls array_values() (or a generator yielding without
+ * keys for LazyCollection), so the inner array always has consecutive
+ * 0-based integer keys.
+ *
+ * A purely stub-based conditional return on all() (e.g. keyed on
+ * `TKey is int<0, max>`) would mislabel any int-keyed collection whose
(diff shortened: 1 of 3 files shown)


alies-dev:
Thanks @Copilot — all 4 comments correctly flag that `int<0, max>` allows sparse non-negative keys (e.g. `[1 => ..., 3 => ...]`) and is not equivalent to list shape.

Force-pushed a fix that replaces the unsound stub conditional with `CollectionValuesAllHandler`, a `MethodReturnTypeProvider` that only narrows `all()` to `list<T>` when the call is chained immediately after `values()` (AST shape detection). No more `int<0, max>` proxy — the stubs are reverted and the handler keys on the literal call shape, which is the only form where listness is provable.

Updated PR description covers the rationale and the two documented syntactic limitations (variable-bound chain and nullsafe), both pinned by dedicated test cases.

PR #936, opened 2026-05-15
Type SoftDeletes `restore`/`restoreOrCreate`/`createOrRestore` on `Builder`

### LINKED ISSUE #929: `Builder::restore()` / `restoreOrCreate()` / `createOrRestore()` not typed
## Problem

`SoftDeletingScope` extends every SoftDeletes-using model's `Builder` with three macros: `restore`, `restoreOrCreate`, `createOrRestore`. These are registered at runtime via `Builder::macro()` and are not declared as `@method static` on the `SoftDeletes` trait (which only declares `withTrashed`, `onlyTrashed`, `withoutTrashed`, `restoreOrCreate`, `createOrRestore`). Builder-instance calls are not intercepted by the plugin.

## Where it breaks

`src/Handlers/Eloquent/ModelRegistrationHandler.php:418` (`extractBuilderReturningMethods`) — filters trait `@method static` declarations to those returning a `Builder` type. `restoreOrCreate`/`createOrRestore` return `static` (the model), so they are excluded. `restore` has no `@method` declaration at all (pure runtime macro).

No stubs match `restore`/`restoreOrCreate`/`createOrRestore` in `stubs/`:

```
grep -rn "restore\|restoreOrCreate\|createOrRestore" stubs/
# (no output)
```

## Reproduction

Test file added at `tests/Type/tests/Builder/SoftDeletesRestoreTest.phpt`:

```php
function test_restore_on_query_returns_int(): void {
    $_result = Customer::query()->restore();
    /** @psalm-check-type-exact $_result = int */
}

function test_restore_or_create_on_query_returns_model(): void {
    $_result = Customer::query()->restoreOrCreate(['email' => 'a@b.c']);
    /** @psalm-check-type-exact $_result = Customer */
}

function test_create_or_restore_on_query_returns_model(): void {
    $_result = Customer::query()->createOrRestore(['email' => 'a@b.c']);
    /** @psalm-check-type-exact $_result = Customer */
}
```

Expected: each `@psalm-check-type-exact` passes. Current behavior: plugin returns `mixed` or emits `UndefinedMagicMethod`.

## Laravel source

`Illuminate\Database\Eloquent\SoftDeletingScope`:

```php
protected function addRestore(Builder $builder)
{
    $builder->macro('restore', function (Builder $builder) {
        $builder->withTrashed();
        return $builder->update([$builder->getModel()->getDeletedAtColumn() => null]);
    });
}

protected function addRestoreOrCreate(Builder $builder)
{
    $builder->macro('restoreOrCreate', function (Builder $builder, array $attributes = [], array $values = []) {
        $builder->withTrashed();
        return tap($builder->firstOrCreate($attributes, $values), function ($instance) {
            $instance->restore();
        });
    });
}
```

Return types:
- `restore` → `int` (count of restored rows via `Builder::update()`)
- `restoreOrCreate` → `TModel`
- `createOrRestore` → `TModel`

## Fix sketch

Two approaches:

**A. Hardcode in `BuilderScopeHandler`** — when the LHS model uses `SoftDeletes`, special-case these three method names. Mirrors Larastan `EloquentBuilderForwardsCallsExtension:96-129`.

**B. Scan `SoftDeletingScope::extend()`** AST for `->macro(name, closure)` calls and extract return type from the closure. Generic enough to handle other scope-extension traits. Higher implementation cost.

(A) is the fast path; (B) is principled but probably over-engineering for one trait.

## References

- Larastan: `.alies/larastan/src/Methods/EloquentBuilderForwardsCallsExtension.php:96-129`
- Laravel: `Illuminate\\Database\\Eloquent\\SoftDeletingScope`

alies-dev:
## Issue to Solve

`SoftDeletingScope::extend` registers three macros on `Builder` at runtime that the plugin did not type:

- `restore` (no `@method` declaration anywhere)
- `restoreOrCreate` and `createOrRestore` (declared on the `SoftDeletes` trait as `@method static static`, so `extractBuilderReturningMethods` filters them out because the return type is the model, not `Builder`)

Builder-instance calls (`Customer::query()->restore()`, `->restoreOrCreate(...)`, `->createOrRestore(...)`) therefore returned `mixed` and broke argument validation.

## Related

Fixes #929.

## Solution Description

Approach A from the issue (mirroring Larastan's `EloquentBuilderForwardsCallsExtension`):

- Track SoftDeletes-using models in a new registry on `BuilderScopeHandler`, populated during model registration via the existing `$storage->used_traits` direct check (mirrors `ModelPropertyHandler::resolveCasts`).
- Synthesise the macro return types (`restore` -> `int` from `Builder::update`, `restoreOrCreate` / `createOrRestore` -> `TModel` from `firstOrCreate` / `createOrFirst`) and parameter shapes (`array<string, mixed> $attributes = [], array<string, mixed> $values = []`) so `checkMethodArgs` does not crash on the magic-call route.
- Static-on-model forms (`Customer::restoreOrCreate(...)`) keep Psalm's native handling of the trait's `@method static static` declarations and resolve to `Customer&static`.
- The instance method `$model->restore()` continues to return `bool` from the real `SoftDeletes::restore`; the new handler is gated to `Builder` LHS only and never shadows the trait method.

Verified with a new type test (`tests/Type/tests/Builder/SoftDeletesRestoreTest.phpt`) covering:

- Builder-instance forms (`restore` -> `int`, `restoreOrCreate` / `createOrRestore` -> `Customer`).
- Chained `onlyTrashed()->where(...)->restore()` to exercise the LHS template-recovery path.
- Zero-arg `restoreOrCreate()` to validate the synthesised default values.
- Sanity assertion that `$customer->restore()` (instance method) keeps its `bool` return.
- Static-on-model forms (`Customer::restoreOrCreate(...)` -> `Customer&static`).
- Negative case on a non-SoftDeletes model (`Part::query()->restore()` -> `mixed`) to lock in the registry guard.
- Arity check (`Customer::query()->restore('bogus')` -> `TooManyArguments`) to validate the synthesised params.

Known limitations (documented in code comments):

- A model that inherits `SoftDeletes` only through a parent class is not detected (matches the existing `ModelPropertyHandler` convention).
- Models with custom Eloquent builders are not covered yet; the `CustomBuilderMethodHandler` mirror is left for a follow-up.

## Checklist

### DIFF
--- a/src/Handlers/Eloquent/BuilderScopeHandler.php
+++ b/src/Handlers/Eloquent/BuilderScopeHandler.php
@@ -18,11 +18,8 @@ use Psalm\Plugin\EventHandler\MethodParamsProviderInterface;
 use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
 use Psalm\StatementsSource;
 use Psalm\Storage\FunctionLikeParameter;
-use Psalm\Type;
-use Psalm\Type\Atomic\TArray;
 use Psalm\Type\Atomic\TGenericObject;
 use Psalm\Type\Atomic\TNamedObject;
-use Psalm\Type\Atomic\TString;
 use Psalm\Type\Union;
 
 /**
@@ -36,8 +33,8 @@ use Psalm\Type\Union;
  *    (e.g., SoftDeletingScope::extend), and declared via @method static returning
  *    Builder<static> on the model trait.
  * 4. SoftDeletes runtime macros that have no Builder-returning @method declaration:
- *    restore, restoreOrCreate, createOrRestore. Hardcoded in {@see SOFT_DELETES_MACROS},
- *    mirroring Larastan's EloquentBuilderForwardsCallsExtension.
+ *    restore, restoreOrCreate, createOrRestore. Encoded in {@see SoftDeletesMacro},
+ *    which mirrors Larastan's EloquentBuilderForwardsCallsExtension.
  *    See https://github.com/psalm/psalm-plugin-laravel/issues/929.
  * 5. @method PHPDoc scopes are handled natively by Psalm
  *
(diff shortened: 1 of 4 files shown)


alies-dev:
I'm looking for a more ellegant and reusable solution

PR #944, opened 2026-05-16
Cover Laravel packages in `NoEnvOutsideConfig` fallback

### LINKED ISSUE #940: `NoEnvOutsideConfig`: false positives when scanning a Laravel package's own config/ dir without explicit `<configDirectory>`
## Description

When scanning a Laravel **package** (not an app), the `NoEnvOutsideConfig` rule fires on every `env(...)` call inside the package's own `config/*.php`, even though those files are the canonical, documented place to call `env()`.

This is a regression in user experience from #756 → #858: #858 added the `<configDirectory>` opt-in (good), and the fallback was tightened to `ApplicationProvider::getApp()->configPath()` (good for apps). But for **packages** that fallback resolves to the bundled Orchestra Testbench config dir, never the package's own `config/`, so every `env()` call in the package's config files is reported.

The user has to know to add `<configDirectory name="config"/>` to the plugin config. Without it the diagnostic is wrong, not just imprecise.

## Reproduction

Audited [imdhemy/laravel-in-app-purchases](https://github.com/imdhemy/laravel-in-app-purchases) ([branch update-psalm](https://github.com/imdhemy/laravel-in-app-purchases/tree/update-psalm)) with plugin v4.x on Psalm 6.16.1.

`psalm.xml`:

```xml
<projectFiles>
    <directory name="config"/>
    <directory name="routes"/>
    <directory name="src"/>
    <ignoreFiles><directory name="vendor"/></ignoreFiles>
</projectFiles>

<plugins>
    <pluginClass class="Psalm\LaravelPlugin\Plugin"/>
</plugins>
```

(No `<configDirectory>` element inside `<pluginClass>`.)

Output:

```
ERROR: NoEnvOutsideConfig at config/liap.php:33:35
    'google_play_package_name' => env('GOOGLE_PLAY_PACKAGE_NAME', 'com.some.thing'),
ERROR: NoEnvOutsideConfig at config/liap.php:44:28
    'appstore_password' => env('APPSTORE_PASSWORD', ''),
ERROR: NoEnvOutsideConfig at config/liap.php:97:34
    'appstore_private_key_id' => env('APPSTORE_PRIVATE_KEY_ID'),
ERROR: NoEnvOutsideConfig at config/liap.php:99:31
    'appstore_private_key' => env('APPSTORE_PRIVATE_KEY'),
ERROR: NoEnvOutsideConfig at config/liap.php:101:29
    'appstore_issuer_id' => env('APPSTORE_ISSUER_ID'),
ERROR: NoEnvOutsideConfig at config/liap.php:103:29
    'appstore_bundle_id' => env('APPSTORE_BUNDLE_ID'),
```

The reported file `config/liap.php` is the package's published config, merged at runtime via `LiapServiceProvider::mergeConfigFrom(__DIR__.'/../../config/liap.php', 'liap')`. It is a config file by Laravel convention.

## Root cause

`Plugin::initNoEnvOutsideConfigHandler()` (src/Plugin.php:266-275):

```php
$directories = $pluginConfig->configDirectories;

if ($directories === []) {
    $directories = [ApplicationProvider::getApp()->configPath()];
}
```

`ApplicationProvider::getApp()` returns the plugin's internal Orchestra Testbench Laravel app. Its `configPath()` is `vendor/orchestra/testbench-core/laravel/config`, not the analysed package's `config/`. For Laravel **packages** the fallback is therefore guaranteed to miss the package's own config directory.

## Proposed fix

Two options, in increasing order of work:

### Option A (minimal): include the project root's `config/` in the fallback

When `<configDirectory>` is empty, fall back to **both**:

1. `ApplicationProvider::getApp()->configPath()` (current behaviour, useful for projects without a top-level `config/`).
2. `getcwd() . '/config'` if it exists.

`getcwd()` is the project root during Psalm runs; Psalm's own config resolution uses the same anchor. This catches every package and every app that follows the Laravel convention without any user action.

### Option B (more accurate): discover config files from `mergeConfigFrom()` / `publishes()` calls

Scan service providers for `mergeConfigFrom(__FILE_PATH__, ...)` and `publishes([... => config_path(...)], 'config')`, resolve `__DIR__`-relative paths at scan time, and add the parent directories to the configDirectory list. This matches the actual semantic ("file is a Laravel config file") rather than the conventional path layout. It also handles non-standard layouts (e.g. `resources/config/`).

Option B is more correct but requires a scanner pass; Option A is a 3-line change in `Plugin.php`.

### Either way

Detect the "package, no `<configDirectory>` set, fallback resolved to zero project-relevant dirs" case explicitly and emit a one-line `$progress->info()` pointing to the [`configDirectory` docs](https://psalm.github.io/psalm-plugin-laravel/config/#configdirectory). The current warning only fires when **all** inputs fail to resolve, which never happens for the testbench fallback.

## Impact

Affects every Laravel package analysed with the plugin that ships a `config/*.php` and does not know to set `<configDirectory>`. In one-shot audits across `spatie/*`, `imdhemy/*`, `laravel-notification-channels/*` packages, this is a consistent stream of false positives that pushes maintainers to suppress the rule entirely rather than fix the (real) `env()` calls outside config.

Documentation alone is not enough here — the fallback path encodes "this is a Laravel app", but the plugin is also widely used on Laravel packages. The default should not silently produce wrong results on the second case.

## Workaround (for users hitting this today)

Add to `psalm.xml`:

```xml
    <configDirectory name="config"/>
</pluginClass>
```

alies-dev:
## Issue to Solve

When scanning a Laravel **package** (not an app), `NoEnvOutsideConfig` fired on every `env(...)` call inside the package's own `config/*.php`. `ApplicationProvider::getApp()` falls back to the Orchestra Testbench app whose `configPath()` lives under `vendor/orchestra/.../config`, so the package's own `config/` was never matched and the rule reported every canonical `env()` call.

## Related

Fixes #940

## Solution Description

Add `getcwd()/config` as a second fallback anchor alongside the Testbench `configPath()` in `Plugin::initNoEnvOutsideConfigHandler`. `NoEnvOutsideConfigHandler::init()` already drops non-existent entries via `realpath()` + `is_dir()`, so listing both unconditionally is safe.

When the fallback is used and `getcwd()/config` does not exist, emit a one-shot `Progress::warning()` naming the `NoEnvOutsideConfig` issue and pointing users at the `<configDirectory>` opt-in. Non-standard layouts then surface explicitly instead of silently misreporting.

Behaviour for projects that already declare `<configDirectory>` is unchanged: the user-supplied list still replaces the default, and the existing typo-case warning (non-empty input resolving to zero dirs) is preserved.

Verified by a new unit test `Tests\Psalm\LaravelPlugin\Unit\PluginNoEnvFallbackTest` that exercises:
 - package fallback skips `cwd/config/*.php`,
 - warning fires once when `cwd` has no `config/`,
 - warning stays silent when the user has set `<configDirectory>`.

`docs/config.md` updated to document the new `cwd/config` anchor and the two distinct warning cases.

## Checklist

- [x] Tests cover the change (unit test in `tests/Unit/PluginNoEnvFallbackTest.php`)
- [x] Documentation is updated (`docs/config.md`)

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -67,7 +67,7 @@ Disable if dynamic where resolution conflicts with your codebase.
 
 ## `configDirectory`
 
-**default**: the booted Laravel app's `config_path()`
+**default**: the booted Laravel app's `config_path()` plus the project root's `config/` directory (if it exists)
 
 Controls which directories are treated as config directories by [`NoEnvOutsideConfig`](issues/NoEnvOutsideConfig.md). `env()` calls inside any of these directories are exempt from the check.
 
@@ -75,7 +75,9 @@ Each entry can be an absolute path or a relative path resolved by PHP's `glob()`
 
 **Defining any `<configDirectory>` replaces the default**, so include `config` (or whatever your project's standard config dir is) explicitly if you still want it covered. Test files (paths containing `/tests/`) are always exempt regardless of this setting.
 
-If no entry resolves to an existing directory at boot, the plugin emits a warning so the typo case (`<configDirectory name="cofnig" />`) is surfaced rather than silently flagging every `env()` call.
+The project-root anchor exists primarily for Laravel **packages**, where the plugin's booted-app `config_path()` falls back to Orchestra Testbench's vendored config dir and would never match the package's own `config/*.php`. If the project root has no `config/` directory and `<configDirectory>` is not configured, the plugin emits a one-shot warning at boot pointing here.
+
+If a user-provided `<configDirectory>` resolves to zero existing directories at boot, the plugin emits a separate warning so the typo case (`<configDirectory name="cofnig" />`) is surfaced rather than silently flagging every `env()` call.
 
 ### Examples
 
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -258,11 +258,28 @@ final class Plugin implements PluginEntryPointInterface
      * Resolve config directory paths and pass them to NoEnvOutsideConfigHandler.
      *
      * When the user has not configured any `<configDirectory>` elements, fall back to
-     * the booted Laravel app's `config_path()`. Relative paths and glob patterns are
-     * left as-is — the handler resolves them via glob() + realpath() at this boot step.
+     * BOTH the booted Laravel app's `config_path()` AND the project root's `config/`.
+     * Two anchors are needed because the plugin is used in two different shapes:
      *
-     * The handler emits a warning via $output if the resolution produces zero directories
-     * for a non-empty input — that's the typo case where every env() call would be flagged.
+     *   - App analysis: `ApplicationProvider::getApp()->configPath()` resolves to the
+     *     analysed app's own `config/`. This is the original #858 behaviour.
+     *   - Package analysis: `getApp()` falls back to Orchestra Testbench's vendored
+     *     Laravel app, whose `configPath()` points into `vendor/orchestra/.../config`
+     *     and never matches the package's own config files. Adding `getcwd()/config`
+     *     covers the package case without requiring `<configDirectory>` opt-in.
+     *
+     * Relative paths and glob patterns are left as-is — the handler resolves them via
+     * glob() + realpath() at boot. Non-existent entries are dropped silently there,
+     * so listing both anchors unconditionally is safe.
+     *
+     * Diagnostics:
+     *   - If neither anchor is a project-relevant directory (no `config/` at `getcwd()`),
+     *     emit a warning pointing to the `<configDirectory>` docs so users with a
+     *     non-standard layout know how to opt in.
+     *   - The handler itself emits a separate warning if a *non-empty* user input
+     *     resolves to zero directories (the typo case).
+     *
+     * @see https://github.com/psalm/psalm-plugin-laravel/issues/940
      */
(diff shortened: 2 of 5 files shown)


alies-dev:
Closing as redundant after #949 (0d3217b2) merged.

#949 fixed root cause: `ApplicationProvider::doGetApp()` now calls `$app->useConfigPath($projectRoot.'/config')` post-creation, with `$projectRoot` resolved from `APP_BASE_PATH` / `TESTBENCH_APP_BASE_PATH` env or `getcwd()` when `composer.json` is present. `ApplicationProvider::getApp()->configPath()` therefore already returns the analysed package's own `config/` directory, which is exactly the path `Plugin::initNoEnvOutsideConfigHandler` feeds into `NoEnvOutsideConfigHandler::init()`.

What this PR added on top of that:

1. `getcwd().'/config'` as a second fallback anchor. Duplicates #949's anchor in the normal case (composer.json present at cwd). The remaining edge case (no composer.json at cwd, no env override) is not a Laravel-package layout to begin with, and the dir would not exist either way.
2. A one-shot `Progress::warning()` pointing users at the `<configDirectory>` opt-in when the fallback dir is missing. Genuine DX win, not covered by #949, but small enough to do as a follow-up.
3. Shared `RecordingProgress` test spy and `PluginNoEnvFallbackTest`. Useful scaffolding regardless of the path-fallback change.

Will re-land items 2 and 3 separately if needed.

Issue #787, opened 2026-04-18
App-owned Facade with string-alias accessor emits UndefinedMagicMethod


alies-dev:
## Description

An application's own `Facade` subclass with a **string-alias accessor** (e.g. `return 'License'`) and an **incomplete `@method` docblock catalogue** emits `UndefinedMagicMethod` on any call whose method is not enumerated in phpdoc. Neither of the two existing discovery paths in the plugin recovers it:

1. `FacadeMapProvider` — the facade is consumed by FQN, not via a class alias, so it never appears in `AliasLoader::getInstance()->getAliases()`.
2. A hypothetical "scan every `Facade` subclass in scope" fallback (proposed in #761) would still fail here: resolving `getFacadeRoot()` for accessor `'License'` requires `ApplicationProvider`'s Testbench app to run the analysed project's service providers — it does not — so `BindingResolutionException` would be caught and the mapping skipped.

This is the **intersection** of #761 (facade discovery) and #766 (string-alias container resolution): a first-party facade whose accessor is a string binding registered by the analysed project's own service provider. Neither fix in isolation would address this repro.

## Reproduction

Audited [koel/koel](https://github.com/koel/koel) with the plugin (dev branch, Psalm 7 beta, `errorLevel=2`).

`app/Facades/License.php`:

```php
<?php

namespace App\Facades;

use Illuminate\Support\Facades\Facade;

/**
 * @method static bool isPlus()
 * @method static bool isCommunity()
 * @see \App\Services\LicenseService
 */
class License extends Facade
{
    protected static function getFacadeAccessor(): string
    {
        return 'License';
    }
}
```

`app/Providers/LicenseServiceProvider.php`:

```php
public function register(): void
{
    app()->singleton('License', static fn (): LicenseServiceInterface => app(LicenseServiceInterface::class));
}
```

`app/Services/LicenseService.php` (excerpt):

```php
public function getStatus(bool $checkCache = true): LicenseStatus { ... }
```

Call site — `app/Console/Commands/DoctorCommand.php:112`:

```php
$status = License::getStatus(checkCache: false);
```

Psalm output:

```
ERROR: UndefinedMagicMethod - app/Console/Commands/DoctorCommand.php:112:23
  Magic method App\Facades\License::getstatus does not exist (see https://psalm.dev/219)
      $status = License::getStatus(checkCache: false);
```

[...]

PR #946, opened 2026-05-16
Defer facade scan-phase warnings off Psalm progress counter

### LINKED ISSUE #941: Facade resolution warnings wedge into scan progress counter
## Symptom

Running Psalm with the plugin against [imdhemy/laravel-in-app-purchases](https://github.com/imdhemy/laravel-in-app-purchases) produces malformed progress output:

```
Scanning files...

2743 / 2743...Warning: Laravel plugin: getFacadeRoot() failed for 'Imdhemy\Purchases\Facades\Subscription': Target class [subscription] does not exist.


Analyzing files...
```

The first warning is glued onto the `2743 / 2743...` counter line. The warnings themselves are legitimate (third-party facades with provider-bound accessors and no `@see` tag, surfaced by the resolver added in #789) — only the rendering is broken.

## Root cause

`Psalm\Progress\LongProgress::taskDone()` writes `\rN / total...` without a trailing newline. `AppFacadeRegistrationHandler::tryGetFacadeRootClass()` is invoked from `afterClassLikeVisit`, which fires per-class during the scan phase. When `getFacadeRoot()` throws, the handler calls `\$progress->warning(\$msg)` immediately — `Progress::warning()` writes `'Warning: ' . \$message . PHP_EOL`, with no leading newline. The warning text therefore appends directly to whatever scan-progress fragment is currently on stderr.

Source: [\`src/Handlers/Facades/AppFacadeRegistrationHandler.php#L222\`](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Handlers/Facades/AppFacadeRegistrationHandler.php#L222).

## Desired output

Warnings should render cleanly between the scan and analyze sections:

```
2743 / 2743...

Analyzing files...
```

## Proposed fix

Defer scan-phase warnings into a static buffer, flush at the top of `afterCodebasePopulated()` (which fires between scan and analyze):

- Add `private static array \$pendingScanWarnings = []` and `private static bool \$wroteScanSeparator = false`.
- Extend `tryGetFacadeRootClass()` with a `bool \$deferWarnings = false` parameter. `afterClassLikeVisit` passes `true`; `afterCodebasePopulated` passes `false`.
- Route every emission through a helper `emitWarning(Progress, string)` that lazily writes a single `PHP_EOL` before the first message (terminating the open `\rN / N...` line).
- `flushScanWarnings(Progress)` drains the buffer at the top of `afterCodebasePopulated`. Populate-phase warnings (e.g. autoload-failed at lines 135-145) also route through `emitWarning` so the separator is consistent regardless of which emit site fires first.

Lazy separator keeps zero-warning runs unchanged (no stray blank line).

## Caveat: threaded scan

`afterClassLikeVisit` may run inside scan worker processes when `--threads N > 1`. Worker-local statics don't propagate back to the main process, so deferred warnings would buffer in workers and be lost on worker exit. Options:

1. Accept the loss (single-process scan is the default and matches the user's repro here).
2. `register_shutdown_function` in each worker to flush before exit.
3. Detect worker context and skip deferral there (fall back to current direct emit; rendering will still be wedged in threaded mode but no information is lost).

Open question: does Psalm fork `afterClassLikeVisit` handlers into workers, or are they main-process only? Worth verifying before picking an option.

## Repro

1. `git clone https://github.com/imdhemy/laravel-in-app-purchases /tmp/iap && cd /tmp/iap && composer install`
2. Install the plugin (any version with #789 merged).
3. `./vendor/bin/psalm` — observe wedged warning on the scan counter line.

The two offending facades:
- `Imdhemy\Purchases\Facades\Subscription` (accessor `'subscription'`)
- `Imdhemy\Purchases\Facades\Product` (accessor `'product'`)

Both have `@method` catalogues but no `@see` tag, and their accessors are bound by `Imdhemy\Purchases\PurchasesServiceProvider` which doesn't run in the plugin's Testbench application.

## Related

- #787 — original facade method resolution issue (closed by #789).
- #761 — facade discovery beyond `AliasLoader`.
- #766 — string-alias container resolution outside Testbench.

alies-dev:
## Issue to Solve

Running Psalm with the plugin against codebases that contain third-party facades whose accessors are bound by service providers Testbench doesn't run (e.g. [imdhemy/laravel-in-app-purchases](https://github.com/imdhemy/laravel-in-app-purchases)) produces malformed progress output:

```
Scanning files...

2743 / 2743...Warning: Laravel plugin: getFacadeRoot() failed for 'Imdhemy\Purchases\Facades\Subscription': Target class [subscription] does not exist.


Analyzing files...
```

The warnings themselves are legitimate (third-party facades with provider-bound accessors and no `@see` tag, surfaced by the resolver added in #789). Only the rendering is broken: `LongProgress::taskDone()` writes `\rN / total...` with no trailing newline, and `Progress::warning()` writes `'Warning: ' . $message . PHP_EOL` with no leading newline, so the warning glues directly onto the open counter line.

## Related

Closes #941. Builds on #787 / #789 (the facade resolver that surfaces these warnings).

## Solution Description

Buffer scan-phase warnings into a static array, drain them at the top of `afterCodebasePopulated()` (runs between scan and analyze phases). Lazy `PHP_EOL` separator before the first emission terminates Psalm's open counter line; zero-warning runs leave output untouched.

Key design points:

- `tryGetFacadeRootClass()` gains a `bool $deferWarnings = false` parameter. `afterClassLikeVisit` passes `true`, `afterCodebasePopulated` passes `false`.
- All emission sites in the handler (the three branches of `tryGetFacadeRootClass`, plus the populate-phase autoload checks) now route through a single private `emitWarning(Progress, string, bool $defer)` helper, so the lazy separator stays consistent.
- `flushScanWarnings()` re-arms the separator gate at every call, so a second `checkPaths()` re-entry (CLI re-analysis, daemon/LSP loops) gets a fresh separator instead of splicing into the new scan-counter line.
- Forked scan workers (`--threads N > 1`) inherit the buffer via copy-on-write but their mutations never reach the main process. A `register_shutdown_function` fallback prints buffered warnings directly to STDERR before the worker exits, so threaded mode degrades to "wedged but visible" instead of "silently lost". The shutdown drain also accepts an optional `$stream` test seam (defaults to STDERR) so it can be exercised in a unit test.

### Verification

Nine unit tests in `tests/Unit/Handlers/Facades/AppFacadeWarningBufferTest.php`:

- `deferred_warning_is_not_written_until_flush` and `flush_emits_separator_then_warning` cover the core defer / drain lifecycle.
- `direct_emission_writes_separator_once_for_multiple_warnings` and `deferred_then_direct_share_the_same_separator` cover the lazy separator across multiple emission paths.
- `flush_with_empty_buffer_writes_nothing` guards the zero-warning case.
- `flush_rearms_separator_gate_for_next_run` covers the cross-run reset.
- `throwable_from_get_facade_root_emits_failure_warning` covers the `Throwable` catch branch via a `ThrowingFacadeFixture` whose accessor throws.
- `repeated_probe_of_same_facade_buffers_only_once` asserts the `failedFacades` short-circuit prevents duplicate buffer entries.
- `shutdown_drain_writes_buffered_warnings_to_stream` covers the worker-shutdown fallback via the `$stream` test seam.

All existing type/unit/psalm/cs/rector checks pass.

## Checklist

- [x] Tests cover the change (9 new unit tests in `tests/Unit/Handlers/Facades/AppFacadeWarningBufferTest.php`)

### DIFF
--- a/src/Handlers/Facades/AppFacadeRegistrationHandler.php
+++ b/src/Handlers/Facades/AppFacadeRegistrationHandler.php
@@ -56,6 +56,36 @@ final class AppFacadeRegistrationHandler implements AfterClassLikeVisitInterface
     private static array $failedFacades = [];
 
     /**
+     * Scan-phase warnings deferred until {@see self::afterCodebasePopulated()} so they do
+     * not splice into Psalm's open `\rN / total...` progress counter (issue #941).
+     * `LongProgress::taskDone()` writes the counter without a trailing newline; emitting a
+     * warning from `afterClassLikeVisit` would otherwise glue `Warning: ...` onto the same
+     * line. Flushed at the top of `afterCodebasePopulated`, between the scan and analysis
+     * phases, where the output buffer is between progress writes.
+     *
+     * @var list<string>
+     */
+    private static array $pendingScanWarnings = [];
+
+    /**
+     * Lazy separator gate: the first warning emission writes a single `PHP_EOL` to terminate
+     * an open `\rN / total...` line. Zero-warning runs leave output untouched (no stray blank
+     * line). Persists for the lifetime of the process — additional warnings after the first
+     * already start at column 0.
+     */
+    private static bool $wroteScanSeparator = false;
+
+    /**
+     * Guards a one-time `register_shutdown_function` registration. Scan workers
+     * (`--threads N > 1`) fork the main process; the worker's copy of
+     * {@see self::$pendingScanWarnings} never propagates back, so without a shutdown hook a
+     * buffered warning would vanish when the worker exits. The hook flushes pending warnings
+     * directly to STDERR at worker exit; in the main process the same hook fires after
+     * `afterCodebasePopulated` already drained the buffer, so it's a no-op there.
+     */
+    private static bool $shutdownHookRegistered = false;
+
+    /**
      * Probe `Facade::getFacadeRoot()` at scan time and queue the resolved root class for
      * scanning. We can't do this in {@see self::afterCodebasePopulated()} because by then
      * the scanner has stopped — a root class not already pulled in via other references
@@ -87,7 +117,7 @@ final class AppFacadeRegistrationHandler implements AfterClassLikeVisitInterface
         }
 
         $progress = $event->getCodebase()->progress;
-        $rootClass = self::tryGetFacadeRootClass($storage->name, $progress);
+        $rootClass = self::tryGetFacadeRootClass($storage->name, $progress, true);
 
         if ($rootClass === null) {
             return;
@@ -101,6 +131,12 @@ final class AppFacadeRegistrationHandler implements AfterClassLikeVisitInterface
     {
         $codebase = $event->getCodebase();
         $progress = $codebase->progress;
+
+        // Drain warnings that the scan-phase hook deferred. Runs first so the lazy
+        // separator terminates Psalm's open `\rN / total...` progress line (issue #941)
+        // before any populate-phase warning could emit its own separator. Also re-arms
+        // the separator gate so a second analysis in the same process re-emits it.
+        self::flushScanWarnings($progress);
 
         foreach ($codebase->classlike_storage_provider::getAll() as $storage) {
             // Abstract classes include the base Facade itself and user-defined abstract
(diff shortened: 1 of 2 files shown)


alies-dev:
closing as too much code for a simple task. I think it should be solved upstream instead

PR #948, opened 2026-05-16
Suppress `PossiblyUnusedMethod` on container-bound `__construct`

### LINKED ISSUE #943: `PossiblyUnusedMethod` on `__construct` of container-bound classes, console commands, and service providers
## Description

`PossiblyUnusedMethod` fires on the `__construct` of classes that are only instantiated through the Laravel container, even though every such instantiation goes through reflection on `__construct` to resolve dependencies. The constructor is the contract Laravel uses to build the object; flagging it as unused is wrong.

Three call sites trigger this today:

1. **Concrete classes on the RHS of `bind()` / `singleton()` / `scoped()` / `instance()`** in a service provider.
2. **Console commands registered via `$this->commands([...])`** in a service provider.
3. **Service providers registered via `$this->app->register(...)`** or `extra.laravel.providers` in `composer.json`.

## Reproduction

Audited [imdhemy/laravel-in-app-purchases](https://github.com/imdhemy/laravel-in-app-purchases) ([branch update-psalm](https://github.com/imdhemy/laravel-in-app-purchases/tree/update-psalm)) with plugin v4.x on Psalm 6.16.1.

`LiapServiceProvider::bindConcretes()`:

```php
$this->app->bind(UrlGeneratorContract::class, UrlGenerator::class);
$this->app->bind(EventFactoryContract::class, EventFactory::class);
$this->app->bind(HandlerHelpersInterface::class, HandlerHelpers::class);
$this->app->bind(JwsServiceInterface::class, JwsService::class);
```

`LiapServiceProvider::bootCommands()`:

```php
$this->commands([
    LiapConfigPublishCommand::class,
    LiapUrlCommand::class,
    RequestTestNotificationCommand::class,
]);
```

`LiapServiceProvider::registerEvents()`:

```php
$this->app->register(EventServiceProvider::class);
```

Plus a runtime resolution of `HandlerFactory` via the container in `ServerNotificationController::__invoke(HandlerFactory $handlerFactory, ...)`.

Psalm output:

```
PossiblyUnusedMethod  src/Console/RequestTestNotificationCommand.php:24   __construct(AppStoreTestBuilder $testBuilder)
PossiblyUnusedMethod  src/Console/UrlGenerator.php:18                     __construct(LaravelUrlGenerator $urlGenerator)
PossiblyUnusedMethod  src/Handlers/AppStoreV2NotificationHandler.php:21   __construct(HandlerHelpersInterface, JwsServiceInterface)
PossiblyUnusedMethod  src/Handlers/HandlerFactory.php:23                  __construct(Application, Request)
PossiblyUnusedMethod  src/Handlers/HandlerHelpers.php:23                  __construct(...)
PossiblyUnusedMethod  src/Handlers/JwsService.php:22                      __construct(JwsParser, JwsVerifier, Request)
PossiblyUnusedMethod  src/ServiceProviders/EventServiceProvider.php:16    __construct(Application $app)
```

Seven `PossiblyUnusedMethod` reports, all on classes whose only call site is the container.

## Why the existing precedent applies

The plugin already does this exact kind of suppression for Mailable subclasses via `SuppressHandler::METHOD_LEVEL_BY_PARENT_CLASS` (src/Handlers/SuppressHandler.php:148-170):

```php
// __construct included because Mailable subclasses typically have their `new` call
// sites only inside controller actions / service methods. Those enclosing methods
// are themselves only invoked by Laravel through reflection (router, container),
// so Psalm marks them unreachable from any visible entry point — and `new MyMail()`
// sitting inside them inherits that unreachability, leaving `__construct` reported
// as `PossiblyUnusedMethod`. Verified against IxDF's real codebase.
```

The same reasoning is even stronger for container-bound classes: the call site is literally inside Laravel's `Container::build()`, never appears as a `new` in user code, and is statically discoverable from `bind()` / `singleton()` arguments. The plugin already scans these call sites (see `OctaneIncompatibleBindingHandler` for an `AfterMethodCallAnalysisInterface` precedent), so the wiring is mostly in place.

## Proposed fix

Add a `ClassLikeStorageProvider`-driven suppression similar to the Mailable hook, populated by:

1. **`AfterMethodCallAnalysisInterface`** on `Application::bind/singleton/scoped/instance/bindIf/singletonIf/scopedIf` and on `Container::bind/singleton/...`. When the second argument is a `ClassConstFetch` to `::class` (or a string literal FQCN), record the target class as "container-bound" and mark its `__construct` as suppressed for `PossiblyUnusedMethod`. Same for first argument when it is a concrete class (single-arg form: `bind(Concrete::class)`).
2. **`AfterMethodCallAnalysisInterface`** on `ServiceProvider::commands([...])` — every `::class` constant inside the array literal is a console command.
3. **`AfterMethodCallAnalysisInterface`** on `Application::register(Provider::class)` plus a one-shot scan of `composer.json`'s `extra.laravel.providers` and `extra.laravel.dont-discover` at plugin boot.
4. **`METHOD_LEVEL_BY_PARENT_CLASS`** fallback: extend with `Illuminate\Support\ServiceProvider => ['__construct']` (registered via `$this->app->register()` always go through reflection; same Mailable-style reasoning).

Item 4 alone kills the EventServiceProvider case and is a 1-line change; items 1–3 require the scanner pass but cover the general case.

## Impact

Affects every Laravel package and every Laravel app that uses dependency injection — i.e. most of them. The seven false positives above are from one small package; mid-sized apps with rich service providers see dozens.

The current state forces maintainers to either:
- Add `@psalm-api` to every container-bound class (noisy, requires knowing the rule),
- Baseline these (hides genuine regressions),
- Or disable `PossiblyUnusedMethod` entirely (loses the signal for genuinely dead code).

A targeted suppression based on container bindings keeps `PossiblyUnusedMethod` useful for genuinely dead code while not flagging the framework's primary instantiation path.

## Related

- #756, #858 — same pattern of "the default behaviour encodes a Laravel app assumption that doesn't hold for packages or non-standard setups".

alies-dev:
## Issue to Solve

`PossiblyUnusedMethod` fires on the `__construct` of every class that Laravel instantiates through the container, even though every such instantiation reflects on `__construct` to inject dependencies. The constructor is the contract Laravel uses to build the object, so flagging it as unused is wrong.

Three call sites trigger the noise today:

1. **Concrete classes** on the RHS of `bind()` / `singleton()` / `scoped()` / `instance()` / `bindIf()` / `singletonIf()` / `scopedIf()` in a service provider.
2. **Console commands** registered via `$this->commands([...])`.
3. **Service providers** registered via `$this->app->register(...)` or `extra.laravel.providers` in `composer.json`.

A mid-sized application's service providers fan this out to dozens of false positives, forcing maintainers to add `@psalm-api` to every container-bound class, baseline the issue, or disable `PossiblyUnusedMethod` entirely.

## Related

Closes #943.

Builds on #947 (which stopped `PropertyNotSetInConstructor` from firing on `ServiceProvider::$app` for subclasses that override `__construct`).

## Solution Description

Two complementary mechanisms.

### Parent-class suppression for framework-discovered classes

`SuppressHandler::METHOD_LEVEL_BY_PARENT_CLASS` gains `__construct` for `Illuminate\Console\Command` and `Illuminate\Support\ServiceProvider`. These are framework-discovered at runtime (`Console\Kernel::resolve()`, `Application::resolveProvider()`, `extra.laravel.providers`), so a blanket parent-class suppression is the right shape. The existing `suppressFrameworkHookMethod()` visibility filter keeps non-public constructors flagged (Container reflection would fail at runtime anyway).

### Synthetic call-graph reference for explicit container bindings

A new `ContainerBoundConstructorHandler` listens on `AfterMethodCallAnalysis`. When it sees:

- `Container::{bind,bindIf,singleton,singletonIf,scoped,scopedIf}(X::class, Concrete::class)` — two-arg form
- `Container::{bind,bindIf,singleton,singletonIf,scoped,scopedIf}(Concrete::class)` — single-arg form
- `ContextualBindingBuilder::give(Concrete::class)` — fluent contextual binding
- `ContextualBindingBuilder::give([C1::class, C2::class])` — variadic contextual binding (Laravel's [documented typed-variadic pattern](https://laravel.com/docs/master/container#binding-typed-variadics))

it records a synthetic reference to `Concrete::__construct` via `FileReferenceProvider::addMethodReferenceToClassMember()`. The synthetic edge piggybacks on Psalm's existing `ShutdownAnalyzerTask` aggregation, so it survives fork-pool workers and reaches the parent process before `checkMethodReferences()` runs.

Out of scope by design:
- **`instance()`** receives a pre-built object whose `new` site Psalm already tracks.
- **`register()`** always targets a `ServiceProvider`, already covered by the parent-class rule.
- **Closure-as-second-arg** and **array-with-closure** forms — Psalm sees the inner `new` naturally.

A public-constructor check (`ClassLikeStorageProvider::get()` + `MethodStorage::$visibility`) guards the synthetic write, so a `private` or `protected` `__construct` on a bound class still surfaces as a real bug — Container reflection via `ReflectionClass::newInstanceArgs()` would throw at runtime.

The new handler is registered alongside `ContainerHandler` / `OffsetHandler` in `Plugin::registerHandlers()`, following the project's `require_once` pattern for psalm.phar compatibility.

### Verification

- **Type tests**: extended `tests/Type/tests/ServiceProviderTest.phpt` with a `BindingServiceProvider` whose `register()` body exercises every binding shape (two-arg, single-arg, closure, `commands([])`, nested `register()`, single `give()`, variadic `give([])`), plus an `__construct(Application $app)` override demonstrating the Item-4 parent-class suppression. `composer test:type` passes (401 tests).
- **Empirical end-to-end check**: a throwaway Laravel project with `findUnusedCode="true"` confirmed:
  - Before the patch: `PossiblyUnusedMethod` on every bound concrete's `__construct`.
  - After the patch: silenced for public constructors across all the shapes above.
  - Protected `__construct` on a bound class is still flagged.
  - `--threads=4` reproduces the suppression (worker aggregation works).
- **Self-analysis + style**: `composer psalm`, `composer cs`, `composer test:unit` all clean.

## Checklist

### DIFF
--- a/src/Handlers/Application/ContainerBoundConstructorHandler.php
+++ b/src/Handlers/Application/ContainerBoundConstructorHandler.php
@@ -1,0 +1,306 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Application;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
+use Psalm\Plugin\EventHandler\AfterMethodCallAnalysisInterface;
+use Psalm\Plugin\EventHandler\Event\AfterMethodCallAnalysisEvent;
+
+/**
+ * Records a synthetic call-graph reference to the `__construct` of every concrete
+ * class registered through Laravel's container binding methods.
+ *
+ * Why: Laravel resolves container-bound classes through `Container::build()`,
+ * which reflects on `__construct`. The class name only appears as a string FQCN
+ * argument to `bind()`/`singleton()`/`scoped()` etc., never as `new Concrete(...)`
+ * in user code, so Psalm's reference graph never picks up the constructor and
+ * reports `PossiblyUnusedMethod`. Adding a synthetic
+ * {@see \Psalm\Internal\Provider\FileReferenceProvider::addMethodReferenceToClassMember()}
+ * entry from the calling site (typically `ServiceProvider::register`) lets the
+ * existing unused-method check short-circuit naturally. See psalm/psalm-plugin-laravel#943.
+ *
+ * Scope:
+ *  - Two-arg form: `bind(Interface::class, Concrete::class)` — Concrete recorded.
+ *  - Single-arg form: `bind(Concrete::class)` — Concrete recorded.
+ *  - Contextual fluent form: `$app->when(X)->needs(Y)->give(Concrete::class)` —
+ *    Concrete recorded via ContextualBindingBuilder::give.
+ *  - Variadic contextual form: `give([Concrete1::class, Concrete2::class])` —
+ *    each class-string item recorded. The Laravel docs document this as the
+ *    standard pattern for typed-variadic injection, so the array branch is
+ *    only enabled for `give()`; `bind()` is `Closure|string|null` (Laravel
+ *    even type-errors on arrays at Container.php:380), so no array branch there.
+ *  - Closure / instance second-arg forms are skipped: the closure body or the
+ *    pre-built instance already contains the `new` expression that Psalm tracks
+ *    naturally, so no synthetic reference is needed.
+ *  - `instance()` and `register()` are intentionally NOT here: `instance()`
+ *    receives a pre-built object whose construction site is visible to Psalm,
+ *    and `register()` always targets a ServiceProvider whose `__construct` is
+ *    already suppressed in SuppressHandler via METHOD_LEVEL_BY_PARENT_CLASS.
+ *
+ * Method-id gate covers both the concrete Container and the contract, matching
+ * the dispatch behaviour of {@see OctaneIncompatibleBindingHandler}: the
+ * `declaring_method_id` points at whichever one is in scope for the receiver.
+ */
+final class ContainerBoundConstructorHandler implements AfterMethodCallAnalysisInterface
+{
+    /**
+     * Method names this handler reacts to. Keys are lowercase.
+     *
+     * `instance` and `register` are intentionally absent — see the class docblock.
+     * `give` is included to cover contextual bindings:
+     * `$this->app->when(X::class)->needs(Y::class)->give(Concrete::class)` reaches
+     * Concrete's `__construct` through the same `Container::build()` reflection path.
+     */
+    private const BINDING_METHOD_NAMES = [
+        'bind' => true,
+        'bindif' => true,
+        'singleton' => true,
+        'singletonif' => true,
+        'scoped' => true,
+        'scopedif' => true,
+        'give' => true,
+    ];
+
+    /**
+     * Declaring method ids (lowercase) that count as "container binding". Covers both
+     * the concrete container and the contract; the user's receiver type determines
+     * which the declaring_method_id resolves to.
+     *
+     * Includes `ContextualBindingBuilder::give` (contract + concrete) because the
+     * fluent `$app->when(...)->needs(...)->give(Concrete::class)` chain hands a
+     * concrete class-string to `Container::build()` exactly like the direct
+     * `bind(...)` family does.
+     */
+    private const CONTAINER_METHOD_IDS_LOWER = [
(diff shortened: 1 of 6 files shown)


alies-dev:
it's out of the focus on upcoming package versions, closing for now, probably will back to this topic later

Issue #766, opened 2026-04-17
`app('package.alias')` returns mixed — string aliases registered by the analysed package's service provider are unknown to the Testbench app


alies-dev:
## Description

When the first argument to `app()` / `resolve()` / `make()` is a **string identifier** (not a class name) that is bound to a concrete class by one of the analysed package's own service providers, the plugin falls back to `mixed`. This cascades into `MixedAssignment`, `MixedMethodCall`, and downstream errors at every call site.

This is distinct from:
- **#757** — `app(Foo::class)` with a literal class-string argument. That issue's proposed `class_exists($abstract)` fallback would not catch plain string keys like `'datatables.request'`.
- **#761** — third-party facades missed by `FacadeMapProvider`. Conceptually related (same root: Testbench app ignorant of the analysed package's providers), but this issue is about the `app()` helper / container resolution, not facade magic methods.

## Root cause

`ContainerResolver::resolveFromApplicationContainer()` ([`src/Util/ContainerResolver.php:27-52`](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Util/ContainerResolver.php#L27-L52)) calls `ApplicationProvider::getApp()->make($abstract)` against the plugin's internal Orchestra Testbench app. That app only runs Laravel framework providers. It never runs the analysed package's own providers, so any string alias the package registers (via `$this->app->singleton('my.alias', ...)`, `$this->app->bind('my.alias', ...)`, `$this->app->alias('my.alias', Foo::class)`, etc.) throws `BindingResolutionException` inside `make()`. The resolver swallows the throwable, returns `null`, and the call site receives `app()`'s documented `mixed` return.

For a literal class argument, `resolveFromLiteralString()` has a `class_exists()` check as a last resort (if #757 is implemented). For a string alias like `'datatables.request'`, no such fallback is possible without knowing the binding — the string is not a loadable class name.

## Reproduction

Audited [yajra/laravel-datatables](https://github.com/yajra/laravel-datatables) with the plugin (v4.8.3, Psalm 7 beta, `errorLevel=1`).

`DataTablesServiceProvider::register()`:

```php
// src/DataTablesServiceProvider.php:27-31
$this->app->singleton('datatables', fn () => new DataTables);
$this->app->singleton('datatables.request', fn () => new Request);
$this->app->singleton('datatables.config', DataTablesConfig::class);
```

Call sites:

```php
// src/QueryDataTable.php:99-107
[...]

PR #955, opened 2026-05-17
Resolve vendor facades via container introspection (replaces #951)

### LINKED ISSUE #942: Resolve in-package facades via static parsing of vendor service providers
## Problem

Third-party packages ship `Facade` subclasses whose accessor is a string alias bound by the package's own `ServiceProvider::register()`. After #789, two resolver paths exist:

1. **`@see` / `@mixin` docblock pointer** — fails when the package author didn't write one.
2. **Runtime `Facade::getFacadeRoot()` probe** — fails because the package's service provider doesn't run inside the plugin's Testbench app, so the accessor isn't bound. `BindingResolutionException` fires and a warning is emitted.

When both paths fail, the plugin falls back to the facade's `@method` catalogue. If the catalogue is complete the user sees no `UndefinedMagicMethod`, but they do see noisy probe-failure warnings on every run, and the resolver gives up on deriving types from the actual service class (which would supersede the hand-maintained catalogue).

## Repro

[imdhemy/laravel-in-app-purchases](https://github.com/imdhemy/laravel-in-app-purchases) ships two such facades:

```php
namespace Imdhemy\Purchases\Facades;

/**
 * @method static \Imdhemy\Purchases\Subscription googlePlay(?ClientInterface $client = null)
 * @method static \Imdhemy\Purchases\Subscription appStore(?ClientInterface $client = null)
 */
class Subscription extends Facade
{
    protected static function getFacadeAccessor() { return 'subscription'; }
}
```

`'subscription'` and `'product'` are bound in `Imdhemy\Purchases\PurchasesServiceProvider::registerSubscriptionClient()` / `registerProductClient()` (or equivalent). That provider never runs in Testbench, so `Subscription::getFacadeRoot()` throws `Target class [subscription] does not exist`.

Result on `vendor/bin/psalm`:

```
```

(Rendering of the warning itself is broken — see #941. This issue is about the resolver, not the print path.)

Same pattern in many ecosystem packages: anything that ships its own facade and binds the accessor by string alias in its provider (Livewire's `Livewire::`, Filament's `Filament::`, Spatie's `Signal::`, Pennant's `Feature::`, etc.).

## Why the existing paths don't cover this

- **Path 1 (`@see`)** requires authors to add the tag. Most third-party packages don't. The koel example in #787 was first-party application code where the author can be asked to add it; the bar is higher for vendored code we don't control.
- **Path 2 (Testbench probe)** would only work if every package's service provider ran in our Testbench boot. That's already declined by #766 / `ApplicationProvider::doGetApp()` for good reasons (it would explode startup time, leak bindings, and break for providers that need real env config).
- **`FacadeMapProvider` (`AliasLoader` scan)** never sees these classes — they're consumed by FQN, not aliased.

The result is that for any installed package whose facade lacks `@see`, the plugin's behaviour today is: noisy warning during scan, fall back to whatever `@method` catalogue the package ships. If the catalogue drifts from the underlying service the user gets stale types or `UndefinedMagicMethod` (covered by #761).

## Proposal

Add a third resolver path: **statically parse vendor service providers** for accessor → class bindings.

During scan, for every class implementing `Illuminate\Contracts\Support\DeferrableProvider` or extending `Illuminate\Support\ServiceProvider`, walk the `register()` method AST and harvest:

```php
$this->app->bind('subscription', SubscriptionClient::class);
$this->app->singleton('subscription', fn (...) => new SubscriptionClient(...));
$this->app->alias(SubscriptionClient::class, 'subscription');
app()->bind('subscription', ...);
```

Three shapes worth covering:

1. **Class-string second argument** — `bind('subscription', SubscriptionClient::class)`. Trivial AST lookup, direct map.
2. **Closure returning an instantiation** — `bind('subscription', fn () => new SubscriptionClient(...))`. Inspect the closure body's `return new X(...)` for the class.
3. **`alias()` form** — `alias(SubscriptionClient::class, 'subscription')`. Reverse direction, also a direct map.

Anything more complex (factories that branch, providers that read config to pick a class, deferred bindings via container interfaces) falls through to the existing paths.

Mapping is registered with `FacadeMapProvider::registerCustomFacade()` exactly as path 1 does today, so downstream `FacadeMethodHandler` and `getFacadeClasses()` lookups need no further changes.

## Scope and tradeoffs

- **Stops the warning.** Most third-party facades resolve through static parsing, the runtime probe is never reached, no noise.
- **Improves types over hand-maintained `@method`.** Once the root class is known, `FacadeMethodHandler` proxies the service's actual reflected signatures. `@method` precedence is preserved (Psalm checks pseudo-methods first), so package authors who deliberately narrow types via `@method` are unaffected.
- **Cost** is one-time at scan: walk `vendor/*/src/**/ServiceProvider.php` AST once, cache by file mtime in `.psalm-cache`. Comparable to the cost of the existing model registration pass.
- **Limits**: dynamic bindings (`bind(config('foo.driver'), ...)`) and runtime-conditional registration are not covered. Those still need `@see` or a future runtime hook.

## Related

- #787 — original facade method resolution (closed by #789).
- #761 — `UndefinedMagicMethod` on third-party facades. This proposal subsumes most of #761's repros once the package author hasn't shipped a complete `@method` catalogue.
- #766 — string-alias container resolution outside Testbench. Same root cause family; static parsing of providers is the version of #766 that doesn't need to boot user providers.
- #941 — cosmetic fix for warning rendering (independent).

alies-dev:
## Issue to Solve

Closes #942. Vendor packages (e.g. `imdhemy/laravel-in-app-purchases`) ship `Facade` subclasses whose accessor is a string alias (`'subscription'`) bound by the package's `ServiceProvider::register()`. After #789, the plugin's two existing resolver paths both fail:

1. `@see` docblock pointer (only works when the package author added one)
2. Runtime `Facade::getFacadeRoot()` probe (throws `BindingResolutionException` when the bound concrete has constructor deps that aren't in our Testbench container)

The probe failure emits a noisy warning every run and the plugin falls back to the `@method` catalogue (types drift from the actual service class).

This PR supersedes #951 (which solved the same problem by statically parsing every `ServiceProvider` file with AST tooling — 2050 lines, 14 files). The container-introspection approach ships ~525 added lines and avoids parsing user code that may not match runtime behaviour.

## Solution Description

Two pieces, both small:

### 1. `Psalm\LaravelPlugin\Providers\ContainerBindingMapProvider` (new)

Snapshots `$app->getBindings()` and the container's protected `$aliases` array post-boot. Two extraction paths per binding closure:

- **Class-string concrete** (`bind('subscription', SubscriptionClient::class)`): `Container::getClosure()` wraps with `use ($abstract, $concrete)`, so `ReflectionFunction::getStaticVariables()` exposes the class-string directly.
- **Factory closure** (`bind('subscription', fn () => new SubscriptionClient())`): re-parses the host file with PhpParser, runs `NameResolver`, locates the first `new X(...)` inside the closure's line range. Known limitation: conditional `new` in the closure body takes the first match (documented in the class docblock).

Aliases (`Container::alias($abstract, $alias)` stores `aliases[$alias] = $abstract`) are merged into the map so a lookup by either name resolves.

### 2. `AppFacadeRegistrationHandler::tryGetFacadeRootClass()`

Map lookup runs BEFORE the `$failedFacades` cache short-circuit so a facade visited and rejected during scan phase (before its provider has registered) can still resolve once the snapshot fills in. Falls through to the existing runtime probe + warning path on map miss.

### Verification

End-to-end against `imdhemy/laravel-purchases` (real Laravel 13 app, `composer require imdhemy/laravel-purchases`):

```php
use Imdhemy\Purchases\Facades\Subscription;
$client = Subscription::googlePlay();
/** @psalm-check-type-exact $client = \Imdhemy\Purchases\Subscription */
```

Psalm reports zero errors on this file and no `getFacadeRoot() failed` warning. Wrong-type assertion produces the expected `CheckType` error.

### Scope

This PR targets branch 1 of `ApplicationProvider::doGetApp()` — the common case where the user's `bootstrap/app.php` runs the kernel and registers all vendor providers before the plugin's snapshot. Branch 3 (analyzing a Laravel **package** from inside its source repo, where Testbench's defaults block vendor discovery) is deferred to follow-up #956.

## Related

Closes #942. Supersedes #951 (please close after merging). Branch-3 case tracked in #956.

#766 (`app('alias')` returning `mixed` for the same accessor shape) is structurally adjacent but out of scope.

## Checklist
- [x] Tests cover the change (5 unit tests in `tests/Unit/Providers/ContainerBindingMapProviderTest.php`, 1 type test in `tests/Type/tests/Facades/StringAliasFacadeViaContainerMapTest.phpt`, fixture under `tests/Application/app/{Facades,Services,Providers}/`)
- [x] Documentation is updated (in-class docblock covers extraction paths and known limitations)

### DIFF
--- a/src/Handlers/Facades/AppFacadeRegistrationHandler.php
+++ b/src/Handlers/Facades/AppFacadeRegistrationHandler.php
@@ -6,6 +6,7 @@ namespace Psalm\LaravelPlugin\Handlers\Facades;
 
 use Illuminate\Support\Facades\Facade;
 use Psalm\Codebase;
+use Psalm\LaravelPlugin\Providers\ContainerBindingMapProvider;
 use Psalm\LaravelPlugin\Util\AnonymousClassNameDetector;
 use Psalm\Plugin\EventHandler\AfterClassLikeVisitInterface;
 use Psalm\Plugin\EventHandler\AfterCodebasePopulatedInterface;
@@ -177,14 +178,40 @@ final class AppFacadeRegistrationHandler implements AfterClassLikeVisitInterface
      * - The accessor is a string alias bound in our Testbench container (first-party
      *   services like `'cache'`, `'router'`, package bindings registered via discovered
      *   providers).
+     * - The accessor is a string alias whose binding is in the container map snapshot
+     *   maintained by {@see ContainerBindingMapProvider} — covers vendor facades whose
+     *   `register()` body binds via a factory closure ({@see https://github.com/psalm/psalm-plugin-laravel/issues/942})
+     *   and resolves silently without invoking the runtime probe (which would throw on
+     *   constructors with un-bound interface deps).
      *
-     * Returns null when the accessor is a string alias bound only by a user service
-     * provider that does not run in Testbench — nothing we can do at this layer.
+     * Returns null when the accessor is neither in the snapshot map nor resolvable via
+     * the runtime probe — typically a user provider that does not run inside our
+     * Testbench app at all.
      *
      * @return ?class-string
      */
     public static function tryGetFacadeRootClass(string $facadeClass, ?Progress $progress = null): ?string
     {
+        // Container binding map lookup runs BEFORE the $failedFacades gate so a facade
+        // visited and rejected during scan phase (e.g. before its provider has registered
+        // the alias) can still resolve once the populated map fills in. Map path is
+        // silent: a hit reports nothing because nothing failed. `getFacadeAccessor()`
+        // is protected on Facade subclasses, so we read it by reflection.
+        try {
+            if (\is_subclass_of($facadeClass, Facade::class)) {
+                /** @psalm-var mixed $accessor */
+                $accessor = (new \ReflectionMethod($facadeClass, 'getFacadeAccessor'))->invoke(null);
+                if (\is_string($accessor) && $accessor !== '') {
+                    $fromMap = ContainerBindingMapProvider::getClass($accessor);
+                    if ($fromMap !== null) {
+                        return $fromMap;
+                    }
+                }
+            }
+        } catch (\Throwable) {
+            // Fall through to runtime probe.
+        }
+
         // $failedFacades gates both the short-circuit return AND the warning emission,
         // so each failure reason is surfaced to the user exactly once per facade across
         // scan-phase + populate-phase invocations (issue #787: silent losses of method
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -9,6 +9,7 @@ use Psalm\Internal\Analyzer\ProjectAnalyzer;
 use Psalm\LaravelPlugin\Providers\AliasStubProvider;
 use Psalm\LaravelPlugin\Providers\ApplicationProvider;
 use Psalm\LaravelPlugin\Providers\CarbonStubProvider;
+use Psalm\LaravelPlugin\Providers\ContainerBindingMapProvider;
 use Psalm\LaravelPlugin\Providers\FacadeMapProvider;
 use Psalm\LaravelPlugin\Providers\SchemaStateProvider;
 use Psalm\LaravelPlugin\Util\InternalErrorReporter;
@@ -40,6 +41,12 @@ final class Plugin implements PluginEntryPointInterface
             // Handlers use FacadeMapProvider::getFacadeClasses() in getClassLikeNames()
             // to also register for facade/alias classes that proxy to their service.
             FacadeMapProvider::init($output);
+
+            // Snapshot accessor → service-class bindings from the booted container.
+            // Consumed by AppFacadeRegistrationHandler::tryGetFacadeRootClass() to
+            // resolve vendor facades whose getFacadeAccessor() returns a string
+            // alias (#942) without triggering the runtime probe.
+            ContainerBindingMapProvider::init();
 
             // Always called — provides type narrowing (string vs array) regardless
             // of whether findMissingTranslations is enabled
--- a/src/Providers/ContainerBindingMapProvider.php
+++ b/src/Providers/ContainerBindingMapProvider.php
@@ -1,0 +1,184 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Providers;
+
+use Illuminate\Container\Container;
+use Illuminate\Foundation\Application as LaravelApplication;
+use PhpParser\Node;
+use PhpParser\NodeFinder;
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\NameResolver;
+use PhpParser\ParserFactory;
+
+/**
+ * Snapshot of container accessor → concrete class from `$app->getBindings()`,
+ * used to resolve vendor facades whose accessor is a string alias bound by a
+ * service provider (issue #942) without invoking the runtime `Facade::getFacadeRoot()`
+ * probe, which throws on constructors with un-bound deps.
+ *
+ * Known limitations:
+ *  - Factory closure with branching (`fn () => $cond ? new A : new B`) — only the
+ *    first `new` is taken.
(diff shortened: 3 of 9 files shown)


alies-dev:
Simplified the PR. Diff dropped from **915 → 525 lines added** (43% reduction).

## Cuts applied

| # | Cut | Result |
|---|-----|--------|
| 1 | All Testbench-tuning code in `ApplicationProvider` (`ignorePackageDiscoveriesFrom` override, `retargetPackageManifestAtProjectRoot`, `safelyRegisterAllProviders`, `safelyBootProviders`, modified `resolveApplicationBootstrappers`) | -305 lines |
| 2 | Class docblock on `ContainerBindingMapProvider` collapsed from ~35 lines design-doc to ~12 lines (limitations only) | -55 lines |
| 3 | Inlined `lookupRootClassFromContainerMap` into `tryGetFacadeRootClass` | -15 lines |
| 4 | Dropped the `$fileCache` map in `parsedFile()` | -20 lines |
| 5 | Skipped — dropping `NameResolver` breaks imdhemy (its `bindFacades()` uses `use Imdhemy\Purchases\Subscription;` + `new Subscription()`, so `Name::toString()` yields the local name and `class_exists` fails) | 0 |

## Verification (cut 1)

The Testbench tuning was needed for branch 3 of `ApplicationProvider::doGetApp()` (analyzing a Laravel package from its source repo, where `bootstrap/app.php` doesn't exist). For the actual #942 scenario — a Laravel **app** with `bootstrap/app.php` running `vendor/bin/psalm` — branch 1 is taken. The user's kernel boots, vendor providers register, the binding snapshot has everything it needs.

Verified end-to-end against `imdhemy/laravel-purchases` in a real Laravel 13 app (`composer create-project laravel/laravel`), with the Testbench tuning fully reverted:

- `Subscription::googlePlay()` resolves to `Imdhemy\Purchases\Subscription` (passes `@psalm-check-type-exact`)
- No `getFacadeRoot() failed for ...` warning anywhere in the run

## Deferred to follow-up

The branch-3 case (analyzing a Laravel package source, not an app) is tracked in #956. The reverted machinery was real work; the audience is smaller (package authors); landing it behind a smaller, opt-in design will happen when there's demand.

## What stayed

Untouched per the simplification spec:
- The unit tests in `tests/Unit/Providers/ContainerBindingMapProviderTest.php`
- The type test + Application fixtures
- Path 1 (class-string extraction via `ReflectionFunction::getStaticVariables()`)
- Alias merging via `readAliases()`
- The `$failedFacades` reorder in `AppFacadeRegistrationHandler::tryGetFacadeRootClass()`

All tests still green (`composer test:type`, `composer test:unit`, `composer test:app`, `composer psalm`, `composer cs`).

alies-dev:
Closing context: empirical re-testing showed master plugin code already resolves the popular case (Laravel applications with a real `bootstrap/app.php` — branch 1 of `ApplicationProvider::doGetApp()`): the user's kernel boots, every vendor `register()` runs, and the runtime `Facade::getFacadeRoot()` probe finds the binding.

The case that genuinely breaks is **branch 3** — Psalm running inside a Laravel **package source** repo (no `bootstrap/app.php`), which falls into Testbench's `createApplication()`. Testbench's `PackageManifest` honors `ignorePackageDiscoveriesFrom() === ['*']` so no vendor providers register and string-alias accessors like `'subscription'` (imdhemy) stay unbound.

#957 adds a branch-3-only `registerDiscoveredVendorProviders()` (~40 LOC) that runs the framework's own `PackageManifest` against the project root and registers each discovered provider with per-provider try/catch. No snapshot infrastructure needed — once the binding is registered, the existing runtime probe in `AppFacadeRegistrationHandler::tryGetFacadeRootClass()` returns the right class on its own.

The container-introspection approach in this PR was designed for a non-problem; the snapshot is unnecessary defensive infrastructure. #957 supersedes it.

PR #957, opened 2026-05-17
Resolve custom Facades in Laravel package source repos (for running on package repos)

### LINKED ISSUE #942: Resolve in-package facades via static parsing of vendor service providers

alies-dev:
## Issue to Solve

When Psalm runs **inside a Laravel package source repo** (no `bootstrap/app.php`), the plugin falls into branch 3 of `ApplicationProvider::doGetApp()` and uses `Orchestra\Testbench`'s `createApplication()`. Testbench swaps in its own `PackageManifest` that honors `CreatesApplication::ignorePackageDiscoveriesFrom()`, which defaults to `['*']`. Effect: no vendor providers register.

Custom Facades whose accessor is a string alias bound by a vendor's `register()` (e.g. `imdhemy/laravel-purchases`'s `Subscription::getFacadeAccessor() → 'subscription'`, bound in `LiapServiceProvider::register()`) then fail `Facade::getFacadeRoot()` with:

```
Target class [subscription] does not exist.
```

The plugin emits the noisy `Laravel plugin: getFacadeRoot() failed for ...` warning every run and silently falls back to the facade's `@method` catalogue. Any method on the underlying class **not** listed in `@method` (e.g. `Subscription::id(...)`) becomes `UndefinedMagicMethod` and `MixedAssignment`.

Real-world repro (verified in `/tmp/imdhemy-pkg-verify`): a package-source layout with `imdhemy/laravel-purchases` installed, calling `Subscription::id('item-1')` produces 4 errors on master + the warning. With this fix: 0 errors, no warning, type inferred as `\Imdhemy\Purchases\Subscription`.

Real **Laravel applications** (branch 1, `bootstrap/app.php` exists) are unaffected and already work — the user's kernel boots and the runtime probe finds the binding. This PR fixes only the package-source case.

## Related

Closes #942. Supersedes #951 (AST harvester) and #955 (slim container introspection): empirical testing showed that master already resolves the popular branch-1 case, so PR #955's snapshot infrastructure was solving a non-problem.

## Solution Description

Add a branch-3-only `registerDiscoveredVendorProviders()` to `ApplicationProvider`. It:

1. Instantiates **Laravel's own** `Illuminate\Foundation\PackageManifest` pointed at the project root (bypasses Testbench's swapped variant that rejects every vendor).
2. Iterates `$pm->providers()` and calls `$app->register($fqcn)` for each, wrapped in per-provider try/catch so one bad vendor `register()`/`boot()` does not disable the plugin.

The new method intentionally:

- Reuses Laravel's `PackageManifest` API (no AST parsing of `installed.json`), so `extra.laravel.dont-discover` and `COMPOSER_VENDOR_DIR` are honored for free and the implementation stays in sync with framework changes.
- Does not add an explicit `BootProviders` bootstrap. By the time this method runs, `createApplication()` has already booted the app, so each `$app->register($fqcn)` inline-fires that provider's `boot()` immediately (Laravel's `Application::register()` semantics when `isBooted() === true`). Per-provider try/catch contains failures from either `register()` or the inline `boot()`.
- Diverges from Laravel's `ProviderRepository::load()` deferred-vs-eager partitioning. Static analysis never performs the container hit that triggers a deferred provider, so we register everything eagerly.

The view binding (previously after the `else` block) is hoisted to run **before** `registerDiscoveredVendorProviders()`. Otherwise, common vendor `boot()` patterns (`View::composer`, `Blade::component`) throw `BindingResolutionException` and the per-provider catch silently swallows the failure for any provider with that boot pattern.

`resolveProjectRoot()` is extracted as a shared helper (used by both `retargetConfigPathAtProjectRoot()` and the new method) with the same resolution order it always had: `APP_BASE_PATH` / `TESTBENCH_APP_BASE_PATH` → `getcwd()` if it contains a `composer.json` → bail.

## Verification

- New unit test `ApplicationProviderVendorRegistrationTest` exercises the isolation contract end to end. Pins a fake project root via `APP_BASE_PATH`, ships a synthetic `installed.json` listing `ThrowingServiceProvider` BEFORE `BindingServiceProvider`, then asserts that `BindingServiceProvider`'s container binding is present after boot. The ordering is deliberate — a regression that aborts iteration on the first thrown exception would fail this test.
- Minimal package-source reproducer at `/tmp/imdhemy-pkg-verify`: master fires 4 errors + warning; fix → 0 errors, no warning; negative case (asserted type flipped to `string`) fails correctly, proving the type is inferred not silenced.
- `composer cs`, `composer rector`, `composer psalm`, `composer test:type`, `composer test:unit`, `composer test:app` all pass.

## Checklist

- [x] Tests cover the change (`tests/Unit/Providers/ApplicationProviderVendorRegistrationTest.php` + fixtures under `tests/Unit/Providers/Fixtures/`)
- [ ] Documentation is updated (if needed, otherwise remove this item)


alies-dev:
## PR delta benchmark vs master

Ran `/psalm-pr-delta` against 14 real-world Laravel apps (aureuserp, bagisto, bookstack, coolify, corcel, monica, pixelfed, solidtime, spatie-dashboard, tastyigniter, unit3d, vito, laravel-excel, filament).

**master `6f941dee` -> head `d845f761`**

| App              | ΔIssues | ΔCov   | ΔTime  |
|------------------|--------:|-------:|-------:|
| aureuserp        |      +0 |  +0.00 |  +0.19s |
| bagisto          |      +0 |  +0.00 |  +0.19s |
| bookstack        |      +0 |  +0.00 |  +0.25s |
| coolify          |      +0 |  -0.00 |  +0.11s |
| corcel           |      +0 |  +0.00 |  +0.08s |
| monica           |      +0 |  +0.00 |  +0.00s |
| pixelfed         |      +0 |  +0.00 |  -0.37s |
| solidtime        |      +0 |  +0.00 |  -0.15s |
| spatie-dashboard |      +0 |  +0.00 |  +0.02s |
| tastyigniter     |      +0 |  +0.00 |  -0.12s |
| unit3d           |      +0 |  +0.00 |  -0.16s |
| vito             |      +0 |  +0.00 |  -0.48s |
| laravel-excel    |      +0 |  +0.00 |  +0.24s |
| filament         |      +0 |  +0.00 |  -0.10s |
| **Total / Avg**  | **+0** | **+0.00** | **-0.30s** |

**Issue type movements:** none.

Expected for this PR. All 14 apps in the benchmark suite are full Laravel applications with `bootstrap/app.php` (branch 1 of `ApplicationProvider::doGetApp()`). The fix only kicks in for branch 3 (Laravel package source repos with no `bootstrap/app.php`), which the suite does not cover. The dedicated repro for the fix is `imdhemy/laravel-in-app-purchases` (package source repo). Running that next.

alies-dev:
Pushed `feat(providers): merge root composer.json providers in branch-3 boot #942` (34128a43).

**What changed:** new private helper `readRootComposerProviders()` reads `composer.json:extra.laravel.providers` from the project root and merges its result into the existing PackageManifest-discovered list before the per-provider `try/catch` loop. JSON decode uses `JSON_THROW_ON_ERROR`; all failure modes (missing file, unreadable, malformed JSON, shape mismatch) return `[]` (best-effort discovery). `$app->register()` is idempotent so duplicates across vendor + root are no-ops.

**Why:** `PackageManifest::build()` only reads `vendor/composer/installed.json` (dependencies). When Psalm analyses a Laravel package's own source repo (imdhemy/laravel-purchases, corcel, laravel-excel, etc.), the package's OWN provider lives in the ROOT manifest. Without it, custom-Facade accessors stay unbound and `Facade::getFacadeRoot()` warns at every static-call site.

**Verified locally:**
- `composer cs` / `composer psalm` / `composer test:unit` / `composer test:type` / `composer test:app` all green.
- imdhemy vendor-case reproducer (`/tmp/imdhemy-pkg-verify`): No errors found.
- imdhemy source-repo case (`/tmp/imdhemy-package-test`): `getFacadeRoot()` warnings for `Subscription` and `Product` are gone.

**PR delta vs master (22cd3487 -> 34128a43):**

| App              | ΔIssues | ΔCov   | ΔTime |
|------------------|--------:|-------:|------:|
| aureuserp        |      +0 |  +0.00 |  +1.22s |
| bagisto          |      +0 |  +0.00 |  +1.03s |
| bookstack        |     +11 |  -0.03 |  +1.23s |
| coolify          |      +0 |  +0.00 |  +2.70s |
| corcel           |      +1 |  -0.06 |  +1.58s |
| monica           |      +0 |  +0.00 |  +3.29s |
| pixelfed         |      +0 |  -0.00 |  -0.47s |
| solidtime        |      +0 |  +0.00 |  -0.02s |
| spatie-dashboard |      +0 |  +0.00 |  -0.06s |
| tastyigniter     |      +0 |  +0.00 |  -0.45s |
| unit3d           |      +0 |  +0.00 |  -0.19s |
| vito             |      +0 |  +0.00 |  -0.54s |
| laravel-excel    |      -2 |  +0.04 |  -0.54s |
| filament         |      +0 |  +0.00 |  -1.96s |
| **Total / Avg**  | **+10** | **-0.00** | **+6.82s** |

Issue-type movements:

| Issue type | Base | PR | Δ |
|------------|-----:|---:|--:|
| MixedArgument | 7565 | 7570 | +5 |
| InvalidArgument | 853 | 851 | -2 |
| PossiblyInvalidArgument | 575 | 577 | +2 |
| PossiblyNullArgument | 1223 | 1221 | -2 |
| RawObjectIteration | 6 | 8 | +2 |
| MissingDependency | 0 | 1 | +1 |
| MixedArgumentTypeCoercion | 779 | 780 | +1 |
| MixedAssignment | 9461 | 9462 | +1 |
| MixedMethodCall | 8618 | 8619 | +1 |
| MixedPropertyFetch | 4312 | 4313 | +1 |

Newly introduced: `MissingDependency` (1).

`corcel +1` and `laravel-excel -2 / +0.04 cov` confirm the root-composer path activates (previous delta was zero on both). `bookstack +11` is new analysis surface — bookstack's own provider now registers, exposing previously-Mixed code paths to type-checking; the increase is concentrated in `MixedArgument` (+5) and related Mixed* types, plus a small `InvalidArgument -2` / `PossiblyNullArgument -2` improvement.

alies-dev:
## PR delta benchmark

Base: `merge-base(master)` (`6f941dee`) → head `7b57dd79`. Plugin run on 17 real-world Laravel apps. Cache-keyed, no rerun overhead.

### Per-app delta

| aureuserp        |    0 |    0 |    +0 |  +0.00 | -0.33s |
| bagisto          |    0 |    0 |    +0 |  +0.00 | +0.13s |
| bookstack        |    0 |    0 |    +0 |  +0.00 | -0.23s |
| coolify          |    0 |    0 |    +0 |  -0.00 | +0.07s |
| corcel           |    0 |    0 |    +0 |  +0.00 | -0.11s |
| monica           |    0 |    0 |    +0 |  +0.00 | -0.30s |
| pixelfed         |    0 |    0 |    +0 |  -0.00 | -0.15s |
| solidtime        |    0 |    0 |    +0 |  +0.00 | -0.13s |
| spatie-dashboard |    0 |    0 |    +0 |  +0.00 | -0.08s |
| tastyigniter     |    0 |    0 |    +0 |  +0.00 | +0.38s |
| unit3d           |    0 |    0 |    +0 |  +0.00 | +0.24s |
| vito             |    0 |    0 |    +0 |  +0.00 | +0.14s |
| laravel-excel    |    0 |    2 |    -2 |  +0.04 | +0.01s |
| filament         |    2 |    0 |    +2 |  -0.00 | -0.09s |
| laravel-backup   |   10 |   11 |    -1 |  +0.06 | -0.22s |
| laravel-permission |   44 |  443 |  -428 |  +5.25 | +0.11s |
| laravel-socialite |    0 |    0 |    +0 |  +0.00 | +0.15s |
| **Total / Weighted** | **56** | **456** | **-429** | **+0.23** | **-0.41s** |

_ΔCov is weighted by base `wall_seconds` (proxy for codebase size)._

### Issue type movements (across all apps)

| MixedMethodCall | 0 | 228 | -228 |
| MixedAssignment | 2 | 126 | -124 |
| MixedPropertyFetch | 1 | 40 | -39 |
| MixedArgument | 5 | 33 | -28 |
| NoInterfaceProperties | 11 | 0 | +11 |
| MixedReturnStatement | 0 | 9 | -9 |
| UndefinedPropertyFetch | 9 | 0 | +9 |
| MixedPropertyAssignment | 0 | 8 | -8 |
| MixedArrayOffset | 0 | 7 | -7 |
| PossiblyNullReference | 6 | 0 | +6 |
| PropertyTypeCoercion | 6 | 0 | +6 |
| MixedArrayAccess | 4 | 0 | +4 |
| LessSpecificReturnStatement | 3 | 0 | +3 |
| MissingClosureReturnType | 0 | 3 | -3 |
| MixedArgumentTypeCoercion | 1 | 2 | -1 |
| MoreSpecificReturnType | 3 | 0 | +3 |
| RiskyTruthyFalsyComparison | 2 | 0 | +2 |
| UndefinedMagicMethod | 2 | 0 | +2 |
| InvalidScope | 1 | 0 | +1 |

### Notes

- Headline: **-429 net issues**, **+0.23% weighted coverage**, wall time flat.
- Big win on `laravel-permission`: -428 net, +5.25% coverage. `MixedMethodCall -228`, `MixedAssignment -124`, `MixedPropertyFetch -39`, `MixedArgument -28`. Consistent with registering vendor service providers so Psalm sees real bound types instead of falling back to `mixed`.
- Regressions worth a look: `NoInterfaceProperties +11`, `UndefinedPropertyFetch +9`, `PossiblyNullReference +6`, `PropertyTypeCoercion +6`. `filament` is the only app with positive ΔNet (+2).

Issue #517, opened 2026-03-24
Auto-resolve HasFactory template parameter from Laravel naming convention


alies-dev:
## Description

When a model uses `use HasFactory` without specifying the generic template parameter, Psalm reports `MissingTemplateParam`. The plugin could automatically resolve the factory class using Laravel's naming convention, eliminating these false positives.

## Reproduction

Found while auditing [monicahq/monica](https://github.com/monicahq/monica) with plugin v4.2.0 + Psalm 7.0.0-beta17. **69 occurrences** — one per model.

```php
// app/Models/Contact.php
class Contact extends VCardResource
{
    use HasFactory; // MissingTemplateParam: expecting HasFactory<ContactFactory>
}
```

Error:
```
MissingTemplateParam: App\Models\Contact has missing template params
when extending Illuminate\Database\Eloquent\Factories\HasFactory, expecting 1
```

Every model in the project uses `use HasFactory` without the template param. This is standard Laravel — the framework resolves the factory class at runtime via naming convention.

## Laravel's Factory Resolution Convention

Laravel resolves the factory class from the model name using this logic (from `HasFactory::newFactory()`):

1. If the model has a `newFactory()` method, use its return value
2. Otherwise, derive from namespace: `App\Models\Contact` → `Database\Factories\ContactFactory`
3. The convention is: take the model's base class name, append `Factory`, place in `Database\Factories\`

## Proposed Fix

Add a handler that automatically resolves the `HasFactory` template parameter:

1. When Psalm encounters `use HasFactory` without a template param on a model class
2. Derive the factory class name using Laravel's convention: `{ModelName}Factory` in `Database\Factories\`
3. Register it as `HasFactory<Database\Factories\{ModelName}Factory>`

This would eliminate all 69 `MissingTemplateParam` errors in Monica and similar counts in virtually every Laravel app.

## Impact

Every Laravel model that uses `use HasFactory` (which is the default in all generated models) will trigger this. It's one of the most common Psalm issues in Laravel apps.

PR #961, opened 2026-05-17
Fix `Model::factory()->count(N)->make()` collapse on bare `HasFactory`

### LINKED ISSUE #960: Factory chain `Model::factory()->count(N)->make()` collapses to `Model` when `HasFactory` lacks template arg
## Describe the bug

When a model uses `use HasFactory;` without the template argument (the Laravel default), the chain `Model::factory()->count(N)->make()` is inferred as a single `Illuminate\Database\Eloquent\Model` instead of `Collection<int, Model>`. Downstream code that iterates the result triggers `PossibleRawObjectIteration`, and `$result->pluck(...)` triggers `MixedMethodCall`.

This is downstream of #517 / #612: that PR silenced the `MissingTemplateParam` noise on `use HasFactory;`, but the runtime type returned by `Model::factory()` is still bare `Factory` (no `TModel` binding), which prevents `FactoryCountTypeProvider` (#693) from firing and prevents the conditional return on `make()` / `create()` from picking the collection branch.

## Reproduction

Found in [BookStackApp/BookStack v26.03.3](https://github.com/BookStackApp/BookStack/blob/v26.03.3/database/seeders/DummyContentSeeder.php#L63-L66):

```php
// app/Entities/Models/Bookshelf.php
class Bookshelf extends Entity
{
    use HasFactory; // no @use HasFactory<BookshelfFactory>
}

// database/seeders/DummyContentSeeder.php
$shelves = Bookshelf::factory()->count(10)->make($byData);
foreach ($shelves as $shelf) { // PossibleRawObjectIteration
    $shelf->save();
}
$largeBook->shelves()->attach($shelves->pluck('id')); // MixedMethodCall
```

## Three-stage trace

```
$f1 = Bookshelf::factory()    -> Factory                          (bare, no template args)
$f2 = $f1->count(10)          -> Factory<Model, int|null>         (FactoryCountTypeProvider bailed)
$shelves = $f2->make([])      -> Model                            (conditional picks true branch)
```

## Why each stage fails

1. **`Bookshelf::factory()` returns bare `Factory`.** `HasFactory::factory()` has `@return TFactory` with `@template TFactory of Factory`. With no template argument on `use HasFactory;`, `TFactory` collapses to the bound `Factory`. The plugin already acknowledges this in `SuppressHandler.php:40-46`:

   > HasFactory has @template TFactory, but many Laravel models omit the annotation — the framework resolves the factory class via naming convention at runtime. Suppress the noise. Users who want type-safe factory() calls can add `@use HasFactory<ConcreteFactory>`.

2. **`FactoryCountTypeProvider` bails** (`src/Handlers/Eloquent/FactoryCountTypeProvider.php:84-128`). `resolveModelType()` cannot recover `TModel`:
   - Receiver has no template params (bare `Factory`), so `$event->getTemplateTypeParameters()` returns null.
   - Fallback `resolveModelFromClass(Factory::class)` reads `template_extended_params[Factory::class]`, which is empty because `Factory` does not extend itself.
   - Handler returns `null`, so Psalm uses the stub's declared `@return Factory<TModel, int|null>` from `stubs/common/Database/Eloquent/Factories/Factory.phpstub:76`. `TModel` is unbound, so Psalm uses the template bound `Model`, giving `Factory<Model, int|null>`.

3. **`make()` conditional picks the wrong branch.** Stub declares `(TCount is null|0|1 ? TModel : Collection<int, TModel>)`. With `TCount = int|null`, Psalm 7's conditional resolution picks the true branch (`TModel`) instead of producing a union of both branches.

## Workaround (app side, zero plugin work)

Add the template argument on the trait use:

```php
/** @use HasFactory<\Database\Factories\Entities\Models\BookshelfFactory> */
use HasFactory;
```

After that the chain works end-to-end:
- `Bookshelf::factory()` -> `Factory<Bookshelf>` (because `TFactory = BookshelfFactory extends Factory<Bookshelf>`, assuming #780 is also resolved or the factory class declares `@extends Factory<Bookshelf>`).
- `count(10)` -> handler fires -> `Factory<Bookshelf, 2>`.
- `make()` -> `Collection<int, Bookshelf>`.

The workaround is rarely applied in real apps — every default `php artisan make:model -f` output omits the template arg.

## Proposed fix

**Option A (recommended)**. Add a `MethodReturnTypeProviderInterface` for `Illuminate\Database\Eloquent\Factories\HasFactory::factory()` that returns `Factory<static>` regardless of `TFactory`. Mirrors what Larastan does. Auto-fixes every Laravel app where models omit the trait template arg (the common case).

**Option B (smaller, complementary).** Make `FactoryCountTypeProvider::resolveModelType()` fall back to `Model` when it cannot resolve `TModel`, so at least the `TCount` narrowing survives. Without Option A this still leaves `Collection<int, Model>` — worse than Option A but better than `Model`, and `foreach` no longer triggers `PossibleRawObjectIteration`.

Both can ship together. Option A is the upstream fix at the layer Laravel users forget; Option B hardens the handler against any other path that produces a bare `Factory`.

## Real-world occurrence

In `bookstack` (v26.03.3) `database/seeders/DummyContentSeeder.php` alone has 4 factory chains that hit this:

- `Book::factory()->count(5)->make($byData)->each(...)` (line 42)
- `Chapter::factory()->count(3)->create($byData)->each(...)` (line 45)
- `Page::factory()->count(3)->make(...)` (lines 47, 50, 58)
- `Bookshelf::factory()->count(10)->make($byData)` (line 63)

Every `foreach`/`pluck`/`each` over those results becomes a false positive.

## Impacted Versions

```
laravel/framework         12.x
psalm/plugin-laravel      4.9.3
vimeo/psalm               7.x
```

## Related

- #517 / #612 — suppressed `MissingTemplateParam` on `use HasFactory;` but did not auto-resolve `TFactory`.
- #693 — introduced `FactoryCountTypeProvider`, which is bypassed in this scenario.
- #780 — symmetric problem on the `Factory` subclass side (`class XFactory extends Factory` without template arg).

alies-dev:
## Issue to Solve

`Model::factory()->count(N)->make()` collapsed to a single `Model` instead of `Collection<int, Model>` when the model wrote `use HasFactory;` without an explicit `@use HasFactory<XFactory>` template arg (the default that `php artisan make:model -f` generates). Every `foreach` over the chain result triggered `PossibleRawObjectIteration`, and `$result->pluck(...)` triggered `MixedMethodCall`. Real-world example: [BookStack v26.03.3 `DummyContentSeeder.php:63`](https://github.com/BookStackApp/BookStack/blob/v26.03.3/database/seeders/DummyContentSeeder.php#L63-L66).

Root cause is a three-stage chain failure documented in the linked issue: the stub's `TFactory` template default falls back to bare `Factory`, which `FactoryCountTypeProvider` cannot bind `TModel` from, and the `make()` conditional then picks the single-model branch because `TCount` is also unbound.

## Related

Closes #960. Downstream of #517 / #612 (which silenced `MissingTemplateParam` on `use HasFactory;` but did not auto-resolve `TFactory`).

## Solution Description

Two-pronged, covering both the common case and any other path that produces a bare `Factory`.

### Option A: new `ModelFactoryMethodTypeProvider`

Hooks `HasFactory::factory()` via the trait FQCN (Psalm dispatches return-type providers through the `declaring_method_id` branch when a trait method is invoked on the using class). Returns `Factory<ModelFqcn, null>` so the downstream chain has both `TModel` and `TCount` to work with. Defers to the stub when the user has written `@use HasFactory<XFactory>` (the documented escape hatch) — `XFactory` is strictly more precise than `Factory<Model>` because it preserves any subclass-specific state methods.

### Option B: `FactoryCountTypeProvider` `Model` fallback

`resolveModelType()` gains a third tier returning `Union[Model]` when both the receiver template lookup and the `extends Factory<X>` binding fail. Keeps the `TCount` narrowing intact for any other path that emits a bare `Factory` (e.g. `__callStatic` on a Facade losing template context). Downgraded vs Option A's precise `Factory<ConcreteModel>` but still iterable, so `foreach` no longer trips `PossibleRawObjectIteration`.

While in the same function, `isModelType()` was switched from `is_a()` (PHP autoload, fails for classes Psalm has scanned but PHP cannot resolve at handler runtime — including the PHPT test fixtures) to a `classlike_storage_provider` lineage check that matches the new handler's lookup shape.

### Known limitation (follow-up candidate)

`Model::factory($count)->make()` with a numeric first arg is not folded into `TCount` here. Laravel's runtime forwards `$count` to `->count()` internally (`HasFactory::factory()` does `$factory->count(is_numeric($count) ? $count : null)`), so `Bookshelf::factory(3)->make()` is functionally identical to `Bookshelf::factory()->count(3)->make()` and should resolve to `Collection<int, Bookshelf>`. With this PR it resolves to `Bookshelf` (the single-model branch of the conditional). Users writing `factory(N)->make()` should switch to `factory()->count(N)->make()` for now; the symmetric handling can be added in a follow-up.

### Verification

- `tests/Type/tests/Model/FactoryMakeOnHasFactoryWithoutTemplateArgTest.phpt` (new) — Option A: `factory()`, `count(N)->make()`, `create()`, `foreach`, plus LSB through an abstract base class (`Subclass extends Base extends Model` where `Base` uses `HasFactory`).
- `tests/Type/tests/Model/FactoryCountFallbackToBaseModelTest.phpt` (new) — Option B: bare `Factory` parameter, fallback to `Collection<int, Model>`, foreach yields `Model`.
- `tests/Type/tests/HasFactoryTemplateParamTest.phpt` (extended) — locks the explicit `@use HasFactory<X>` escape hatch (the prior fixture had `@use` on the class instead of the `use` statement; Psalm only populates `template_extended_params` from the `use`-statement docblock, so the binding was previously untested).
- Full type suite (`composer test:type`): 404/404 pass.
- Psalm self-analysis (`composer psalm`): no errors.
- Unit suite (`composer test:unit`): 624/624 pass.

## Checklist

- [x] Tests cover the change (3 type tests in `tests/Type/tests/Model/` and an extension to `HasFactoryTemplateParamTest.phpt`)

### DIFF
--- a/src/Handlers/Eloquent/FactoryCountTypeProvider.php
+++ b/src/Handlers/Eloquent/FactoryCountTypeProvider.php
@@ -108,6 +108,19 @@ final class FactoryCountTypeProvider implements MethodReturnTypeProviderInterfac
      *  2. The called class's `extends Factory<X>` binding (covers user
      *     subclasses like `class UserFactory extends Factory<User>` and
      *     static calls `UserFactory::times(N)`).
+     *  3. Last-resort fallback to base `Model`. Without this, an unbound
+     *     receiver (bare `Factory` with no template params and no subclass
+     *     binding — produced e.g. when a model uses `HasFactory` without
+     *     `@use HasFactory<XFactory>`) collapses through the stub's
+     *     `@return Factory<TModel, int|null>` and `make()`'s conditional
+     *     picks the single-model branch. Falling back to `Model` keeps the
+     *     `TCount` narrowing intact so `count(N)->make()` still resolves to
+     *     `Collection<int, Model>` (downgraded but iterable) rather than a
+     *     bare `Model`. Companion handler `ModelFactoryMethodTypeProvider`
+     *     covers the common case with precise `Factory<ConcreteModel>`.
+     *
+     * @see ModelFactoryMethodTypeProvider
+     * @see https://github.com/psalm/psalm-plugin-laravel/issues/960
      *
      * @psalm-mutation-free
      */
@@ -117,25 +130,53 @@ final class FactoryCountTypeProvider implements MethodReturnTypeProviderInterfac
 
         if ($templateParams !== null) {
             $modelType = $templateParams[0] ?? null;
-            if (self::isModelType($modelType)) {
+            if (self::isModelType($modelType, $codebase)) {
                 return $modelType;
             }
         }
 
         $calledClass = $event->getCalledFqClasslikeName() ?? $event->getFqClasslikeName();
+        $resolved = self::resolveModelFromClass($calledClass, $codebase);
+        if ($resolved instanceof Union) {
+            return $resolved;
+        }
 
-        return self::resolveModelFromClass($calledClass, $codebase);
+        // Allocated per-fallback (rather than cached as a static) to keep
+        // resolveModelType()'s @psalm-mutation-free contract intact. This
+        // path only fires for unresolvable receivers — uncommon enough that
+        // the extra allocation does not warrant relaxing the purity marker.
+        return new Union([new TNamedObject(Model::class)]);
     }
 
-    /** @psalm-mutation-free */
-    private static function isModelType(?Union $type): bool
+    /**
+     * Lineage check via Psalm's class storage. `is_a()` was previously used here
+     * but breaks for classes Psalm scans yet PHP's autoloader cannot resolve
+     * at handler runtime (e.g. PHPT fixture models declared inline).
+     *
+     * @psalm-mutation-free
+     */
+    private static function isModelType(?Union $type, Codebase $codebase): bool
     {
         if (!$type instanceof Union) {
             return false;
         }
 
         foreach ($type->getAtomicTypes() as $atomic) {
-            if (!$atomic instanceof TNamedObject || !\is_a($atomic->value, Model::class, true)) {
+            if (!$atomic instanceof TNamedObject) {
+                return false;
+            }
+
+            if ($atomic->value === Model::class) {
+                continue;
+            }
+
+            try {
+                $storage = $codebase->classlike_storage_provider->get($atomic->value);
+            } catch (\InvalidArgumentException) {
+                return false;
+            }
+
+            if (!isset($storage->parent_classes[\strtolower(Model::class)])) {
                 return false;
             }
         }
@@ -170,7 +211,7 @@ final class FactoryCountTypeProvider implements MethodReturnTypeProviderInterfac
 
         $modelType = $extended['TModel'] ?? null;
 
-        return self::isModelType($modelType) ? $modelType : null;
+        return self::isModelType($modelType, $codebase) ? $modelType : null;
     }
 
     /**
--- a/src/Handlers/Eloquent/ModelFactoryMethodTypeProvider.php
+++ b/src/Handlers/Eloquent/ModelFactoryMethodTypeProvider.php
@@ -1,0 +1,142 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Eloquent;
+
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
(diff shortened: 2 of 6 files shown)


alies-dev:
Superseded by #964, which uses Larastan-style `Factory::resolveFactoryName()` discovery to return the concrete `XFactory` subclass when available, exposing subclass state methods (`forUser()`, `withOwner()`, etc.) on the chain without requiring `@use HasFactory<XFactory>`. Falls back to this PR's `Factory<modelFqcn, null>` shape when no concrete factory is discoverable. Verified equivalent on bookstack/solidtime and −37 issues on monica.

Issue #486, opened 2026-03-22
pluck() should infer value type from @property annotations


alies-dev:
## Summary

`Collection::pluck()` on a typed Eloquent model always returns `Collection<int, mixed>`, even when the plucked column has a known type from `@property` annotations.

## Example

```php
/**
 * @property string $notification_class
 */
class User extends Model {}

// Current: Collection<int, mixed>
// Expected: Collection<int, string>
$emails = User::query()
    ->select('email')
    ->distinct()
    ->get()
    ->pluck('email');
```

## Problem

Because `pluck()` returns `mixed` values, downstream operations like `combine()` trigger `InvalidTemplateParam` errors since `combine()` expects `array-key` keys:

```php
// ERROR: InvalidTemplateParam - TKey expects array-key, mixed given
$emails->combine($emails)->all();
```

Users must work around this with explicit `@var` annotations:

```php
/** @var \Illuminate\Support\Collection<int, string> $emails */
$emails = User::query()->...->pluck('email');
```

## Expected behavior

When `pluck('column_name')` is called on an `Eloquent\Collection` (or query builder result) where the model has `@property Type $column_name`, the plugin should infer the return type as `Collection<int, Type>` instead of `Collection<int, mixed>`.

This would eliminate a class of false-positive `InvalidTemplateParam` errors and reduce the need for manual type annotations.

PR #968, opened 2026-05-17
Narrow `pluck($value, $key)` key type and cover relation chains

### LINKED ISSUE #967: `pluck($value, $key)->all()` returns `array<array-key, mixed>` instead of `array<KeyType, ValueType>` from @property annotations
## Summary

When `Collection::pluck($valueColumn, $keyColumn)` is called on an Eloquent collection with both arguments, and both columns have known types via `@property` annotations on the model, the resulting `->all()` is still inferred as `array<array-key, mixed>` instead of `array<KeyType, ValueType>`. This forces a `LessSpecificReturnStatement` on any method that declares the narrowed shape as its return type.

This is the two-argument sibling of #486 (which covered single-argument `pluck('column')` and is closed completed). Two-argument `pluck` is not narrowed today.

## Repro

```php
use Illuminate\Database\Eloquent\Relations\HasMany;

/**
 * @property string $title
 * @property string $slug
 */
class Article extends Model {}

/**
 * @property string $name
 *
 * @method HasMany<Article, $this> articles()
 */
class Author extends Model
{
    public function articles(): HasMany
    {
        return $this->hasMany(Article::class);
    }
}

class Presenter
{
    /**
     * @return array<string, string>
     */
    public function titlesBySlug(Author $author): array
    {
        return $author->articles()->pluck('title', 'slug')->all();
    }
}
```

Psalm reports:

```
ERROR: LessSpecificReturnStatement
The type 'array<array-key, mixed>' is more general than the declared return type
'array<string, string>' for Presenter::titlesBySlug (see https://psalm.dev/129)
    return $author->articles()->pluck('title', 'slug')->all();
```

Same shape applies when the collection comes from `Model::query()->get()->pluck('value_col', 'key_col')`, or from any `Eloquent\Collection` whose model declares both columns via `@property`.

## Expected

For `Collection<TKey, TModel of Model>::pluck(string \$value, string \$key)`:

- if `TModel` has `@property V \$value` and `@property K \$key` (where `K` resolves to an `array-key` type), the return should be `Support\Collection<K, V>`;
- consequently `->all()` should resolve to `array<K, V>`.

Today the plugin appears to leave both `TKey` and `TValue` unresolved, so the call chain collapses to `array<array-key, mixed>`.

## Environment

- `psalm/plugin-laravel`: `dev-master`
- PHP: 8.4

## Workaround

```php
/** @var array<string, string> $titlesBySlug */
$titlesBySlug = $author->articles()->pluck('title', 'slug')->all();
return $titlesBySlug;
```

…or splitting into two `pluck` calls and recombining, which is strictly worse.

## Related

- #486 — closed; covered single-argument `pluck('column')` value inference. The fix there did not extend to the two-argument form, so the key side is still `array-key` and the value side falls back to `mixed`.

alies-dev:
## Issue to Solve

Two-argument `Collection::pluck($value, $key)` collapses to `Collection<array-key, mixed>` and `->all()` to `array<array-key, mixed>`, forcing `LessSpecificReturnStatement` on callers that declare the narrowed shape as the return type. The same problem appears for `$relation->pluck(...)` even when the relation's models have full `@property` metadata.

## Related

Closes #967. Sibling of #486 (single-argument `pluck` value narrowing).

## Solution Description

Two-axis narrowing for `pluck()`:

- **Key column.** `ModelPropertyResolver::resolvePluckReturnType` resolves the second argument's `@property` type. When it is a subset of `array-key` (e.g. `string`, `int<0, max>`, `non-empty-string`), it becomes the new `TKey`. When it is not (e.g. `CarbonInterface|null`), the resolver falls back to `array-key` so the resulting `Collection<TKey, TValue>` never receives an invalid key type.
- **Relation chains.** Psalm rewrites `$relation->pluck(...)` through `@mixin Builder<TRelatedModel>`, but the template stays unsubstituted by the time the `MethodReturnTypeProvider` fires (only `TRelatedModel as Model` survives). The resolver now takes the call's LHS expression as a fallback input and re-extracts the concrete model from the LHS atomic's generic params. `BuilderPluckHandler` and `CollectionPluckHandler` pass `$stmt->var` for `MethodCall` nodes; static calls keep the existing template-parameter path.

Verification: type tests in `tests/Type/tests/Builder/BuilderPluckTest.phpt` cover Builder, Eloquent `Collection`, `HasMany`, `BelongsTo`, and `HasOne` plus negative cases (dynamic key column, non-array-key `@property`, unknown column). `composer test:type`, `composer psalm`, and `composer test:unit` all clean.

## Checklist

- [x] Tests cover the change (type tests in `tests/Type/tests/Builder/BuilderPluckTest.phpt`)

### DIFF
--- a/src/Handlers/Collections/CollectionPluckHandler.php
+++ b/src/Handlers/Collections/CollectionPluckHandler.php
@@ -46,6 +46,9 @@ final class CollectionPluckHandler implements MethodReturnTypeProviderInterface
             return null;
         }
 
+        $stmt = $event->getStmt();
+        $lhsExpr = $stmt instanceof \PhpParser\Node\Expr\MethodCall ? $stmt->var : null;
+
         // Collection<TKey, TValue> — TValue (the model) is template param at index 1.
         // For non-Model collections, resolvePluckReturnType returns null and Psalm uses
         // its default type inference.
@@ -55,6 +58,7 @@ final class CollectionPluckHandler implements MethodReturnTypeProviderInterface
             modelTemplateIndex: 1,
             nodeTypeProvider: $event->getSource()->getNodeTypeProvider(),
             codebase: $event->getSource()->getCodebase(),
+            lhsExpr: $lhsExpr,
         );
     }
 }
--- a/src/Handlers/Eloquent/BuilderPluckHandler.php
+++ b/src/Handlers/Eloquent/BuilderPluckHandler.php
@@ -42,6 +42,9 @@ final class BuilderPluckHandler implements MethodReturnTypeProviderInterface
             return null;
         }
 
+        $stmt = $event->getStmt();
+        $lhsExpr = $stmt instanceof \PhpParser\Node\Expr\MethodCall ? $stmt->var : null;
+
         // Builder<TModel> — TModel is template param at index 0
         return ModelPropertyResolver::resolvePluckReturnType(
             args: $event->getCallArgs(),
@@ -49,6 +52,7 @@ final class BuilderPluckHandler implements MethodReturnTypeProviderInterface
             modelTemplateIndex: 0,
             nodeTypeProvider: $event->getSource()->getNodeTypeProvider(),
             codebase: $event->getSource()->getCodebase(),
+            lhsExpr: $lhsExpr,
         );
     }
 }
--- a/src/Util/ModelPropertyResolver.php
+++ b/src/Util/ModelPropertyResolver.php
@@ -112,7 +112,7 @@ final class ModelPropertyResolver
         // still contain the *unsubstituted* template (TTemplateParam) instead of the
         // concrete model. Fall back to inspecting the call's LHS expression, whose
         // type carries the concrete generic arguments.
-        if ($modelClass === null && $lhsExpr !== null) {
+        if ($modelClass === null && $lhsExpr instanceof \PhpParser\Node\Expr) {
             $modelClass = self::extractModelFromLhsType(
                 $nodeTypeProvider->getType($lhsExpr),
                 $modelTemplateIndex,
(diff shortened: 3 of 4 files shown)


alies-dev:
## PR delta on real-world Laravel apps

Ran psalm-plugin-laravel against 14 Laravel apps (master vs `1b2a09e`).

| App              | ΔIssues | ΔCov   | ΔTime |
|------------------|--------:|-------:|------:|
| aureuserp        |      +0 |  +0.00 |  +0.01s |
| bagisto          |      +0 |  +0.00 |  -0.10s |
| bookstack        |      +0 |  +0.00 |  -0.19s |
| coolify          |      +0 |  +0.00 |  -0.57s |
| corcel           |      +0 |  +0.00 |  -0.05s |
| monica           |      +0 |  +0.00 |  -0.46s |
| pixelfed         |      +0 |  -0.00 |  -0.11s |
| solidtime        |      +0 |  +0.00 |  -0.28s |
| spatie-dashboard |      +0 |  +0.00 |  -0.16s |
| tastyigniter     |      +0 |  +0.00 |  -0.06s |
| unit3d           |      +0 |  +0.00 |  -0.26s |
| vito             |      +0 |  +0.00 |  -0.26s |
| laravel-excel    |      +0 |  +0.00 |  -0.11s |
| filament         |      +2 |  -0.00 |  -0.41s |
| **Total / Avg**  | **+2** | **-0.00** | **-3.01s** |

### Issue type movements

| Issue type | Base | PR | Δ |
|------------|-----:|---:|--:|
| MixedArgument | 7564 | 7565 | +1 |
| MixedArgumentTypeCoercion | 778 | 779 | +1 |

Net change: +2 issues in filament. Narrower `pluck()` return types surface 2 downstream call sites that previously hid behind `array<array-key, mixed>` and now flow as concrete types into APIs expecting different shapes. Pre-existing latent issues in filament, surfaced by improved inference. Coverage and time both neutral.

Issue #977, opened 2026-05-20
Stub: Filesystem contract missing url(), download(), directoryExists()


alies-dev:
Extends #973 (specifically \`url\`). Same root cause, broader scope.

Counts in invoiceninja:
- \`Illuminate\Contracts\Filesystem\Filesystem::url\` — 6 occurrences
- \`Illuminate\Contracts\Filesystem\Filesystem::download\` — 5 occurrences
- \`Illuminate\Contracts\Filesystem\Filesystem::directoryExists\` — 1 occurrence
- More likely surfaces in other projects (\`temporaryUrl\`, \`temporaryUploadUrl\`, \`putFile\`, etc.)

All of these live on the concrete \`Illuminate\Filesystem\FilesystemAdapter\` (the runtime return of \`Storage::disk()\`) but not on the \`Filesystem\` interface.

**Fix vector (mirrors Larastan):** extend \`stubs/common/Contracts/Filesystem/Filesystem.phpstub\` (new file) with the missing methods, or narrow the \`Storage\` facade @method static returns to \`FilesystemAdapter\` (cleaner, fixes all variants at once).

Once landed, the reproducer in #973 / \`tests/Type/tests/Facades/StorageDiskUrlTest.phpt\` can have its \`--EXPECTF--\` flipped to empty.

alies-dev:
Related: #802 (`Storage::disk()` returns the `Filesystem` contract; concrete-only methods unreachable). **Filesystem stub-completeness cluster** — complementary, not a duplicate:
- This issue: augment the `Filesystem` **contract stub** with `url()`/`download()`/`directoryExists()`/etc. — reaches DI-injected `function foo(Filesystem $fs)` call sites.
- #802: narrow `Storage::disk()`'s **return** to the concrete adapter — reaches `Storage::disk(...)->...` call sites.

Keep both open; coordinate the stub work so contract additions and concrete-return narrowing stay consistent.

alies-dev:
Closing this.

Re-checked against invoiceninja on the current `dev-master` (Psalm 7). Every real occurrence is the idiomatic `Storage::disk($x)->url()` / `->download()` / `->directoryExists()` form, and all of them now resolve cleanly: #802 narrows `disk()`/`drive()` to the concrete `Illuminate\Filesystem\FilesystemAdapter`, which carries these methods. A full scan of the affected files reports zero `UndefinedMethod` on `url`/`download`/`directoryExists`.

[...]

PR #982, opened 2026-05-21
Narrow `Storage::disk()` return to `Cloud` for cloud-driver disks

### LINKED ISSUE #973: Stub: Storage::disk()->url() — Filesystem contract lacks url() method
**Reproducer:** \`tests/Type/tests/Facades/StorageDiskUrlTest.phpt\` (#971).

\`\`\`php
Storage::disk('local')->url('foo.png');
// UndefinedInterfaceMethod: Method Illuminate\\Contracts\\Filesystem\\Filesystem::url does not exist
\`\`\`

The Storage facade's \`@method static disk(...)\` declares the return type as \`\\Illuminate\\Contracts\\Filesystem\\Filesystem\`, which does not have \`url()\`. The \`url()\` method lives on \`Cloud extends Filesystem\` and on the concrete \`FilesystemAdapter\`. At runtime, every disk driver returns a \`FilesystemAdapter\` so the call always succeeds.

**Source:** invoiceninja \`app/Console/Commands/BackupUpdate.php\` (3 occurrences).

**Fix vector (mirrors Larastan):** add \`stubs/common/Support/Facades/Storage.phpstub\` overriding \`disk()\` to return \`\\Illuminate\\Filesystem\\FilesystemAdapter\`. Same treatment for \`drive()\` and \`build()\`. Then flip the EXPECTF in the linked test to empty.

alies-dev:
## Issue to Solve

`Storage::disk('s3')->url($path)` raised `UndefinedInterfaceMethod` even though the call works at runtime. Laravel declares `Storage::disk()` (and `drive()`) as returning `\Illuminate\Contracts\Filesystem\Filesystem`, but `url()` lives on the `\Illuminate\Contracts\Filesystem\Cloud extends Filesystem` sub-contract. Every disk type returns a `FilesystemAdapter` at runtime, so the contract-level error is statically valid yet practically a nuisance for projects on S3.

## Related

Closes #973.

Partially mitigates #977 (only the `url()` case; `download()`, `directoryExists()`, `temporaryUrl()`, `temporaryUploadUrl()` live exclusively on the concrete `FilesystemAdapter`, not on `Cloud`, so cannot be addressed at the contract layer without lying about Laravel's contract design).

## Solution Description

**Layer choice.** Larastan's fix vector blanket-narrows `disk()` to `\Illuminate\Filesystem\FilesystemAdapter`. That silently permits `Storage::disk('local')->url(...)` even though the file-only `Filesystem` contract does not declare `url()` — conflating Laravel's intentional Cloud/Filesystem split. Projects that swap a `local` disk for a custom file-only adapter would silently break.

**This PR narrows at the correct layer:** a new `StorageHandler` reads `filesystems.disks.<name>.driver` from the booted app's config and only emits `Cloud` when the driver is `'s3'` (the sole built-in `createXDriver()` typed `@return Cloud` in `FilesystemManager`). `local` / `ftp` / `sftp` / `scoped` keep the contract-correct `Filesystem` return type, and `Storage::disk('local')->url(...)` remains a correctly reported `UndefinedInterfaceMethod`.

**Surface coverage.** The handler registers for three receivers via `getClassLikeNames()`:
- `\Illuminate\Support\Facades\Storage` (facade `__callStatic` path)
- `\Illuminate\Filesystem\FilesystemManager` (DI target)
- `\Illuminate\Contracts\Filesystem\Factory` (DI by interface)

**Why a paired `MethodParamsProviderInterface`.** Registering only `MethodReturnTypeProviderInterface` on the `Storage` facade crashes Psalm 7 with `UnexpectedValueException: Cannot get method params for Illuminate\Support\Facades\Storage::disk` because `disk()` exists only as `@method` on the facade — same failure mode documented on `AuthHandler::getMethodParams()`. The override applies only to the facade FQCN; non-facade receivers fall through to Psalm's native param derivation.

**Pattern conformance.** Mirrors `src/Handlers/Auth/` faithfully: singleton config analyzer, paired return-type/params provider, `require_once` + `registerHooksFromClass()` in `Plugin::registerHandlers()`.

**Performance.** All hot-path allocations cached as static singletons:
- The `Cloud` return `Union` is built once per worker and reused.
- The facade params override builds the `\UnitEnum|string|null` `Union` + `FunctionLikeParameter` once.
- `FilesystemConfigAnalyzer` memoises `getDriverForDisk()` and `getDefaultDisk()` lookups so the dotted-path config walk runs at most once per disk name.

Method-name filter (`isset(self::DISK_METHODS[$lowercase_name])`) is the cheapest possible reject for the dominant non-`disk` call path.

**Soundness.** Conservative narrow only: `'s3'` literal-equals to `Cloud`, everything else falls through to no-narrow. Custom drivers registered via `Storage::extend('foo', ...)` that return Cloud lose the narrowing (closure return type is opaque to static analysis), but no file-only driver is ever mis-narrowed to Cloud — which would silence a real runtime `url()` error.

**Out of scope (deliberate gaps).**
- `UnitEnum` literal disk-name arguments (`Storage::disk(Disks::S3)`) fall through with no narrowing — sound, but a missed opportunity. Could be addressed with a `ClassConstFetch` walk against the scanner's enum case storage.
- `Storage::build(['driver' => 's3', ...])` array-literal narrowing — runtime returns `Cloud`, but rare enough in practice to defer.
- `scoped` driver pointing at an s3 parent stays `Filesystem` per Laravel's own `@return` annotation on `createScopedDriver()` — mirroring Laravel's source-level imprecision.

**Verification (10 type-test scenarios):**
- `Storage::disk('s3')->url(...)` → narrowed to `Cloud`, no error.
- `Storage::drive('s3')->url(...)` → same via alias.
- `Storage::disk('local')->url(...)` → stays `Filesystem`, correctly errors.
- `Storage::disk()` no-arg → resolves default disk via config.
- `Storage::disk(null)` literal-null → resolves default disk via config.
- `Storage::disk($dynamic)` → opaque, stays `Filesystem` (no over-narrowing).
- `Storage::disk('disk-not-in-config')` → unknown disk, stays `Filesystem`.
- `Storage::disk(42)` → emits `InvalidArgument` (pins the facade params override).
- DI `FilesystemManager $mgr; $mgr->disk('s3')->url(...)` → narrowed to `Cloud`.
- DI `Factory $f; $f->disk('s3')->url(...)` → narrowed to `Cloud`.

Two `/psalm-review` rounds across 8 specialised agents (code, psalm-expert, architecture, performance, laravel-expert, simplify, tests, types) closed all Critical/Important findings before this PR.

Suites: type 408/408, unit 632/632 (1 skipped, pre-existing), `composer psalm` exit 0, `composer cs` exit 0.

## Checklist

- [x] Tests cover the change (10-scenario type test at `tests/Type/tests/Facades/StorageDiskUrlTest.phpt`)

### DIFF
--- a/src/Handlers/Filesystem/FilesystemConfigAnalyzer.php
+++ b/src/Handlers/Filesystem/FilesystemConfigAnalyzer.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Filesystem;
+
+use Illuminate\Contracts\Config\Repository as ConfigRepository;
+use Psalm\LaravelPlugin\Providers\ConfigRepositoryProvider;
+
+/**
+ * Reads `config/filesystems.php` to determine which disk maps to which driver.
+ * The driver name is then matched against Laravel's built-in `createXDriver()`
+ * methods to decide whether the disk returns a {@see \Illuminate\Contracts\Filesystem\Cloud}
+ * (ships `url()`) or the base {@see \Illuminate\Contracts\Filesystem\Filesystem}
+ * (file-only, does NOT declare `url()`).
+ *
+ * Custom drivers registered via `Storage::extend('foo', ...)` may also return
+ * `Cloud`, but static analysis cannot peek inside the registered closure to
+ * tell. We only narrow for built-ins (currently `'s3'`), which keeps us sound:
+ * a custom cloud driver loses the narrowing but no file-only driver is ever
+ * mis-narrowed to Cloud (which would silence a real `url()` runtime error).
+ *
+ * Mirrors the pattern of {@see \Psalm\LaravelPlugin\Handlers\Auth\AuthConfigAnalyzer}:
+ * single per-process instance, lazy singleton, immutable wrap of the booted
+ * Laravel app's config repository.
+ *
+ * @internal
+ */
+final class FilesystemConfigAnalyzer
+{
+    private static ?FilesystemConfigAnalyzer $instance = null;
+
+    private ?string $default_disk_cache = null;
+
+    private bool $default_disk_loaded = false;
+
+    /** @var array<string, string|null> */
+    private array $driver_cache = [];
+
+    /** @psalm-mutation-free */
+    private function __construct(private readonly ConfigRepository $config) {}
+
+    public static function instance(): self
+    {
+        if (!self::$instance instanceof self) {
+            self::$instance = new self(ConfigRepositoryProvider::get());
+        }
+
+        return self::$instance;
+    }
+
+    public function getDefaultDisk(): ?string
+    {
+        if (!$this->default_disk_loaded) {
+            /** @var mixed $disk */
+            $disk = $this->config->get('filesystems.default');
+            $this->default_disk_cache = \is_string($disk) ? $disk : null;
+            $this->default_disk_loaded = true;
+        }
+
+        return $this->default_disk_cache;
+    }
+
+    /**
+     * Return the configured driver name for a disk, or null if the disk is not
+     * declared (or its `driver` key is missing/non-string). Memoised — repeat
+     * calls for the same disk skip the dotted-path config walk.
+     */
+    public function getDriverForDisk(string $disk): ?string
+    {
+        if (\array_key_exists($disk, $this->driver_cache)) {
+            return $this->driver_cache[$disk];
+        }
+
+        /** @var mixed $driver */
+        $driver = $this->config->get("filesystems.disks.{$disk}.driver");
+
+        return $this->driver_cache[$disk] = \is_string($driver) ? $driver : null;
+    }
+}
--- a/src/Handlers/Filesystem/FilesystemConfigAnalyzer.php
+++ b/src/Handlers/Filesystem/FilesystemConfigAnalyzer.php
@@ -52,7 +52,6 @@ final class FilesystemConfigAnalyzer
     public function getDefaultDisk(): ?string
     {
         if (!$this->default_disk_loaded) {
-            /** @var mixed $disk */
             $disk = $this->config->get('filesystems.default');
             $this->default_disk_cache = \is_string($disk) ? $disk : null;
             $this->default_disk_loaded = true;
@@ -72,7 +71,6 @@ final class FilesystemConfigAnalyzer
             return $this->driver_cache[$disk];
         }
 
-        /** @var mixed $driver */
         $driver = $this->config->get("filesystems.disks.{$disk}.driver");
 
         return $this->driver_cache[$disk] = \is_string($driver) ? $driver : null;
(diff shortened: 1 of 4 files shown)


alies-dev:
## PR delta benchmark

Ran `/psalm-pr-delta` across 17 real-world Laravel apps. Base: `merge-base(master)` = `05791595`. Head: `c794abbe`.

### Per-app delta

| aureuserp        |    0 |    0 |    +0 |  +0.00 | +0.07s |
| bagisto          |    0 |    0 |    +0 |  +0.00 | +0.03s |
| bookstack        |    0 |    0 |    +0 |  +0.00 | +0.32s |
| coolify          |    0 |    0 |    +0 |  -0.00 | -0.34s |
| corcel           |    0 |    0 |    +0 |  +0.00 | +0.07s |
| monica           |    0 |    0 |    +0 |  +0.00 | -0.14s |
| **pixelfed**     |  **1** |  **2** |  **-1** |  **+0.00** | **+0.92s** |
| solidtime        |    0 |    0 |    +0 |  +0.00 | -0.20s |
| spatie-dashboard |    0 |    0 |    +0 |  +0.00 | +0.86s |
| tastyigniter     |    0 |    0 |    +0 |  +0.00 | +0.33s |
| unit3d           |    0 |    0 |    +0 |  +0.00 | -0.99s |
| vito             |    0 |    0 |    +0 |  +0.00 | +0.25s |
| laravel-excel    |    0 |    0 |    +0 |  +0.00 | +0.45s |
| filament         |    0 |    0 |    +0 |  +0.00 | -0.24s |
| laravel-backup   |    0 |    0 |    +0 |  +0.00 | +0.41s |
| laravel-permission |  0 |    0 |    +0 |  +0.00 | -0.04s |
| laravel-socialite |   0 |    0 |    +0 |  +0.00 | +0.41s |
| **Total / Weighted** | **1** | **2** | **-1** | **+0.00** | **+2.17s** |

_ΔCov weighted by base `wall_seconds` (proxy for codebase size). ΔTime noise on this hardware is ±1s per app._

### Issue movements

| `MixedAssignment` | 0 | 1 | -1 |
| `PossiblyFalseArgument` | 1 | 0 | +1 |
| `UndefinedInterfaceMethod` | 0 | 1 | -1 |

### Pixelfed detail

Single call site `app/Console/Commands/BackupToCloud.php:68` drives every delta.

**Before:**
- `UndefinedInterfaceMethod`: `Illuminate\Contracts\Filesystem\Filesystem::url does not exist`
- `MixedAssignment`: `Unable to determine the type that $url is being assigned to` (cascading from the line above)

**After:**
- `Storage::disk('s3')->url(...)` now narrows to `Cloud::url(string)` → primary false positive gone
- `MixedAssignment` cascade gone — `$url` is now typed `string`
- New `PossiblyFalseArgument`: argument 1 of `Cloud::url` cannot be `false`, possibly `string` value expected — **this is a real latent bug in pixelfed surfaced by the better narrowing**

### Interpretation

- Target issue (#973) eliminated where predicted, zero collateral damage across the other 16 apps.
- The single `+1` is a true positive that was previously masked by the `UndefinedInterfaceMethod` short-circuit — a net improvement, not a regression.
- No perf regression: total +2.17s across 17 apps is within run-to-run noise.

PR #983, opened 2026-05-21
Keep canonical `Auth` facade in narrowing handler list

alies-dev:
## Issue to Solve

After #981 merged, apps that trim `config/app.php`'s `aliases` array surface false-positive `UndefinedInterfaceMethod` errors on every `Auth::guard($name)->{statefulMethod}()` call. Concretely IxDF (which registers only `Vite` as an alias) sees:

```
  Auth::guard('member')->login($member);
  Method Illuminate\Contracts\Auth\Guard::login does not exist
```

Plus the same shape for `->logout()`, `->loginUsingId()`, and any other `StatefulGuard`-only method. ~10+ occurrences in IxDF alone.

## Related

Follows #981 (regression fix).

## Solution Description

`AuthMethodHandler::getClassLikeNames()` sourced the canonical facade FQCN (`Illuminate\Support\Facades\Auth`) solely through `FacadeMapProvider::getFacadeClasses(AuthManager::class)`. `FacadeMapProvider` iterates Laravel's `AliasLoader`, so its lookup returns an empty list for `AuthManager` in apps that trim the default aliases (e.g. IxDF registering only `Vite`). The canonical facade then silently disappeared from the handler list, and narrowing was skipped for every `Illuminate\Support\Facades\Auth::guard(...)` call site. The stub `@method` union (`Guard|StatefulGuard`) fell back to the failing `Guard` branch when chained with a `StatefulGuard`-only method.

Hardcode the canonical facade FQCN; keep the `FacadeMapProvider` spread for alias discovery on apps that DO register them (e.g. invoiceninja uses `use Auth;` resolving to the root-namespace alias).

Verification on real apps with this branch applied:

| App | Threads=1 | Default threads |
|---|---:|---:|
| IxDF `Guard::login` errors | 0 | 0 |
| invoiceninja `Guard::logout` errors | 0 | 0 |

`composer test:type` (413/413) and `composer psalm` clean.

## Checklist

- [x] Tests cover the change (existing AuthTest.phpt cases keep passing; the regression is real-world-app-only because Testbench registers the default aliases)

### DIFF
--- a/src/Handlers/Auth/AuthConfigAnalyzer.php
+++ b/src/Handlers/Auth/AuthConfigAnalyzer.php
@@ -11,6 +11,20 @@ final class AuthConfigAnalyzer
 {
     private static ?AuthConfigAnalyzer $instance = null;
 
+    /**
+     * Memoizes `getGuardFQCN()` so the per-call-site Psalm hook avoids
+     * re-walking `auth.guards.<name>.driver` through `Repository::get()`.
+     * The negative result (`null`) is cached too — `array_key_exists` is
+     * used at the call site to distinguish "not yet computed" from "known
+     * unresolvable" (custom driver / unknown guard).
+     *
+     * @var array<string, class-string<\Illuminate\Contracts\Auth\Guard>|null>
+     */
+    private array $guardFqcnCache = [];
+
+    private ?string $defaultGuardCache = null;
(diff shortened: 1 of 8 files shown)


alies-dev:
Wrong head branch — would revert #982. Reopening from clean branch off current master.

Issue #899, opened 2026-05-10
Extend Macro support: facade propagation, Contracts interfaces, multi-target facades, docblock-aware types


alies-dev:
Follow-ups identified during the deep-dive on Larastan's macro implementation (`larastan/larastan/src/Methods/MacroMethodsClassReflectionExtension.php` + `Macro.php`) while building the macro foundation in #894. These ideas extend the foundation rather than fix it, and each can land independently.

The macro foundation in #894 (Strategy B — runtime reflection) is in place. The umbrella tracking issue is #758.

---

## 1. Docblock-aware closure type extraction (Strategy C, AST scan)

**Today:** `MacroRegistry::buildDefinition()` reads param + return types via `ReflectionType::__toString()` and `Type::parseString()`. Native PHP types only — no docblock annotations.

**Larastan equivalent:** `ClosureTypeFactory::fromClosureObject($closure)`:

1. Reflects the closure → `ReflectionFunction::getFileName()` + `getStartLine()`.
2. Re-parses that source file with PHP-Parser.
3. Locates the closure node at the given offset.
4. Recovers param and return types from `@param` / `@return` docblocks (and native types).

**Why it matters:** users who write rich closures lose precision today. Examples:

```php
Builder::macro('whereLike', function (string $column, string $value): static {
    /** @return Builder<TModel> */
    return $this->where($column, 'like', "%{$value}%");
});

Stringable::macro('toCollection', function (string $separator): Collection {
    /** @return Collection<int, string> */
    return collect(explode($separator, $this->value));
});
```

We currently surface `Builder` and `Collection` (loose). With AST recovery we'd surface `Builder<TModel>` and `Collection<int, string>`.

**Suggested implementation:** when `\ReflectionFunction::getFileName()` is a path Psalm has scanned, look up the closure in `FunctionLikeStorage` keyed by `(file, line)` and copy `params` + `return_type` directly. Falls back to today's native-types extraction when the closure's source isn't scanned (e.g. eval'd, or registered in a file outside `<projectFiles>`).

**Scope:** moderate. Largely additive (a new branch in `buildDefinition()` that prefers Psalm's storage over reflection). Tests need a fixture closure with explicit docblock annotations.

**Reference:** `larastan/src/Methods/MacroMethodsClassReflectionExtension.php` lines 144–171.

---

## 2. `Illuminate\Contracts\*` interface dispatch

[...]

PR #989, opened 2026-05-22
Use `Macroable` return type info from docblock Psalm storage

alies-dev:
## Issue to Solve

`MacroRegistry` builds `MacroDefinition` instances from runtime reflection of `Macroable::$macros`. For Closure callables this means `ReflectionFunction::getReturnType()` / `getParameters()->getType()` — native PHP types only. Docblock annotations on the registered closure are invisible: `@return Collection<int, string>`, `@param positive-int $count`, `@return non-empty-string`, generic templates, etc. all collapse to whatever the native type happens to be (often `mixed` when the closure has no native return).

## Related

Closes one branch of #899 (idea #1, "Docblock-aware closure type extraction" / Strategy C for the closure case).

## Solution Description

For Closure callables, `MacroRegistry::buildDefinition()` now consults Psalm's pre-scanned `FunctionLikeStorage` before falling back to reflection.

`MacroRegistry::recoverClosureStorage()` looks the closure up in the file's storage table by file + line. The closure-id format Psalm uses (`<lowercase-path>:<line>:<startFilePos>:-:closure`, set in `FunctionLikeNodeScanner` line 1140) includes a byte offset that reflection doesn't expose, so the lookup matches on the `<path>:<line>:` prefix and `:-:closure` suffix. If exactly one closure starts on that line we use it; zero or multiple matches fall back to reflection.

`MacroRegistry::buildDefinitionFromStorage()` copies the storage's `params` (via `clone` to preserve `out_type` / `default_type` / `attributes` / `has_docblock_type` / etc. that the constructor would drop) and `return_type` / `signature_return_type` directly into the new `MacroDefinition`. Defensive copy keeps the registry isolated from Psalm's source-of-truth storage; `Union` types are immutable in Psalm 7 so sharing those references is safe.

Coverage of the new path:

- App-level macros registered in `App\Providers\*::boot()`: in `<projectFiles>`, scanned.
- Test fixtures and bootstrap-time registrations through the `autoloader` attribute: Psalm's `Config::collectPredefinedFunctions` routes them through `addFileToDeepScan`, so they end up in `FileStorage` too.
- Stub-registered macros: scanned as part of the stub pass.

Falls back to reflection when storage is unavailable: vendor closures outside projectFiles, eval'd code, missing source files, or the Testbench-fallback case where the analysed package's own provider is never booted. That remaining gap is the broader Strategy C AST-scan follow-up tracked in `MacroHandler`'s TODO note (idea #1's vendor coverage + the AST-scan-proper item).

`MacroRegistry::init()` now takes an optional `Codebase` parameter so unit tests (which don't boot Psalm) keep working unchanged via the `null` default. `MacroHandler` passes `$event->getCodebase()` through.

### How verified

Three layers:

- Type test: `tests/Type/tests/Macros/MacroDocblockClosureTest.phpt` (6 assertions) covers docblock return-type recovery (`@return non-empty-string`), generic recovery (`@return Collection<int, string>`), `@param positive-int` narrowing including a hard `InvalidArgument` negative, plus static-dispatch and subclass-instance coverage of the same recovered macro.
- Type test fixtures: two new closures in `tests/Type/macro-fixtures.php` (`docblockReturnTest`, `docblockGenericTest`) registered with docblocks that have no native equivalent.
- Existing tests retained: `MacroDispatchTest.phpt`, `MacroFluentStaticTest.phpt`, `BuilderMacroTest.phpt`, `MacroFacadeDispatchTest.phpt` continue to pass, locking in the unchanged reflection-only behaviour for callables that don't (or can't) recover storage.
- `composer test:type` (full): 415/415 pass. `composer psalm`: clean. `composer cs`: clean.

## Checklist
- [x] Documentation is updated (TODO note in `MacroHandler` updated to reflect partial closure-of Strategy C)

### DIFF
--- a/src/Handlers/Magic/MacroHandler.php
+++ b/src/Handlers/Magic/MacroHandler.php
@@ -88,6 +88,16 @@ use Psalm\Storage\MethodStorage;
  * because `Builder<User>` is the direct dispatch target.
  *
  * TODO Strategy C / follow-up:
+ * - **Docblock-aware closure types** (issue #899 idea #1): IMPLEMENTED for closures
+ *   whose source Psalm has scanned. {@see MacroRegistry::recoverClosureStorage()}
+ *   looks up the closure in Psalm's pre-scanned {@see \Psalm\Storage\FunctionLikeStorage}
+ *   by file + line and lifts docblock-merged `@return` / `@param` types into the
+ *   pseudo-method. Coverage: autoloader files
+ *   ({@see \Psalm\Config::collectPredefinedFunctions} routes them through
+ *   `addFileToDeepScan`), projectFiles, and stubs. Out of scope: vendor closures
+ *   when those paths sit outside `<projectFiles>`, eval'd code, and the
+ *   Testbench-fallback case where the analysed package's own provider is never
+ *   booted (covered by the AST-scan follow-up below).
  * - **AST scan** (Strategy C proper): cover macros registered in the analysed
  *   package's own provider when running on a package without `bootstrap/app.php`
  *   (Testbench fallback path — issue #766). Walk every literal
@@ -203,7 +213,7 @@ final class MacroHandler implements AfterCodebasePopulatedInterface
     {
         $codebase = $event->getCodebase();
 
-        MacroRegistry::init($codebase->progress);
+        MacroRegistry::init($codebase->progress, $codebase);
 
         $macroableClasses = MacroRegistry::getKnownMacroableClasses();
         if ($macroableClasses === []) {
--- a/src/Providers/MacroRegistry.php
+++ b/src/Providers/MacroRegistry.php
@@ -294,7 +294,7 @@ final class MacroRegistry
         // generic templates — none of which reflection can surface. Falls back to
         // reflection-based extraction when storage is unavailable (no codebase,
         // closure source not scanned, or multiple closures share the start line).
-        if ($callable instanceof \Closure && $codebase !== null) {
+        if ($callable instanceof \Closure && $codebase instanceof \Psalm\Codebase) {
             $closureStorage = self::recoverClosureStorage($reflection, $codebase);
             if ($closureStorage instanceof FunctionLikeStorage) {
                 return self::buildDefinitionFromStorage(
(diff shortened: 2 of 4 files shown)


alies-dev:
## App benchmark delta (17 public apps)

Ran `/psalm-pr-delta` against `merge-base(master) = a420452c` → `3c9fbb6e` (PR head).

|---|---:|---:|---:|---:|---:|
| aureuserp | 0 | 0 | +0 | +0.00 | -0.23s |
| bagisto | 0 | 0 | +0 | +0.00 | -0.10s |
| bookstack | 0 | 0 | +0 | +0.00 | -0.26s |
| coolify | 0 | 0 | +0 | +0.00 | -0.96s |
| corcel | 0 | 0 | +0 | +0.00 | -0.47s |
| monica | 0 | 0 | +0 | +0.00 | -0.02s |
| pixelfed | 0 | 0 | +0 | +0.00 | -0.08s |
| solidtime | 0 | 0 | +0 | +0.00 | +0.17s |
| spatie-dashboard | 0 | 0 | +0 | +0.00 | -0.56s |
| tastyigniter | 0 | 0 | +0 | +0.00 | +0.22s |
| unit3d | 0 | 0 | +0 | +0.00 | +0.59s |
| vito | 0 | 0 | +0 | +0.00 | -0.62s |
| laravel-excel | 0 | 0 | +0 | +0.00 | +0.11s |
| filament | 0 | 0 | +0 | +0.00 | -1.50s |
| laravel-backup | 0 | 0 | +0 | +0.00 | -0.60s |
| laravel-permission | 0 | 0 | +0 | +0.00 | -0.62s |
| laravel-socialite | 0 | 0 | +0 | +0.00 | -0.21s |
| **Total / Weighted** | **0** | **0** | **+0** | **+0.00** | **-5.14s** |

No issue movement, no coverage delta, no measurable wall-time regression. Expected: the new storage-recovery branch is gated on `$callable instanceof \Closure && $codebase !== null`, and none of the benchmarked apps register Macroable closures with docblock-only types in scanned paths. The fall-back to reflection is unchanged.

The behaviour change is purely additive. The WIN path is exercised by the new type test `MacroDocblockClosureTest.phpt`; in-the-wild gains will surface as apps add docblock annotations to their `*::macro(...)` closures (Filament's `ComponentAttributeBag::macro`, Spatie's `Carbon::macro`, app-level `Builder::macro` callbacks, etc.).

Issue #758, opened 2026-04-17
Discover ::macro() registrations on Macroable classes and synthesize virtual methods


alies-dev:
## Description

Laravel's `Macroable` trait is used across the framework (`Illuminate\View\ComponentAttributeBag`, `Illuminate\Support\Collection`, `Illuminate\Support\Str`, `Illuminate\Support\Stringable`, `Illuminate\Http\Request`, `Illuminate\Http\Response`, `Illuminate\Routing\Router`, `Illuminate\Database\Query\Builder`, `Illuminate\Database\Eloquent\Builder`, etc.) to let packages and applications register methods at runtime via `::macro('name', fn (...) => ...)`.

The plugin currently does not discover these registrations, so every call to a registered macro reports as `UndefinedMagicMethod` and every chained call after it cascades into `MixedMethodCall` / `PossiblyInvalidArgument`.

## Scope: broader than #648

#648 is open for `Builder::macro` specifically on relation chains. This issue is the larger umbrella: macros are not resolved **anywhere** on **any** Macroable class. Closing this would close #648 as a special case.

## Reproduction

Surfaced in a v4.8.3 audit of [filamentphp/filament](https://github.com/filamentphp/filament) (`errorLevel=1`). Example hotspots:

```php
// packages/actions/src/ActionGroup.php:764
// packages/forms/src/Components/Concerns/HasExtraInputAttributes.php:37
// packages/infolists/src/Components/IconEntry.php:298 etc.

$attributes->color($color);        // UndefinedMagicMethod: color does not exist
$attributes->grid($cols);          // UndefinedMagicMethod: grid does not exist
$attributes->gridColumn($cols);    // UndefinedMagicMethod: gridColumn does not exist
$attributes->merge([...]);         // PossiblyInvalidArgument when narrowed by chained macros
```

These are all registered through `ComponentAttributeBag::macro(...)` in Filament (and other packages) service providers.

## Impact

In Filament v4.8.3 alone, directly attributable:

| Count | Issue | Surface |
| ---: | --- | --- |
| 24 | `UndefinedMagicMethod` | `ComponentAttributeBag::color` |
| 3 | `UndefinedMagicMethod` | `ComponentAttributeBag::grid` |
| 3 | `UndefinedMagicMethod` | `ComponentAttributeBag::gridColumn` |
| 14 | `PossiblyInvalidArgument` | `ComponentAttributeBag::merge` (shape narrowing downstream of the above) |

[...]

PR #994, opened 2026-05-22
Recover macro closure docblocks from vendor packages via AST scan

### LINKED ISSUE #991: Macros: AST scan for closures in unscanned paths (vendor packages)
Follow-up to #989. The storage-based docblock recovery shipped there only fires when Psalm has scanned the closure's source file. For closures living in `vendor/` outside `<projectFiles>`, the storage lookup misses and the registry falls back to reflection-only types — discarding the docblock entirely.

Larastan's equivalent `ClosureTypeFactory::fromClosureObject` parses the source file directly with PHP-Parser on demand, so it doesn't depend on the analyser having scanned the file ahead of time.

## Real-world example

`vendor/inertiajs/inertia-laravel/src/ServiceProvider.php` (current master):

```php
protected function registerRouterMacro(): void
{
    /**
     * @param  array<array-key, mixed>  $props
     */
    Router::macro('inertia', function ($uri, $component, $props = []) {
        return $this->match(['GET', 'HEAD'], $uri, '\\'.Controller::class)
            ->defaults('component', $component)
            ->defaults('props', $props);
    });
}
```

The closure has no native param types and no native return type. The docblock narrows `$props` to `array<array-key, mixed>`. Today's behaviour with a typical app config (`<projectFiles><directory name="app"/></projectFiles>`):

1. Plugin boots Laravel → Inertia provider runs → macro registered. `MacroRegistry::init` reflects it.
2. `MacroRegistry::recoverClosureStorage`: `getFileName()` → `vendor/inertiajs/.../ServiceProvider.php`. `file_storage_provider->has()` → **false** (not in `<projectFiles>`).
3. Falls back to reflection: all three params type-erased to `mixed`, return type `mixed`.

Result on the call site `Route::inertia('/dashboard', 'Dashboard')`: parameters accept anything, return is `mixed`.

## What AST scan would cover

After the storage-lookup miss, parse the source file with `nikic/php-parser`, locate the closure node at the reflected file + line, hand the attached Doc comment to Psalm's `CommentAnalyzer` (or `FunctionLikeDocblockScanner`), and copy the parsed `@param` / `@return` Unions into the synthesised pseudo-method.

| Pattern | Storage path (#989) | AST scan |
|---|---|---|
| Closure in `app/Providers/*::boot()` with docblock | recovered | redundant (same result) |
| Closure in `vendor/*` with docblock-only types (Inertia case above) | reflection-only, docblock lost | recovered |
| Closure in `vendor/*` with native return type + `@param`-only narrowing | partial (native return only) | full |
| Closure registered via the Testbench-fallback path (#766) — user's own provider boot never runs | not even reflected at runtime | not addressed (separate problem: AST scan over the `*::macro()` call sites in the analysed package, tracked in #766) |
| `eval()`'d closure | `getFileName()` returns synthetic path | not addressable (no source to parse) |

## Suggested implementation

In `MacroRegistry::buildDefinition()`, after `recoverClosureStorage()` returns null and before falling back to `buildParameter()` + `reflectionTypeToUnion()`:

```php
$closureStorage = self::recoverClosureStorageFromSource($reflection, $codebase);
if ($closureStorage instanceof FunctionLikeStorage) {
    return self::buildDefinitionFromStorage($declaringClass, $name, $closureStorage);
}
```

The new helper:

1. Reads `getFileName()` + `getStartLine()` from the reflection (already plumbed).
2. Parses the file with `nikic/php-parser` (already in `composer.json` via Psalm).
3. Locates the `Closure` / `ArrowFunction` node at the matching line via a small `NodeVisitor`.
4. Hands the node's docblock to `CommentAnalyzer::extractFunctionDocblockInfo()` (or whichever facade Psalm 7 exposes).
5. Builds a `FunctionLikeStorage` (or shaped equivalent) from the parsed result.

Cache the parse result per (file, mtime) — Psalm already does per-file mtime caching for its own scan; mirroring that prevents re-parsing the same vendor file once per macro registered in it.

## Caveats

- Files Psalm doesn't have read access to (broken symlinks, removed packages with cached registrations) must degrade silently to reflection.
- Multiple closures on the same start line (rare) — same `null` fallback as the storage path.
- `eval()`'d closures: `getFileName()` returns a synthetic path; bail out cleanly.

## Scope

Medium. ~80 LoC + a cache layer + a fixture or two to lock in the vendor-recovery case (the existing test fixtures live in `<projectFiles>`-scanned territory, so they don't exercise this path).

## Reference

- Larastan: `larastan/src/Methods/MacroMethodsClassReflectionExtension.php` and `larastan/src/Types/ClosureTypeFactory.php`.
- Real-world target: `inertiajs/inertia-laravel/src/ServiceProvider.php` line ~111-119 (current master).
- Parent: #758 (umbrella), #899 (closed parent).

alies-dev:
## Issue to Solve

The storage-based closure docblock recovery shipped in #989 only fires when Psalm has scanned the closure's source file. For closures registered from `vendor/` packages outside `<projectFiles>`, the storage lookup misses and the registry falls back to reflection-only types, discarding the docblock entirely.

Motivating example (Inertia, current master):

```php
// vendor/inertiajs/inertia-laravel/src/ServiceProvider.php
/**
 * @param array<array-key, mixed> $props
 */
Router::macro('inertia', function ($uri, $component, $props = []) {
    return $this->match(['GET', 'HEAD'], $uri, '\\'.Controller::class)
        ->defaults('component', $component)
        ->defaults('props', $props);
});
```

The closure has no native param types and no native return type. The docblock narrows `$props` to `array<array-key, mixed>`. Under a typical project config (`<projectFiles><directory name="app"/></projectFiles>`), the closure's source file is never deep-scanned by Psalm and the docblock narrowing is lost.

## Related

Closes #991. Builds on #989 (storage-path recovery). Parent #758 (umbrella).

## Solution Description

Adds `Psalm\LaravelPlugin\Util\Ast\ClosureDocblockFactory` (analogue of Larastan's `ClosureTypeFactory::fromClosureObject`) which parses the closure's source file directly with `nikic/php-parser` on demand, locates the matching `Closure` / `ArrowFunction` by start line via a `ClosureDocblockIndexVisitor`, and lifts `@param` / `@return` Unions from the docblock attached either directly to the closure node OR to the immediately wrapping `Stmt\Expression`. Reflection still owns parameter order, by-ref, variadic, and defaults; the docblock contributes type narrowing only.

`MacroRegistry::buildDefinition()` now prioritises the AST path: AST first, then the pre-#989 storage path as second-tier fallback, then reflection-only. AST-first matters even for files Psalm did scan — Psalm's own scanner does not attach a `Stmt\Expression` docblock to an inner closure, so the Inertia-style pattern would otherwise miss even when storage exists for the file.

Caching:
- Per-(file, mtime) parse results so a vendor file registering N macros parses once per analysis run.
- Per-namespace-frame `Aliases` so N closures sharing one namespace allocate one shared `Aliases` instance.
- Negative cache entries for parse errors / unreadable files so a single broken file is not re-attempted N times.
- `MacroRegistry::reset()` cascades into `ClosureDocblockFactory::reset()` for test isolation.

Doc-attachment policy is deliberately strict: own docblock first, then the wrapping `Stmt\Expression` only. Walking further up the ancestor chain would risk attaching an outer function's `/** @return Foo */` to a nested closure.

Failure modes (file unreadable, parse error, no closure starts at the reflected line, multiple closures share the start line, no docblock) all degrade silently to the storage fallback or reflection. Vendor-package code paths should never surface plugin warnings for files the user does not control.

## Coverage

- **Type test** `tests/Type/tests/Macros/MacroAstScanVendorClosureTest.phpt` registers macros from `tests/Type/macro-fixtures-vendor-style.php`, a file deliberately excluded from `<projectFiles>`, `<stubs>`, and the `autoloader` attribute. It only reaches PHP via `require_once` from the autoloader file, mirroring the Inertia case. Exercises:
  - Full `@param positive-int` + `@return non-empty-string` narrowing through the synthesised pseudo-method.
  - Generic return-type recovery (`@return Collection<int, string>`) with `use`-alias resolution.
  - Matrix row 3: `@param`-only narrowing while the native string return survives.
  - Arrow-function syntax (`static fn(int $count) => ...`).
  - `InvalidArgument` rejection of non-positive ints on the docblock-narrowed parameter.
- **Unit test** `tests/Unit/Util/Ast/ClosureDocblockFactoryTest.php` locks in the wrapping-`Stmt\Expression` doc attachment, outer-function docblock isolation, the multi-closure-on-same-line ambiguity bail, and the no-docblock null return.

## Checklist

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -21,6 +21,13 @@ return RectorConfig::configure()
     // that never matches the lowercase method id. (::class IS valid in const arrays on
     // PHP 8.3+; the problem is the casing, not the const-context.)
     ->withSkipPath('src/Handlers/Rules/OctaneIncompatibleBindingHandler.php')
+    // The vendor-style macro fixture (PR #991 + PR #994) deliberately exercises
+    // closures without native return types or docblock `@return` so the
+    // AST-scan + body-inference paths are the only sources of narrowing.
+    // Rector's `typeDeclarations` set keeps "fixing" them back to declared
+    // return types, which silently turns the body-inference assertions into
+    // native-return-type assertions and the feature stops being tested.
+    ->withSkipPath('tests/Type/macro-fixtures-vendor-style.php')
     ->withPhpVersion(PhpVersion::PHP_82)
     ->withSets([PHPUnitSetList::PHPUNIT_120])
     ->withPreparedSets(deadCode: true, codingStyle: true, typeDeclarations: true, codeQuality: true, phpunitCodeQuality: true)
--- a/src/Providers/MacroRegistry.php
+++ b/src/Providers/MacroRegistry.php
@@ -323,7 +323,7 @@ final class MacroRegistry
         if ($callable instanceof \Closure) {
             $closureType = CachedClosureTypeFactory::fromClosureObject($callable);
 
-            if ($closureType === null && $codebase instanceof \Psalm\Codebase) {
+            if (!$closureType instanceof \Psalm\Type\Atomic\TClosure && $codebase instanceof \Psalm\Codebase) {
                 $closureStorage = self::recoverClosureStorage($reflection, $codebase);
                 if ($closureStorage instanceof FunctionLikeStorage) {
                     return self::buildDefinitionFromStorage(
--- a/src/Providers/MacroRegistry.php
+++ b/src/Providers/MacroRegistry.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Psalm\LaravelPlugin\Providers;
 
 use Psalm\Codebase;
+use Psalm\LaravelPlugin\Util\Ast\CachedClosureTypeFactory;
+use Psalm\LaravelPlugin\Util\Ast\ClosureTypeFactory;
 use Psalm\Progress\Progress;
 use Psalm\Storage\FunctionLikeParameter;
 use Psalm\Storage\FunctionLikeStorage;
@@ -244,6 +246,11 @@ final class MacroRegistry
     {
         self::$macros = [];
         self::$knownMacroableClasses = [];
+        // Cascade into the AST docblock cache — tests rely on this for isolation
+        // between runs, and the cache is keyed by realpath+mtime so leaking it
+        // across test cases would produce phantom hits on fixture files that
+        // were rewritten between runs.
+        CachedClosureTypeFactory::reset();
     }
 
     /**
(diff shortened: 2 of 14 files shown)


alies-dev:
## PR delta benchmark vs `master`

Ran `/psalm-pr-delta` from `merge-base(master)` (`b87ff094`) to PR HEAD (`ff955e33`) across 17 public benchmark apps.

### Per-app delta

|---|---:|---:|---:|---:|---:|
| aureuserp | 0 | 0 | +0 | +0.00 | +0.15s |
| bagisto | 0 | 0 | +0 | +0.00 | +0.81s |
| bookstack | 0 | 0 | +0 | +0.00 | +0.09s |
| coolify | 0 | 0 | +0 | +0.00 | +0.29s |
| corcel | 0 | 0 | +0 | +0.00 | -0.03s |
| monica | 0 | 0 | +0 | +0.00 | +0.69s |
| pixelfed | 0 | 0 | +0 | -0.00 | +0.83s |
| solidtime | 0 | 0 | +0 | +0.00 | -2.84s |
| spatie-dashboard | 0 | 0 | +0 | +0.00 | -0.53s |
| tastyigniter | 0 | 0 | +0 | +0.00 | -0.20s |
| unit3d | 0 | 0 | +0 | +0.00 | +0.49s |
| vito | 0 | 0 | +0 | +0.00 | +0.58s |
| laravel-excel | 0 | 0 | +0 | +0.00 | +0.38s |
| **filament** | **2** | **0** | **+2** | **-0.00** | **+0.54s** |
| laravel-backup | 0 | 0 | +0 | +0.00 | +0.60s |
| laravel-permission | 0 | 0 | +0 | +0.00 | -0.17s |
| laravel-socialite | 0 | 0 | +0 | +0.00 | +0.17s |
| **Total / Weighted** | **2** | **0** | **+2** | **-0.00** | **+1.85s** |

_ΔCov weighted by base `wall_seconds`._

### Issue type movements

|---|---:|---:|---:|
| MixedArgument | 1 | 0 | +1 |
| MixedArgumentTypeCoercion | 1 | 0 | +1 |

### Detail on filament regression

Both new issues land in `filament/4.x/packages/tables/src/Concerns/HasBulkActions.php:120-121`:

```
MixedArgumentTypeCoercion at L120
  Argument 1 of Illuminate\Support\Collection::reduce expects
  impure-callable(array<array-key, mixed>, mixed, array-key): array<array-key, mixed>,
  but parent type impure-Closure(array<...>, ...) provided

MixedArgument at L121
  Argument 1 of Filament\Tables\Table::isRecordSelectable cannot be mixed,
  expecting Illuminate\Database\Eloquent\Model|array<string, mixed>
```

Source:

```php
return $records->reduce(
    function (array $carry, Model | array $record, string $key): array {
        if (! $this->getTable()->isRecordSelectable($record)) {  // L121
```

This is filament's own code passing `mixed` into a typed `Model | array` parameter inside a `Collection::reduce` callback. PR #994's stricter `TClosure`-based macro typing is what surfaces it — the previous weaker shape masked the coercion. The issue is a real type imprecision in filament-side code, not a regression in plugin behaviour.

### Read

- **Net analysis impact**: precision-improving, not noise-introducing. The two new issues describe a genuine `mixed` leak into a typed callable parameter.
- **Wall-time impact**: +1.85s aggregate across 17 apps (~0.1s/app). Within run-to-run noise.
- **Coverage**: unchanged (rounded to 0.00 weighted).
- **Real-world body-inference firing rate**: separate grep over the same 17 apps shows zero `Foo::macro('name', fn() => 'literal')`-style registrations in app code (42 macro registrations total, all multi-statement and/or explicitly typed). The body-inference rule table is statistically rare in mature Laravel codebases; the real win lands when targeting Inertia-style packages or future Phase-2 rules (match/ternary).

Safe to merge.

PR #996, opened 2026-05-22
Recover closure return type from body when docblock and reflection are silent

alies-dev:
## Issue to Solve

Closures whose return type is implied by their body — `fn () => 'hello'`, `fn () => 1 + 2`, `function () { if (...) return 1; return 'x'; }` — currently surface as `mixed` from the macro recovery path when no docblock `@return` and no native return type are present. Reflection cannot peek inside the body; the docblock-based recovery shipped in #991 has nothing to read. PHPStan's `ClosureTypeFactory` infers from the body transparently in this case; psalm-plugin-laravel did not.

## Related

Continues #994 (base branch `worktree-991-macros-ast-scan-vendor`). Closes the "body-flow inference" follow-up gap noted in the #991 file-header docblock and #994 description.

## Solution Description

Extends `src/Util/Ast/ClosureTypeFactory.php` with a narrow body-flow inference path that fires only when both the docblock `@return` AND the native reflection return are absent. Anything more sophisticated (variable type-flow, generic template inference, `StatementsAnalyzer` re-entry) is explicitly out of scope; the rule table is small and conservative.

**Rule table** (anything outside it bails the entire inference, deliberately):

| AST node                                           | Inferred                                  |
|----------------------------------------------------|-------------------------------------------|
| `Scalar\String_`                                   | `TLiteralString` via `TLiteralString::make` |
| `Scalar\Int_`                                      | `Type::getInt(false, $value)`             |
| `Scalar\Float_`                                    | `Type::getFloat($value)`                  |
| `ConstFetch true / false / null`                   | matching constant `Union`                 |
| `Array_` (keys + values inferable)                 | shaped array via `TKeyedArray::make`      |
| `BinaryOp\Concat` of literal strings               | folded literal-string                     |
| `Plus/Minus/Mul/Div/Mod/Pow` of inferable numerics | `int\|float`                              |
| anything else                                      | bail (caller falls back to `mixed`)       |

**Multi-return closures:** the new `BodyReturnCollectorVisitor` walks the body collecting top-level returns while honouring `NodeVisitor::DONT_TRAVERSE_CHILDREN` for nested closures / arrow functions / function declarations / class methods (their returns belong to the inner function). The per-return unions are then `Type::combineUnionTypeArray`-folded.

**Implicit-`null` fall-through guard:** a conservative `bodyAlwaysTerminates()` check requires the last top-level statement to be a `return` or `throw`. Closures like `function () { if (cond) return 1; }` — which would silently return `null` at runtime — bail rather than produce a partial type that disagrees with PHP semantics.

**Memory hygiene:** `CachedClosureTypeFactory::reset()` now runs at the end of `MacroRegistry::init()`. The visitor's per-file index retains `Closure` / `ArrowFunction` AST subtrees so body inference can run without re-parsing; dropping the cache before Psalm's `pcntl_fork` phase avoids copy-on-write replication of those subtrees into every analysis worker.

## Verified by

- Type tests: `tests/Type/tests/Macros/MacroAstBodyInferenceTest.phpt` exercises literal-string, multi-return-union, concat-folding, and bail paths end-to-end through `MacroFixtureBag`.
- Unit tests: `tests/Unit/Util/Ast/ClosureTypeFactoryTest.php` adds one test per inference rule plus regression coverage for the implicit-`null` fall-through bail, nested-closure shielding, throw-only termination, try/catch bail, chained concat, mixed-key arrays, and a data-provided sweep over all six arithmetic operators.
- Cached layer: `tests/Unit/Util/Ast/CachedClosureTypeFactoryTest.php` adds a body-flow round-trip to verify the memoizing wrapper does not block inference results.
- Self-analysis: `composer psalm` exits 0, no new `psalm-baseline.xml` entries (none exists).

## Checklist

- [x] Documentation is updated (class-level docblocks in `ClosureTypeFactory` and `BodyReturnCollectorVisitor`)

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -21,6 +21,13 @@ return RectorConfig::configure()
     // that never matches the lowercase method id. (::class IS valid in const arrays on
     // PHP 8.3+; the problem is the casing, not the const-context.)
     ->withSkipPath('src/Handlers/Rules/OctaneIncompatibleBindingHandler.php')
+    // The vendor-style macro fixture (PR #991 + PR #994) deliberately exercises
+    // closures without native return types or docblock `@return` so the
+    // AST-scan + body-inference paths are the only sources of narrowing.
+    // Rector's `typeDeclarations` set keeps "fixing" them back to declared
+    // return types, which silently turns the body-inference assertions into
+    // native-return-type assertions and the feature stops being tested.
+    ->withSkipPath('tests/Type/macro-fixtures-vendor-style.php')
     ->withPhpVersion(PhpVersion::PHP_82)
     ->withSets([PHPUnitSetList::PHPUNIT_120])
     ->withPreparedSets(deadCode: true, codingStyle: true, typeDeclarations: true, codeQuality: true, phpunitCodeQuality: true)
--- a/src/Providers/MacroRegistry.php
+++ b/src/Providers/MacroRegistry.php
@@ -188,6 +188,16 @@ final class MacroRegistry
             self::$macros[\strtolower($className)] = $classMacros;
             self::$knownMacroableClasses[] = $className;
         }
+
+        // Drop the AST-scan cache now that every macro definition is built.
+        // The cache holds full `Closure` / `ArrowFunction` PhpParser subtrees
+        // per visited file; once we're past the foreach above, none of those
+        // nodes are reachable from `MacroDefinition` instances anymore. Holding
+        // them until session teardown would (a) keep the AST pinned for the
+        // rest of analysis and (b) replicate it into every pcntl fork worker
+        // via copy-on-write, which is exactly the cost we wanted to avoid by
+        // not re-parsing in the first place.
+        CachedClosureTypeFactory::reset();
     }
 
     /**
--- a/src/Util/Ast/BodyReturnCollectorVisitor.php
+++ b/src/Util/Ast/BodyReturnCollectorVisitor.php
@@ -88,7 +88,7 @@ final class BodyReturnCollectorVisitor extends NodeVisitorAbstract
         }
 
         if ($node instanceof Node\Stmt\Return_) {
-            if ($node->expr === null) {
+            if (!$node->expr instanceof \PhpParser\Node\Expr) {
                 $this->bailed = true;
                 return NodeVisitor::STOP_TRAVERSAL;
             }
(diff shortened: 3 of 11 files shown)


alies-dev:
## PR delta benchmark — 17 public Laravel apps

Base `0fedce24` → Head `bb75c96f`.

<details>
<summary>Per-app delta (click to expand)</summary>

| App                | +Add | -Rem | ΔNet | ΔCov  | ΔTime  |
|--------------------|-----:|-----:|-----:|------:|-------:|
| aureuserp          |    0 |    0 |   +0 | +0.00 | -0.99s |
| bagisto            |    0 |    0 |   +0 | +0.00 | -0.79s |
| bookstack          |    0 |    0 |   +0 | +0.00 | -0.92s |
| coolify            |    0 |    0 |   +0 | -0.00 | -1.29s |
| corcel             |    0 |    0 |   +0 | +0.00 | -0.41s |
| monica             |    0 |    0 |   +0 | +0.00 | -1.19s |
| pixelfed           |    0 |    0 |   +0 | -0.00 | -1.15s |
| solidtime          |    0 |    0 |   +0 | +0.00 | -0.82s |
| spatie-dashboard   |    0 |    0 |   +0 | +0.00 | -0.61s |
| tastyigniter       |    0 |    0 |   +0 | +0.00 | -0.60s |
| unit3d             |    0 |    0 |   +0 | +0.00 | -0.10s |
| vito               |    0 |    0 |   +0 | +0.00 | -0.08s |
| laravel-excel      |    0 |    0 |   +0 | +0.00 | -0.18s |
| filament           |    0 |    0 |   +0 | +0.00 | -0.19s |
| laravel-backup     |    0 |    0 |   +0 | +0.00 | -0.24s |
| laravel-permission |    0 |    0 |   +0 | +0.00 | -0.21s |
| laravel-socialite  |    0 |    0 |   +0 | +0.00 | -0.31s |
| **Total / Weighted** | **0** | **0** | **+0** | **-0.00** | **-10.08s** |

</details>

**Issue movements:** none.
**Coverage:** unchanged across all 17 apps.
**Wall time:** -10s aggregate (within noise; the `CachedClosureTypeFactory::reset()` call at end of `MacroRegistry::init()` may have helped marginally).

**Read:** zero net issue change means the body-flow inference path didn't fire for any macro in any of these codebases — none register closures that lack BOTH a docblock `@return` AND a native return type while having a return expression in the rule table. Expected — the PR targets the Inertia / vendor-package shape (`fn () => 'x'` without annotations) which is rare in established mature codebases that have already typed their macros.

No regressions. Safe to merge.

PR #1002, opened 2026-05-22
Resolve `Builder::find($mixed)` widening to `Collection|TModel|null`

### LINKED ISSUE #975: Stub: Builder::find($mixed) widens to Collection|TModel|null
**Reproducer:** \`tests/Type/tests/Builder/FindMixedNarrowingTest.phpt\` (#971).

The current stub uses a conditional:

\`\`\`php
// stubs/common/Database/Eloquent/Builder.phpstub
@psalm-return (T is (array|Arrayable) ? Collection<int, TModel> : TModel|null)
\`\`\`

When \`T = mixed\` (e.g. \`\\DB::select(...)\` returns \`list<stdClass>\` and \`\$row->id\` is \`mixed\`), Psalm cannot prove the negative branch and widens to \`Collection<int, TModel>|TModel|null\`. Property assignment downstream surfaces as \`UndefinedPropertyAssignment ... Collection::\$<col>\`.

**Source:** invoiceninja \`app/Console/Commands/CheckData.php:676/725/825\`:

\`\`\`php
\$client = Client::withTrashed()->find(\$_client->client_id);
\$client->paid_to_date = \$total_paid_to_date;  // fails on Collection branch
\`\`\`

**Fix vector (Larastan approach):** ship a \`@method static\` overload pair on the stub — one for scalar (\`T of int|string\` → \`TModel|null\`) and one for array/Arrayable (→ \`Collection<int, TModel>\`). This lets Psalm pick the correct branch even when the caller's \`\$id\` is \`mixed\` and Psalm keeps the union of both *signatures* rather than the union of both *return branches of a single signature*.

Flip the array/scalar positive controls into the main path and drop the EXPECTF lines on the reproducer once fixed.

alies-dev:
## Issue to Solve

`Builder::find/findOrFail/findOrNew/findOr` widen to `Collection<int, TModel>|TModel|null` when the argument is `mixed`. Psalm's `TemplateInferredTypeReplacer` cannot place `mixed` in either branch of the stub's `T is (array|Arrayable)` conditional, so it combines both. Downstream `$model->prop = ...` then surfaces as `UndefinedPropertyAssignment` on the spurious Collection branch.

The motivating real-world case is invoiceninja's `Client::withTrashed()->find($_client->client_id)` where `$_client` is a `\stdClass` from `\DB::select()` — every property read is `mixed` even though the runtime value is scalar.

## Related

Closes #975

## Solution Description

Larastan ships an `@method` overload pair on the stub. Psalm stores pseudo-methods last-write-wins in `$storage->pseudo_methods`, so the overload approach is not portable. New `BuilderFindMixedHandler` (`MethodReturnTypeProviderInterface`) collapses the mixed case to the scalar-id branch via the same trade-off:

- `find` → `TModel|null`
- `findOrFail` / `findOrNew` → `TModel`
- `findOr` → `TModel|TValue` (TValue from the callback closure at arg 1 or 2)

Concrete scalar/array arguments still narrow through the stub conditional.

Covers `Builder` and the four relation classes that re-declare the methods. `BelongsToMany`'s `&object{pivot: TPivotModel}` intersection is reconstructed from template index 2; other relations have no pivot.

The accepted soundness gap (mixed-truly-array → wrong narrowing) is pinned in `FindMixedTradeOffTest.phpt` so future readers see it.

Verification: 423/423 type tests, 632 unit, app exit 0, psalm self-check clean, cs clean.

## Checklist

- [x] Tests cover the change (4 new `.phpt` files under `tests/Type/tests/Builder/`)
- [x] Documentation: handler docblock + test header docs describe the trade-off

### DIFF
--- a/src/Handlers/Eloquent/BuilderFindMixedHandler.php
+++ b/src/Handlers/Eloquent/BuilderFindMixedHandler.php
@@ -1,0 +1,257 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Eloquent;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\HasManyThrough;
+use Illuminate\Database\Eloquent\Relations\HasOneOrMany;
+use Illuminate\Database\Eloquent\Relations\HasOneOrManyThrough;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\MethodCall;
+use Psalm\NodeTypeProvider;
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
+use Psalm\Type;
+use Psalm\Type\Atomic\TCallableObject;
+use Psalm\Type\Atomic\TClosure;
+use Psalm\Type\Atomic\TIterable;
+use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Atomic\TObjectWithProperties;
+use Psalm\Type\Atomic\TTemplateParam;
+use Psalm\Type\Union;
+
+/**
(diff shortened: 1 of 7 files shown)


alies-dev:
IMO that's ok, won't fix

Issue #854, opened 2026-04-26
Crash: Cannot get method params for AuthManager::authenticate (regression in 4.9.0)


alies-dev:
## Summary

After upgrading the plugin from 4.8.0 to 4.9.0, Psalm crashes during analysis with `Cannot get method params for Illuminate\Auth\AuthManager::authenticate` whenever a call is made to a method that `AuthManager` forwards via `__call` (e.g. `authenticate()`, `logout()`, `attempt()`, ...).

## Repro

`composer.json` (anything with the plugin v4.9.0 + Laravel 12/13 will do):

```bash
composer require --dev psalm/plugin-laravel:^4.9 vimeo/psalm:7.0.0-beta19
```

`psalm.xml`:

```xml
<?xml version="1.0"?>
<psalm errorLevel="3" findUnusedCode="false"
       xmlns="https://getpsalm.org/schema/config">
    <projectFiles><directory name="app" /></projectFiles>
    <plugins><pluginClass class="Psalm\LaravelPlugin\Plugin"/></plugins>
</psalm>
```

`app/Demo.php`:

```php
<?php

declare(strict_types=1);

namespace App;

use Illuminate\Auth\AuthManager;

final class Demo
{
    public function __construct(private AuthManager $auth) {}

    public function run(): void
    {
        $this->auth->authenticate();
    }
}
```

Run:

```bash
vendor/bin/psalm --no-cache
```

Result: `UnexpectedValueException: Cannot get method params for Illuminate\Auth\AuthManager::authenticate` and `Psalm 7.0.0-beta19 crashed due to an uncaught Throwable`.

The same crash reproduces in BookStack at `tests/Theme/LogicalThemeEventsTest.php` and other call sites that hit `auth()->authenticate()` / `Auth::logout()` / etc.

## Bisect

Regressed between v4.8.0 and v4.9.0. The trigger is PR #773 / commit dd93a61e (`fix: narrow AuthManager instance calls, not just the Auth facade`), which added `Illuminate\Auth\AuthManager` and `Illuminate\Contracts\Auth\Factory` to `AuthHandler::getClassLikeNames()`. Reverting that registration makes the crash go away.

[...]

Issue #454, opened 2026-03-16
Auth facade methods crash in Psalm 7: Cannot get method params for Auth::getUser


alies-dev:
## Bug

When running psalm-plugin-laravel 4.x (dev-master) against projects that call `Auth::getUser()` or other `@method`-annotated Auth facade methods, Psalm 7 crashes with:

```
Uncaught UnexpectedValueException: Cannot get method params for Illuminate\Support\Facades\Auth::getuser
in vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Methods.php:484
```

## Root cause

`AuthHandler::getMethodParams()` only handles the `user` method and returns `null` for all others. In Psalm 7, returning `null` from a `MethodParamsProviderInterface` for methods that only exist as `@method` annotations causes a crash — Psalm 6 handled this gracefully.

`Auth::getUser()` appears 146 times across 79 files in IXP-Manager, making this a blocking issue for real-world adoption.

## Fix

PR #453 — return explicit `FunctionLikeParameter` arrays for all seven handled Auth facade methods.

## Affected methods

- `Auth::getUser()`
- `Auth::authenticate()`
- `Auth::getLastAttempted()`
- `Auth::loginUsingId()`
- `Auth::onceUsingId()`
- `Auth::logoutOtherDevices()`

## Discovered in

Tested against top open-source dependents of psalm/plugin-laravel:
- **IXP-Manager** (421⭐) — crashed
- **laravel-in-app-purchases** (382⭐) — passed cleanly

PR #1006, opened 2026-05-23
Narrow `config()` and `Repository::get()` return types

### LINKED ISSUE #752: Add `FunctionReturnTypeProvider` for `config()` — resolve values via booted Application
## Description

`config('some.key')` currently returns `mixed`. The plugin has all the infrastructure to resolve these calls at analysis time (the Laravel app is already booted via Testbench) but no handler consumes it for the `config()` helper.

This shows up in real-world audits as a meaningful chunk of `MixedAssignment` / `MixedArgument` errors. In filamentphp/filament at `errorLevel=1`, 21 `MixedAssignment` come directly from `config('filament.*')` calls.

## Current state

- Stub `stubs/common/Foundation/helpers.stubphp:102-111` returns `mixed` for string keys:
  ```
  @return ($key is null ? Repository : ($key is array ? null : mixed))
  ```
- `src/Providers/ConfigRepositoryProvider.php` wraps the booted `Illuminate\Contracts\Config\Repository`, but is only consumed by `AuthConfigAnalyzer` (for `auth.guards.*.provider.model`).
- No `FunctionReturnTypeProviderInterface` handler for `config`.

## What Larastan does (for reference)

Larastan registers `ConfigFunctionDynamicFunctionReturnTypeExtension` (see `larastan/larastan/src/ReturnTypes/Helpers/`) which:
- Requires opt-in via `configPaths: [config]` in `phpstan.neon`.
- Walks `config/*.php` files **statically** with php-parser, extracts each returned array shape, and maps dot-paths to the inferred type of that subtree.
- Unions with the `\$default` argument's generalized type when present.

Never boots the Laravel app.

## Proposal

Add a `ConfigHelperHandler` under `src/Handlers/Helpers/` implementing `FunctionReturnTypeProviderInterface` with `getFunctionIds(): ['config']`.

Resolution strategy (runtime, piggy-backing on the booted app):

1. If no args → `TNamedObject(Illuminate\\Config\\Repository)` (already handled by the stub).
2. If first arg is an array literal → `null` (setter form, stub already correct).
3. If first arg is a single-literal string `\$key`:
   - Call `ConfigRepositoryProvider::get()->get(\$key)`.
   - Reflect the returned value into a Psalm `Union`:
     - scalars → `TLiteralString` / `TLiteralInt` / `TLiteralFloat` / `TTrue`/`TFalse`.
     - arrays → recurse into `TKeyedArray` where feasible, else `TArray<array-key, mixed>`.
     - objects → `TNamedObject(\$class)`.
     - `null` → consider the `\$default` arg (if present) or `TNull`.
   - If `get()` throws → fall back to `mixed` (do not emit an internal error).
4. If the default value (second arg) is present and typed, union it with the resolved type.

Cache: lookups are pure once the app is booted, so a simple `static array<string, Union> \$cache` mirrors `ContainerResolver::\$cache`.

## Advantages over Larastan's approach

- **Zero user configuration** — no `configPaths` setting; the booted app already knows its config paths (including paths from registered service providers).
- **Accurate for package config merged via `\$this->mergeConfigFrom(...)`** — Larastan's static parser can't follow provider-registered merges; runtime resolution sees the merged state.
- **Shape-aware** — can emit `TKeyedArray` from the actual array structure, not just the top-level PHPDoc.

## Caveats

- Values from `.env` / user overrides won't be visible (Testbench boot uses defaults). Same limitation as Larastan's static approach, just from the other direction.
- Catch `\Throwable` inside `ConfigRepositoryProvider::get()->get()` — avoid letting a package's broken config closure kill analysis.
- Keep the stub's `@return (...)` for the no-args / array-arg branches so the handler only needs to cover the literal-string case.

## Reference

Larastan equivalents:
- `larastan/larastan/src/ReturnTypes/Helpers/ConfigFunctionDynamicFunctionReturnTypeExtension.php`
- `larastan/larastan/src/Internal/ConfigHelper.php`
- `larastan/larastan/src/Support/ConfigParser.php`

Spotted during audit of filamentphp/filament (see #750 for the related `app(static::class)` gap from the same audit).

alies-dev:
## Issue to Solve

`config('some.key')`, `\Illuminate\Config\Repository::get('some.key')`, and `Config::get('some.key')` all return `mixed`. Every callsite then triggers `MixedAssignment`/`MixedArgument`/`MixedOperand` issues, which is the headline of #752 and a major source of noise in real-world Laravel apps (Filament, Bagisto, etc.).

## Related

Closes #752.

## Solution Description

A Psalm type provider (`ConfigHelperHandler` + `ConfigRepositoryMethodHandler`) routes the three call shapes through a singleton `ConfigKeyResolver`. The resolver reads the value from the booted Laravel app via the existing `ConfigRepositoryProvider`, reflects it into a Psalm `Union` (`ConfigValueReflector`), and merges that with the default-arg type to produce a narrowed return.

**Why scalars stay general.** Booted Laravel resolves env-driven defaults to a single concrete observation at analysis time. Narrowing `config('app.debug')` to `false` (its boot-time value) would surface spurious `TypeDoesNotContainType` errors on every `if (config('app.debug'))` callsite; narrowing `config('app.name')` to `'Laravel'` would break every `===` against a production override. The reflector therefore generalizes literals: `bool`/`int`/`string`/`float` rather than `TLiteralBool`/`TLiteralInt`/etc. Arrays preserve their shape up to depth 5 with a 64-key cap per level; beyond that the type degrades to `array<array-key, mixed>` to avoid pathological recursion (cyclic refs, megabyte-scale type strings).

**Default-arg merge mirrors `Arr::get` runtime semantics.**

| key state                 | default arg | returned type                       |
|---------------------------|-------------|--------------------------------------|
| absent                    | any         | generalized default                  |
| present, value not null   | any         | reflected value (default ignored)    |
| present, value is null    | any         | `null \| generalized default`        |

Closure defaults follow through to their declared return type via recursive generalization. Untyped closures contribute `mixed` without discarding atomics already collected from other union members (so `string | Closure(): void` resolves to `string | mixed`, not just `mixed`).

**Literal AST inspection.** Psalm's `NodeTypeProvider` returns `null` for argument nodes inside the facade `__callStatic` dispatch — `Config::get('app.name')` would otherwise drop to `mixed` because the second arg's type is unknown at handler time. Both `extractLiteralKey` and `readDefaultTypeAt` inspect the raw AST (`String_`, `Int_`, `Float_`, `ConstFetch`, `Array_`) first and fall back to the type provider only when the expression is not a literal.

**`Config` facade routing.** Listing `\Illuminate\Support\Facades\Config::class` (plus `FacadeMapProvider::getFacadeClasses(\Illuminate\Config\Repository::class)`) in `getClassLikeNames` makes the handler fire for the facade. `MethodParamsProviderInterface` is implemented alongside the return-type hook because Psalm 7 throws `Cannot get method params for Illuminate\Support\Facades\Config::get` when the registered method only exists as a `@method` annotation (same crash class as #454 / #854). The provider synthesises params for the facade FQCN and defers to source for the real `Repository`/contract.

**Failure-safe resolver bootstrap.** `ConfigKeyResolver::instance()` wraps `ConfigRepositoryProvider::get()` in a try/catch and falls back to a `ThrowingConfigRepository` sentinel. Without that, an unbound binding (or an exploding service provider during plugin init) would propagate from a Psalm hook handler and crash analysis on every `config()` call site. The sentinel's methods all throw, so `ConfigKeyResolver::warm()`'s existing catch arm caches `mixed` (matching the pre-PR stub ceiling).

**Out of scope (follow-up PR).** Reading `@return` PHPDoc on `config/*.php` files (Larastan PR #2445 equivalent). Different source of truth (static AST vs runtime), different fixtures, different review surface — shipped separately to keep this PR reviewable.

## Verification

- `composer test:type` (423 tests) — type tests cover all narrowing branches across `config()`, `Repository::get`, and `Config::get`.
- `composer test:unit` (707 tests) — unit tests for the reflector (granularity table, depth/key caps) and resolver (cache trichotomy, generalization, closure handling, throwing repo).
- `composer psalm` — clean.
- `composer test:app` — clean.

## Checklist

- [x] Tests cover the change (type tests in `tests/Type/` + unit tests in `tests/Unit/`)

### DIFF
--- a/src/Handlers/Config/ConfigRepositoryMethodHandler.php
+++ b/src/Handlers/Config/ConfigRepositoryMethodHandler.php
@@ -14,28 +14,21 @@ use Psalm\Storage\FunctionLikeParameter;
 use Psalm\Type;
 
 /**
- * Narrows `Repository::get('some.key', $default)` (and the equivalent call on the
- * `\Illuminate\Contracts\Config\Repository` contract) to the runtime type of
- * `some.key`. Cross-receiver consistency with `Config::get(...)` is handled by
- * Psalm's facade pipeline — the facade resolves to the concrete Repository and
- * triggers this handler.
+ * Narrows `Repository::get()` (concrete + contract) and `Config::get()` (facade)
+ * to the runtime type reflected from the booted Laravel app.
  *
- * Other typed accessors (`string`, `integer`, `float`, `boolean`, `array`,
- * `collection`) are intentionally not handled here: their stub return types in
- * `stubs/common/Config/Repository.phpstub` are already precise (the stub even
- * tightens `$default` to reject scalars that would crash at runtime).
+ * Typed accessors (`string`, `integer`, `float`, `boolean`, `array`,
+ * `collection`) are not handled here — their stub return types in
+ * `stubs/common/Config/Repository.phpstub` are already precise.
  *
- * Solution 3 of https://github.com/psalm/psalm-plugin-laravel/issues/752.
+ * See https://github.com/psalm/psalm-plugin-laravel/issues/752.
  */
 final class ConfigRepositoryMethodHandler implements MethodReturnTypeProviderInterface, MethodParamsProviderInterface
 {
     /**
-     * Register for the concrete Repository, the contract, the Config facade,
-     * and any root-namespace alias resolving to the Repository binding.
-     * Psalm dispatches MethodReturnTypeProvider hooks by exact class name, so
-     * `Config::get(...)` (which uses `__callStatic` on the facade) needs the
-     * facade FQCN listed here — otherwise the facade pipeline falls through
-     * to the stub and the call site stays at `mixed`.
+     * Psalm dispatches `MethodReturnTypeProvider` by exact class name, so the
+     * facade FQCN (and any root-namespace alias) must be registered alongside
+     * the concrete/contract — otherwise `Config::get()` falls back to the stub.
      *
      * @inheritDoc
      * @return list<string>
@@ -60,10 +53,6 @@ final class ConfigRepositoryMethodHandler implements MethodReturnTypeProviderInt
             return null;
         }
 
-        // resolveFromCallArgs returns null for shapes the stub handles:
-        //   - 0 args
-        //   - array first arg (multi-key getMany form)
-        //   - dynamic first arg → mixed
         return ConfigKeyResolver::resolveFromCallArgs(
             $event->getCallArgs(),
             $event->getSource()->getNodeTypeProvider(),
@@ -71,19 +60,11 @@ final class ConfigRepositoryMethodHandler implements MethodReturnTypeProviderInt
     }
 
     /**
-     * Provide explicit `get()` params for the Config facade FQCN and any
-     * root-namespace alias that proxies to it.
-     *
-     * In Psalm 7, registering a class for a MethodReturnTypeProviderInterface
-     * and then returning null from getMethodParams for a method that only
-     * exists as a `@method` annotation on that class causes an
-     * UnexpectedValueException: Cannot get method params for ...::get crash.
-     * The same crash hit AuthMethodHandler — see #454 and #854.
-     *
-     * For the concrete Repository and the contract, defer to Psalm by returning
-     * null: get() is a real method on both, and letting Psalm derive params
-     * from the source keeps the signature in sync with future Laravel
-     * versions automatically.
+     * Synthesise `get()` params for the Facade FQCN (`@method`-declared only).
+     * Without this, Psalm 7 crashes with
+     * `Cannot get method params for ...::get` — same crash class as #454/#854.
+     * Defers to source for the real Repository + contract so future Laravel
+     * signature changes are picked up automatically.
      *
      * @inheritDoc
      * @psalm-mutation-free
@@ -103,6 +84,7 @@ final class ConfigRepositoryMethodHandler implements MethodReturnTypeProviderInt
             return null;
         }
 
+        // `@method static mixed get(array|string $key, mixed $default = null)`
         $stringOrArray = new Type\Union([
             new Type\Atomic\TString(),
             new Type\Atomic\TArray([Type::getArrayKey(), Type::getMixed()]),
--- a/src/Handlers/Config/ConfigRepositoryMethodHandler.php
+++ b/src/Handlers/Config/ConfigRepositoryMethodHandler.php
@@ -1,0 +1,123 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Config;
+
+use Psalm\LaravelPlugin\Providers\FacadeMapProvider;
+use Psalm\LaravelPlugin\Util\ConfigKeyResolver;
+use Psalm\Plugin\EventHandler\Event\MethodParamsProviderEvent;
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\MethodParamsProviderInterface;
+use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
+use Psalm\Storage\FunctionLikeParameter;
+use Psalm\Type;
+
+/**
+ * Narrows `Repository::get('some.key', $default)` (and the equivalent call on the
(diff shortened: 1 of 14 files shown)


alies-dev:
## Real-world apps delta (17 apps, base `2b598066` → head `0cac1ed4`)

### Per-app delta

| aureuserp        |   17 |   72 |   -57 |  +7.89 | +0.18s |
| bagisto          |    1 |    1 |    +0 |  +0.19 | +0.56s |
| bookstack        |   25 |   79 |   -56 |  +0.22 | +0.40s |
| coolify          |   46 |  222 |  -178 |  +0.17 | +0.42s |
| corcel           |    8 |   11 |    -3 |  +0.39 | +0.14s |
| monica           |   28 |   81 |   -53 |  +0.19 | +0.14s |
| pixelfed         |  159 |  684 |  -544 |  +0.46 | +0.05s |
| solidtime        |    2 |   64 |   -63 |  +0.34 | -0.44s |
| spatie-dashboard |   10 |   13 |    -3 |  +1.30 | -0.12s |
| tastyigniter     |    0 |    0 |    +0 |  +0.00 | +0.29s |
| unit3d           |    0 |    0 |    +0 |  +0.00 | -0.18s |
| vito             |   11 |  131 |  -120 |  +0.54 | -0.01s |
| laravel-excel    |    7 |   35 |   -28 |  +0.62 | -0.02s |
| filament         |   33 |   61 |   -28 |  +0.06 | -0.01s |
| laravel-backup   |   10 |   12 |    -2 |  +0.30 | -0.35s |
| laravel-permission |   97 |  102 |    -3 |  +0.42 | -0.49s |
| laravel-socialite |   14 |   28 |   -14 |  +2.26 | -0.32s |
| **Total / Weighted** | **468** | **1596** | **-1152** | **+0.87** | **+0.24s** |

_ΔCov is weighted by base `wall_seconds` (proxy for codebase size)._

### Issue type movements (across all apps)

| MixedArgument | 20 | 619 | -599 |
| MixedAssignment | 0 | 378 | -378 |
| MixedOperand | 4 | 183 | -179 |
| NullArgument | 98 | 0 | +98 |
| MixedArrayAccess | 1 | 95 | -94 |
| MixedReturnStatement | 0 | 76 | -76 |
| TypeDoesNotContainType | 57 | 2 | +55 |
| RedundantCondition | 52 | 2 | +50 |
| PossiblyNullPropertyFetch | 0 | 41 | -41 |
| MissingReturnType | 18 | 18 | +0 |
| RedundantCast | 33 | 1 | +32 |
| NullArrayAccess | 29 | 0 | +29 |
| NullOperand | 27 | 0 | +27 |
| MixedMethodCall | 0 | 22 | -22 |
| InvalidOperand | 4 | 17 | -13 |
| MixedArrayAssignment | 0 | 21 | -21 |
| InvalidReturnType | 15 | 2 | +13 |
| NoValue | 17 | 0 | +17 |
| PossiblyNullOperand | 0 | 17 | -17 |
| TypeDoesNotContainNull | 15 | 1 | +14 |
| PossiblyNullArgument | 9 | 3 | +6 |
| InvalidReturnStatement | 7 | 4 | +3 |
| RiskyTruthyFalsyComparison | 2 | 9 | -7 |
| MixedArgumentTypeCoercion | 2 | 8 | -6 |
| MixedArrayOffset | 0 | 10 | -10 |
| InvalidNullableReturnType | 9 | 0 | +9 |
| NullableReturnStatement | 8 | 1 | +7 |
| PossiblyUndefinedVariable | 0 | 9 | -9 |
| MixedPropertyAssignment | 0 | 8 | -8 |
| MixedPropertyFetch | 0 | 8 | -8 |

_(24 more types changed; top 30 shown.)_

### Newly introduced issue types

- `NullArrayAccess` (31)
- `NullIterator` (4)

### Summary

Net: **-1152 issues** across the 17-app fleet, with no app showing a positive net delta. The added issues (`NullArgument`, `TypeDoesNotContainType`, `RedundantCondition`, `NullArrayAccess`, `NullOperand`) are legitimate findings surfaced by the narrowed types — callers that were previously hidden behind `mixed` now show their real null-flow bugs. Wall time is essentially flat (+0.24s total across 17 apps).

tm1000:
@alies-dev wonderful work on this project

alies-dev:
> @alies-dev wonderful work on this project

Thank you @tm1000 , such feedbacks inspire even more! 💪

PR #1013, opened 2026-05-23
Detect timing-unsafe comparisons of secrets (CWE-208)

### LINKED ISSUE #570: Detect timing attack vulnerability: === instead of `hash_equals()` for secrets (CWE-208)
## Problem

String comparison with `===`, `==`, or `strcmp()` on secrets is vulnerable to timing attacks. An attacker can determine the correct value character-by-character by measuring response time differences. PHP's `hash_equals()` provides constant-time comparison.

Laravel's own Auth code uses `hash_equals()` correctly, but application code often doesn't.

## Examples

```php
// Vulnerable — timing attack
if ($request->input('token') === $expectedToken) { ... }
if ($request->header('X-Webhook-Signature') == $computedSignature) { ... }
if (strcmp($apiKey, config('services.stripe.secret')) === 0) { ... }

// Secure
if (hash_equals($expectedToken, $request->input('token'))) { ... }
```

## Implementation

Custom Psalm handler that detects `===`/`==`/`strcmp`/`strcasecmp` comparisons where one operand carries `user_secret` or `system_secret` taint (from `$request->input()`, `$request->header()`, config values).

Emit a custom issue like `TimingUnsafeComparison` suggesting `hash_equals()`.

## Scope

Only flag comparisons involving tainted secrets, not all string comparisons. This keeps false positives low.

## References

- [CWE-208: Observable Timing Discrepancy](https://cwe.mitre.org/data/definitions/208.html)
- [PHP hash_equals](https://www.php.net/hash_equals)

alies-dev:
## Issue to Solve

Timing-attack vulnerabilities (CWE-208): user / framework code that compares a secret with `===`, `==`, `strcmp()`, etc. instead of `hash_equals()`. The variable-time comparison leaks the secret character-by-character to an attacker who can observe response timing.

Re-implements the closed PR #576 (auto-closed when the topic branch's history was rebuilt). Same scope, with several round-1/2/3 review-driven improvements.

## Related

Closes #570
Re-implements #576 (closed)
Blocked-by (issue message): vimeo/psalm#11762

## Solution Description

Adds `TimingUnsafeComparisonHandler` (`src/Handlers/Rules/TimingUnsafeComparisonHandler.php`). It implements `AfterExpressionAnalysisInterface` and, during `--taint-analysis` runs, registers `USER_SECRET | SYSTEM_SECRET` taint sinks at:

- Equality/inequality operators: `===`, `==`, `!==`, `!=`
- C-level byte-by-byte string-compare functions: `strcmp`, `strcasecmp`, `strncmp`, `strncasecmp`, `substr_compare`

When a value carrying `user_secret` (e.g. `Authenticatable::getAuthPassword()` / `getRememberToken()`) or `system_secret` (e.g. `Encrypter::decryptString()`) taint flows into one of these sinks, Psalm reports `TaintedUserSecret` / `TaintedSystemSecret`. The fix at the user site is to use `hash_equals()` for constant-time comparison.

**Literal-operand carve-out.** Comparisons against a constant the developer wrote directly are skipped, so idiomatic defensive checks (`if ($token === null)`, `if ($apiKey === '')`) don't false-positive. Accepted shapes: integer/float/string scalars, magic constants (`__FILE__`, `__LINE__`, ...), `null`/`true`/`false`, unary `+`/`-` over a literal, and concatenation of two literals (recursive). Class constants (`Foo::BAR`) and enum cases (`Status::Active`) are intentionally NOT exempted. An attacker-controlled indirection could resolve to either at runtime, so the rule errs on flagging a rare FP rather than silently disarming for an entire common pattern. The check matches by AST shape, not by Psalm's inferred type, so a `string $x` narrowed by Psalm to `''` after a prior check still flags (which is correct — the data flow is real).

**Runtime cost.** Per-expression hook, but the body exits immediately when `$source->taint_flow_graph` is not a `TaintFlowGraph` (i.e. when `--taint-analysis` is not enabled). On the non-taint path, the cost is one `instanceof StatementsAnalyzer` + one field read + one `instanceof TaintFlowGraph`. During taint runs, the relevant-expression filter (`instanceof BinaryOp` → 4-arm narrow chain; otherwise `FuncCall` with `count(args) >= 2` checked before `toLowerString()` so unrelated nullary/unary calls don't pay for the lowercase allocation) keeps the hot path tight.

**Upstream limitation (vimeo/psalm#11762).** Psalm 7 hardcodes the issue message per taint kind in `TaintFlowGraph::connectSinksAndSources()`, so the emitted text is the generic `"Detected tainted user secret leaking"` rather than a CWE-208-specific one. The data-flow trace itself still pinpoints the timing-unsafe comparison site, so the report is actionable today. Once the upstream API lands, the handler will switch to a CWE-tagged message.

**Known gaps.** Documented in [docs/contributing/taint-analysis.md](docs/contributing/taint-analysis.md) under the new "Plugin-emitted taint sinks (handler-driven)" section: `switch ($secret) { case $candidate: }`, `match ($secret) { ... }`, and partial-leak operations (`str_starts_with`, `str_ends_with`, `str_contains`, `preg_match` with attacker-controlled pattern, `in_array($secret, $list, false)`, `Str::of($secret)->is($candidate)`) are intentionally out of scope for this PR and tracked as follow-ups.

### Verification

- 432 type tests pass (`composer test:type`), including 8 new PHPT fixtures covering positive cases (`===` / `strcmp` / `strncmp` / `strcasecmp` on `user_secret` and `system_secret` sources), the multi-operator regression test for `==` / `!==` / `!=`, the literal-operand exemption (`=== null`, `=== ''`, `!== 'unset'`, `!= false`, `=== __FILE__`, `=== 'sentinel-' . 'marker'`), and the negative cases for `hash_equals()` and non-secret input.
- 719 unit tests pass (`composer test:unit`).
- 100% Psalm type coverage maintained (`composer psalm -- --stats`).
- `composer cs` and `composer rector` clean.

## Checklist
- [x] Tests cover the change (8 new PHPT fixtures in `tests/Type/tests/TaintAnalysis/`)
- [x] Documentation is updated (`docs/contributing/taint-analysis.md` — new "Plugin-emitted taint sinks" section)


alies-dev:
## psalm-delta — whole-PR impact on real-world apps

Base `ae8707bf` (fork point from `master`, rule absent) → head `f1b9ff2c` (rule present, incl. the relational-operator and named-argument fixes). 18 real Laravel apps, Psalm + plugin with taint analysis.

**Net effect: +4 new findings, all in filament. Zero regressions, zero churn elsewhere, type coverage flat.**

### Issue type movements

| TaintedUserSecret | 2 | 0 | +2 |
| TaintedSystemSecret | 2 | 0 | +2 |

All four are genuine timing-unsafe secret comparisons surfaced by the new rule. No previously-passing finding changed, and no other app moved.

### Per-app (non-zero rows only)

| App | +Add | -Rem | ΔCov | ΔTime |
|-----|-----:|-----:|-----:|------:|
| filament | 4 | 0 | -0.00 | +8.4s |
| _all 17 others_ | 0 | 0 | +0.00 | noise |

Wall-time deltas elsewhere (unit3d +6.7s, pixelfed +5.9s, private-project-1 +8.0s) are run-to-run variance, not attributable to the rule — those apps show 0 issue movement. The handler bails immediately when `taint_flow_graph` is absent, so non-taint runs pay ~zero.

Numbers cover the full rule (the PR as a whole). Happy to re-run an incremental delta (`43457285` → `f1b9ff2c`) to isolate just the relational/named-argument commit if useful.

Issue #1015, opened 2026-05-26
Type narrowing missing for `$this->input('field')` inside FormRequest


alies-dev:
## Summary

Plugin narrows `validated()` / `safe()->input()` calls on a `FormRequest` from outside (controllers, services). It does **not** narrow `$this->input('field')` (or sibling accessors) when called **inside** the same `FormRequest` subclass, e.g. inside `prepareForValidation()`, `passedValidation()`, `withValidator()`, or any custom helper. Such calls fall through to `Illuminate\Http\Concerns\InteractsWithInput::input()`'s stubbed `mixed` return.

This is a precision gap, not a soundness bug: `mixed` is correct, just not narrow.

## Reproduction

Pinned in `tests/Type/tests/FormRequestSelfInputTest.phpt`:

```php
class SelfEmailRequest extends FormRequest
{
    public function rules(): array
    {
        return [
            'email' => ['required', 'email'],
            'age'   => ['required', 'integer'],
            'role'  => ['required', 'in:admin,user,guest'],
        ];
    }

    public function prepareForValidation(): void
    {
        $_email = $this->input('email');
        /** @psalm-check-type-exact $_email = mixed */  // current

        $_age = $this->input('age');
        /** @psalm-check-type-exact $_age = mixed */    // current

        $_role = $this->input('role');
        /** @psalm-check-type-exact $_role = mixed */   // current
    }
}

function fromController(SelfEmailRequest $request): void
{
    $_email = $request->validated('email');
    /** @psalm-check-type-exact $_email = string */                       // works

    $_age = $request->validated('age');
    /** @psalm-check-type-exact $_age = int|numeric-string */             // works

    $_role = $request->validated('role');
    /** @psalm-check-type-exact $_role = 'admin'|'guest'|'user' */        // works
}
```

## Expected

`$this->input('field')` on a `FormRequest` subclass with a `rules()` array should narrow to the same type that `$request->validated('field')` already produces from outside.

Important nuance: `input()` reads pre-validation request data. The narrowed type from a `rules()` array is only sound after validation has run. Two valid options:

1. **Mirror `validated()` precisely.** Narrow `$this->input('field')` to the rule's type for fields whose rule is `required` (presence-guaranteed). Live with possible unsoundness for code paths that run before validation (`prepareForValidation`).
2. **Narrow to `rule_type|null` (or just keep `mixed` when the field is `sometimes|...`).** Reflects that the field may not be present yet.

[...]

PR #1022, opened 2026-05-31
Narrow `FormRequest` magic property reads (`$request->field` calls)

### LINKED ISSUE #1016: Type narrowing for `$this->email` property access inside FormRequest
## Summary

Follow-up to #1015. Plugin now narrows `$this->input('field')` inside FormRequest subclasses to the rule type when the field is presence-guaranteed (`required` / `present` / `accepted` / `declined`, sans `sometimes`). The analogous magic-property access form, `$this->email`, is still untyped: it falls through to `Request::__get()` and the static analyzer sees no defined property.

## Reproduction

```php
class StoreUserRequest extends FormRequest
{
    public function rules(): array
    {
        return ['email' => ['required', 'email']];
    }

    public function prepareForValidation(): void
    {
        $_input = $this->input('email');  // string  (post #1015)
        $_email = $this->email;           // currently: UndefinedThisPropertyFetch / mixed
    }
}
```

## Expected

`$this->email` narrows to the same type as `$this->input('email')` for presence-guaranteed fields, with `sometimes`/non-required falling through.

## Scope

- New `PropertyTypeProviderInterface` + `PropertyExistenceProviderInterface` handlers in `src/Handlers/Validation/` (no existing property handler for the Request hierarchy; Eloquent is the only precedent in `src/Handlers/Eloquent/`).
- Gating rules identical to `ValidatedTypeHandler::resolveSelfInput` (presence-guarantee + sane `sometimes` bail-out).
- Mirror the taint flow in `ValidationTaintHandler` if a property-source hook silences a stub source.

## Caveats

`Request::__get($key)` runtime reads from `$this->all() + $this->route()->parameters()`, i.e. input merged with route params. The route-param fallback widens the value source beyond what `input()` covers. Out of scope here, but the design must either acknowledge the fallback or restrict narrowing to fields that cannot collide with a route binding (the strict approach matches what `$this->input()` already does).

## Out of scope

- `$this->all()`, `$this->only([...])`, `$this->except([...])` (same as #1015).
- Narrowing across `$this->merge([...])` mutations (would need flow-sensitive analysis).
- `$this->array(...)` / `$this->collect(...)` — see the deferred coverage in #1015.

alies-dev:
## Issue to Solve

Inside a `FormRequest` subclass with `'email' => ['required', 'email']` in `rules()`, `$this->email` and `$req->email` reads currently fall through to `Request::__get()` and surface as `mixed` + `UndefinedThisPropertyFetch`. The matching `$this->input('email')` form was narrowed in #1015 but the magic-property shape was left out, blocking the natural ergonomic style for accessing validated fields.

## Related

Closes #1016. Follow-up to #1015 (method-call shape).

## Solution Description

Per-FormRequest-subclass property providers register at `AfterCodebasePopulated`, mirroring the `ModelRegistrationHandler` pattern. The resolver in `FormRequestPropertyHandler::resolveRuleForProperty` is shared by both the type narrowing and the paired taint paths in `ValidationTaintHandler`, so the same gates apply to both:

- declared property (`public ?int $email`) or `@property` / `@property-read` PHPDoc opts the field out (`hasDeclaredProperty` reads `declaring_property_ids` + `pseudo_property_get_types`, including inherited ones);
- only presence-guaranteed rules narrow (`required` / `present` / `accepted` / `declined`, sans `sometimes`);
- non-FormRequest callers and writes are skipped.

Soundness gate matches `ValidatedTypeHandler::resolveSelfInput`. Route-param fallback in `Request::__get` is deliberately ignored — same strictness as `input()`.

Taint specifics:

- `addTaints` re-sources `ALL_INPUT` for narrowed reads (no stub source exists on `__get`), de-duplicated per `PropertyFetch` to avoid double-emission from the dual dispatch in `AtomicPropertyFetchAnalyzer` + `ArgumentAnalyzer`.
- `removeTaints` applies `ResolvedRule::removedTaints` via the same resolver.
- Assignment-LHS `PropertyFetch` nodes are tracked by `spl_object_id` (populated in `beforeExpressionAnalysis`) and skipped in `resolvePropertyFetchRule`, preventing the rule's escape mask from being applied to writes.
- Marker sets flush per function-like and per file to bound the footprint and avoid `spl_object_id` reuse across files.

A `ResolvedRule::guaranteesPresence()` helper is extracted so `ValidatedTypeHandler::resolveSelfInput`, `FormRequestPropertyHandler`, and `ValidationTaintHandler` share one source of truth for the presence-gate predicate.

Verified by:

- `tests/Type/tests/FormRequestSelfPropertyTest.phpt` — rule narrowing, declared-property defer, `@property` defer, `@property-read` defer, inherited declared property, inherited `rules()`, external controller access.
- `tests/Type/tests/TaintAnalysis/SafeFormRequestSelfPropertyEmailEscapesHeader.phpt` — `email` rule's `INPUT_HEADER` escape applies; HTML sink still reports.
- `tests/Type/tests/TaintAnalysis/SafeFormRequestDeclaredPropertyNoPhantomTaint.phpt` — declared `public string $email` + matching rule produces zero taint reports (regression guard for the type/taint gate alignment).

## Checklist

- [x] Tests cover the change (type tests in `tests/Type/tests/`, taint tests in `tests/Type/tests/TaintAnalysis/`)


alies-dev:
Reworked to treat `$request->email` as another spelling of a validated field read rather than a separate subsystem. Pushed as `refactor(validation): unify validated-field reads behind one resolver`.

### What changed

- **One resolver owns the question.** New `ValidatedFieldReadResolver::resolve(AddRemoveTaintsEvent)` returns a small `ValidatedFieldRead { sourceTaints, removedTaints }` for every syntax: keyed accessor methods (`input('email')`), `ValidatedInput` accessors (`safe()->input('email')`), magic property fetches (`$req->email`), and tracked inline-validate variables. `addTaints`/`removeTaints` each call it once and read one facet. The parallel method-vs-property branches are gone.
- **`ValidationTaintHandler` is now pure taint-graph mechanism**, 785 to 206 lines: source re-emission, per-node dedupe, cleanup. No rule lookup, no caller-class resolution.
- **Removed the assignment-LHS tracking machinery** (`BeforeExpressionAnalysis` hook + `$assignmentLhsPropertyFetchIds` marker set). A declared property opts out of narrowing, and an undeclared magic property cannot be written (`Request` has no `__set`, so the write is already `UndefinedPropertyAssignment`). The guard defended an unreachable state. Verified the full taint suite is byte-identical with it removed, and added `UnsafeFormRequestPropertyWriteNoPhantomSource.phpt` pinning that the write produces no phantom source (one `UndefinedPropertyAssignment`, one `TaintedHtml` from the read, no duplicate).

### On the two suggestions I did *not* take

- **`__get` stub with `@psalm-taint-source input`.** Tested it: it does not fire for `$req->email`. The property type has to come from a provider (each field's type differs by rule, so a static stub can't express it), and a provider-resolved `PropertyFetch` means Psalm never synthesizes a `__get` call node for the source to attach to. Same shape as #11765 for the method path. So the `addTaints` re-source is required, not custom overhead. The per-node dedupe is also mechanism, not parallel logic: the read pass and the argument-binding pass dispatch the same node twice, and without dedupe each `$req->email` reaching a sink reports 2x.
- **A standalone FormRequest metadata registry.** That registry effectively already exists, distributed across three places: the FormRequest class set (`FormRequestPropertyRegistrationHandler`), the memoized resolved rules (`ValidationRuleAnalyzer::getRulesForFormRequest`), and the per-(class, property) resolution cache (`FormRequestPropertyHandler::$cache`). Introducing a new aggregate type on top would add code, not remove it, so I kept the existing pieces and routed the taint paths through the shared resolver instead.

Per-class property registration stays (Psalm property lookup is exact-class), as you noted.

alies-dev:
Round-2 simplifications applied (net -328 lines, behaviour unchanged — 483 type + 792 unit pass, self-analysis clean):

- **[P2] Parse the method-call shape once.** `fromMethodCall` now resolves the caller class a single time and splits into `methodSourcesInput` + `methodEscape`, replacing the four passes (`isValidationMethodCall` / `isValidatedInputAccessor` / `matchKeyedAccessor` / `resolveFormRequestForAccessor`) that each re-walked the same `MethodCall` and caller type. Name-bail stays first so non-accessor calls still short-circuit before any type resolution.
- **[P3] Merged `FormRequestPropertyRegistrationHandler` into `FormRequestPropertyHandler`.** The handler now implements `AfterCodebasePopulatedInterface` and owns its own per-subclass registration + the FormRequest class set. One fewer file and one fewer Plugin.php block.
- **[P3] Trimmed the docblocks** across the validation taint files (and the property type test) to invariants only.
- **`ValidatedFieldRead`** is now two properties + `@psalm-immutable`, param docs only.

One I handled differently: rather than fold the shared rule matrix from `FormRequestSelfPropertyTest.phpt` into the input test, I kept the property assertions in place (so a property-path regression fails a property test, not the input one) and only cut the prose. The file dropped from 221 to ~160 lines without losing coverage locality. Happy to merge them if you'd still prefer one matrix.

alies-dev:
## psalm-delta (base 756e09fd → head 6def9e46, 18 real-world apps)

Net **+0** issues, **0.00** coverage delta, **−2.22s** total time (noise). Only one app moves:

| pixelfed | 2 | 2 | +0 | +0.00 | -0.22s |
| _(17 others)_ | 0 | 0 | +0 | +0.00 | — |
| **Total / Weighted** | **2** | **2** | **+0** | **+0.00** | **-2.22s** |

**Issue type movements:**

| TaintedSSRF | 2 | 0 | +2 |
| UndefinedMagicPropertyFetch | 0 | 1 | -1 |
| UndefinedThisPropertyFetch | 0 | 1 | -1 |

This is the feature working end-to-end: pixelfed's `$request->field` magic reads stopped raising `Undefined{Magic,This}PropertyFetch` (the plugin now recognises them as validated reads, −2 false positives), and the paired re-source then traced that validated input into an SSRF sink (+2 real `TaintedSSRF` findings). No coverage or performance regression anywhere.

alies-dev:
## Delta breakdown (pixelfed) + perf

### TaintedSSRF +2 — real findings surfaced

Both new SSRF findings trace to the **same source**, `$this->acct` — the property whose false-positive `UndefinedThisPropertyFetch` this PR removes. Narrowing turns that magic read into a recognised validated source, which then flows into two outbound HTTP calls:

| Sink | Source | Hops |
|------|--------|-----:|
| `app/Services/FetchCacheService.php:56` — `->get($url)` | `$this->acct` | 27 |
| `app/Services/WebfingerService.php:79` — `->get($url)` | `$this->acct` | 26 |

Origin: `app/Http/Requests/ProfileMigrationStoreRequest.php:62` — `$acct = WebfingerService::rawGet($this->acct)`.

### Undefined*PropertyFetch −2 — false positives cleared

| Location | Type | Code |
|---|---|---|
| `ProfileMigrationStoreRequest.php:62` | `UndefinedThisPropertyFetch` | `WebfingerService::rawGet($this->acct)` |
| `UserAppSettingsController.php:42` | `UndefinedMagicPropertyFetch` | `'common' => $request->common` |

Both are FormRequests with a matching rule for the accessed field. The `acct` one is the head of both SSRF traces — same read both clears a FP **and** exposes a real vuln; the `common` one just clears a FP.

### Perf

Negligible, within run-to-run noise. **Total wall 233.63s → 231.41s, Δ −2.22s (−1.0%)** across 18 apps; no systematic overhead. Peak memory flat (pixelfed 23 → 22 MB). The merged registration handler adds no per-expression cost — the taint resolver's leading `instanceof` bail keeps non-accessor expressions free. Largest swings (coolify −0.77s, laravel-excel +0.41s) are jitter.

PR #872, opened 2026-05-05
🔎 Explore Blade-aware taint analysis

alies-dev:
## Issue to Solve
Explore a low-false-positive path for restoring Blade view-layer XSS taint coverage after whole-array `html` sinks were removed from view data.

## Related
Refs #581

## Solution Description
- Adds a Blade template scanner spike that classifies escaped, raw, and `@php` output contexts.
- Builds a `BladeSafetyMap` from view roots to unsafe top-level keys used in raw output.
- Documents the tradeoffs between scanner-based refinement, compiled Blade analysis, and `view:cache`-based approaches.
- Adds focused unit coverage for scanner behavior, view-path resolution, escaped braces, comments, verbatim blocks, scope locals, line preservation, and known limitations.

Verification:
- `composer test:unit -- --filter Blade --no-progress --colors=never --display-errors --display-warnings`
- `composer psalm -- --no-progress --no-suggestions --output-format=compact`

## Checklist
- [x] Tests cover the change (type test in `tests/Type/` and/or unit test in `tests/Unit/`)
- [x] Documentation is updated (if needed, otherwise remove this item)

Issue #581, opened 2026-03-27
Explore Blade template-aware taint analysis to reduce false positives


alies-dev:
Currently, all view data parameters are marked as `@psalm-taint-sink html` (#541, PR #580). This is correct-by-default (assumes worst case), but produces false positives when templates only use escaped `{{ }}` output.

The actual XSS risk depends on how data is rendered in Blade:
- `{{ $var }}` compiles to `<?php echo e($var); ?>` — safe (htmlspecialchars)
- `{!! $var !!}` compiles to `<?php echo $var; ?>` — dangerous
- `@php echo $var @endphp` — dangerous

### A. Status quo — accept false positives
- Effort: none
- Users baseline false positives from safe `{{ }}` usage

### B. Simple Blade scanner — report `{!! !!}` as a custom issue
- Effort: 1-2 weeks
- A handler collects `view('name', $data)` calls, resolves to `.blade.php`, scans for `{!! !!}`, reports `UnsafeBladeRawOutput`
- Pattern-based, no taint graph integration

### C. Template-aware sink refinement — only sink keys used in `{!! !!}`
- Effort: 2-4 weeks
- At boot, scan all `.blade.php` files and build a map of `template -> variables used in {!! !!}`
- Intercept `view()`/`Factory::make()` calls — only apply `@psalm-taint-sink html` for data keys that appear in raw output
- Eliminates most false positives without full compiled-PHP analysis
- Does not track through `@include` chains or components

### D. Compile-and-analyze — Bladestan-like for Psalm taint
- Effort: 2-6 months
- Compile Blade to PHP, generate synthetic files bridging data arrays via `extract()`, let Psalm trace taint through compiled templates
- Full taint tracking — Psalm sees `echo e($var)` (safe) vs `echo $var` (dangerous)
- Enormous complexity: template inheritance, components, dynamic includes

### E. Standalone CLI tool
- Effort: 1-2 weeks
- Separate tool scanning `.blade.php` for `{!! !!}` and `@php echo`
- No Psalm integration, no taint context

## Prior art

- **Bladestan** (tomasvotruba/bladestan): PHPStan extension that compiles Blade to PHP and analyzes the output. Type checking only, no taint analysis.
- **TwigStan**: Same concept for Twig/Symfony.
- No existing tool does taint analysis through Blade templates.

## Key challenges

- Blade compilation requires runtime (plugin already boots Laravel, so compiler is available)
[...]

PR #1023, opened 2026-05-31
Blade: cover anonymous `<x-foo />` component data flow

alies-dev:
## Issue to Solve

Anonymous Blade component tags (`<x-foo :bar="$expr" />`) classified as `ComponentTag` UNKNOWN by the v1 scanner (PR-1 #920), forcing any template using them to the whole-data sink fallback at the handler layer. That removes per-key precision for the most common Blade composition pattern in modern Laravel apps and inflates false-positive rate on `view()` calls that pass safe data to a template containing a component tag.

## Related

Sub-PR of #872 (umbrella branch for #581 Blade taint exploration). Stacks on PR-1 #920, PR-3 #926, PR-4 #934, PR-5 #985, PR-6a #995.

## Solution Description

Anonymous component tags now record `BladeComponentEdge` data-flow edges that the safety map's fixed-point propagation pass folds into the parent template's unsafe-key set, mirroring PR-4's literal-`@include` shape but with component-specific propagation semantics.

### Scope

- `BladeComponentTagParser` extracts self-closing tag records from the post-raw-block source via a hand-rolled char-by-char scanner that tracks quote and paren state. Required because Blade attribute values routinely contain PHP expressions (`<x-foo :bar="$x => 'a'" />`, `<x-foo @class([...]) />`) whose `>` characters break naïve regex anchoring.
- `BladeComponentEdge` carries Laravel's three anonymous-component candidate view names per `ComponentTagCompiler::guessAnonymousComponentUsingPaths()` (`components.{name}`, `components.{name}.index`, `components.{name}.{last-segment}`). The safety map resolves the first candidate present in the scanned view roots.
- Attribute names are camelized (`user-name` -> `userName`) to match the variable identity the child template sees as a scope local.
- `BladeSafetyMap::propagateEdges()` (renamed from `propagateIncludeEdges`) walks both include and component edges in one fixed-point pass. Cycle detection is unified across edge types so a cross-kind cycle (`@include('child')` plus child's `<x-parent />`) is detected.
- Component propagation has **no mergeData fall-through** (unlike includes): a child unsafe key contributes to the parent only when the parent binds it via an attribute. The `$attributes` bag reads (`$attributes->get('bio')`, `$attributes->whereStartsWith(...)`) are special-cased to union every parent variable bound on the edge, since the bag exposes every attribute without HTML escaping.
- New `BladeUncertaintyReason::ComponentResolved` is the intermediate marker (mirrors `IncludeResolved`); `ComponentTag` keeps its role as the "scanner could not model this" UNKNOWN reason. A single `BladeUncertaintyReason::isIntermediate()` predicate centralises the eligibility / stripping check so future intermediate markers do not need synchronized edits to two unrelated `BladeSafetyMap` methods.

### Out of scope for v1 (forced to `ComponentTag` UNKNOWN, deferred to follow-up PR)

- Opening tags with bodies (slot content).
- Namespaced anonymous components (`<x-mail::message />`).
- `<x-dynamic-component>`.
- Legacy `@component(...)` / `@slot(...)` directive forms.
- Class components (any `<x-name>` whose name does not resolve to an anonymous template on disk).
- Attribute strings containing `@class(...)` / `@style(...)` / `{{ $attributes }}` echoes / variable-variable expressions.
- Variable-bound component-builder chains (deferred to a future PR alongside the same gap for `view()`-builder chains).

### Implementation notes

- `PsalmBladeCompiler::compileComponentTags()` no longer flips a bare `sawComponentTag` flag. It captures the post-raw-block source on `lastTagScanSource()` so the scanner owns all classification. The compiler stays focused on source-to-PHP transformation; the parser stays focused on tag tokenisation; the scanner stays focused on producing a `BladeTemplateAnalysis`.
- `BladeSafetyMap::finaliseSafetyForView()` was switched from return-the-accumulator to `array &$finalised` by reference. The former shape forced PHP to copy-on-write the hashtable bucket array on every recursive descent (O(depth x finalised_size) extra allocations during boot). The public seam still narrows the `non-empty-string` key invariant on its way out.
- The bound vs static duplicate-attribute case (`<x-foo bar="literal" :bar="$x" />`, which Laravel resolves last-write-wins via `mapWithKeys()`) is soundness-handled by skipping the static loop's append when the camelized name is already in the bound map. Source order is lost in `BladeComponentTagParser`'s partitioned output, so we err toward the bound entry (over-reports taint at worst, never under-reports).

### Decisions and trade-offs

- The `$attributes` bag pathway is conservatively over-tainting: any child raw-echo sourced from `$attributes` unions every parent-bound var on the edge, not just the one the reader name selects. Bag reads (`only(['k'])`, `whereStartsWith('k')`) can mix many attributes at runtime, so per-key precision is unrecoverable from the static side without modeling each `ComponentAttributeBag` method individually.
- Component edge view-name resolution lives on the scanner side today (`BladeAstAnalysisVisitor::componentCandidateViewNames()`) rather than on the edge VO or a dedicated resolver. The architecture-agent suggested extracting; deferred until a second consumer appears, since the producer and the safety-map resolver are the only two call sites today.
- Class components stay UNKNOWN. A class component's render output is produced at runtime by a `View\Component::render()` method that the static analyzer cannot inspect. Modeling them would require either compiled-Blade analysis (the long-term approach D in the exploration doc) or a per-class stub for every component a user defines.

## Verification

- `composer test:unit -- --filter Blade --no-progress` — 221 Blade unit tests pass (40 new in this PR).
- `composer test:type -- --no-progress` — 432 type tests pass.
- `composer psalm -- --no-progress --no-suggestions --output-format=compact` — clean, 100% type coverage.
- `composer cs -- --show-progress=none --no-ansi -n` — clean.
- `composer rector -- --no-progress-bar --no-ansi` — clean.

## Checklist

- [x] Tests cover the change (unit tests in `tests/Unit/Blade/`).
- [x] Documentation is updated (`docs/issues/581-blade-taint-exploration.md` PR-6b row).


alies-dev:
closing as an outdated PoC

PR #1055, opened 2026-06-13
Fix `mixed` collapse when chaining off relation methods (replace `$this` by `static` in generic templates, use `@template-covariant` for relationships)

### LINKED ISSUE #913: MixedMethodCall when chaining any method off a relation factory (`$this->belongsTo()/hasMany()/...->...`) inside relation methods

alies-dev:
## Issue to Solve

Chaining a Builder/Relation method directly off a relation factory inside a relation method collapsed the receiver to `mixed` and raised `MixedMethodCall`:

```php
final class AbandonedCart extends Model
{
    /** @return BelongsTo<Customer, self> */
    public function customer(): BelongsTo
    {
        return $this->belongsTo(Customer::class)->withoutGlobalScopes(); // MixedMethodCall
    }
}
```

The relation return docblocks templated the declaring model as `$this`, and Psalm 7 does not late-static-substitute `$this` when the returned relation is chained, so `Relation<TRelated, $this>` degraded to `mixed`.

## Related

Closes #913

### Two coupled halves (doing only the first regresses)

1. **`$this` to `static` in the 11 relation return docblocks** of `HasRelationships.phpstub` (`hasOne`, `hasOneThrough`, `morphOne`, `belongsTo`, `morphTo`, `hasMany`, `hasManyThrough`, `morphMany`, `belongsToMany`, `morphToMany`, `morphedByMany`). Unlike `$this`, Psalm substitutes `static` when the relation is chained.
2. **`@template TDeclaringModel` to `@template-covariant TDeclaringModel`** across the 14 relation classes that declare it (`Relation`, `BelongsTo`, `HasOne`, `HasMany`, `HasOneOrMany`, `HasOneOrManyThrough`, `HasOneThrough`, `HasManyThrough`, `BelongsToMany`, `MorphTo`, `MorphOne`, `MorphMany`, `MorphOneOrMany`, `MorphToMany`). On a non-final model the factory infers `BelongsTo<Customer, Order&static>`, which only satisfies a call-site `@return BelongsTo<Customer, self>` under covariance (`Order&static` is a subtype of `Order`). With the previous invariant template this failed with `InvalidReturnStatement` (invariance demands equality).

### Soundness audit

Covariance on `TDeclaringModel` is sound only if it is never consumed through an input (contravariant) position. Every `TDeclaringModel` occurrence across the 14 stubs was audited:

- Output positions only on the public surface: `getParent(): TDeclaringModel`, the protected `$parent` property (written only at construction, from a plain `Model`), and `@template-extends` propagation.
- The single input-position use is the internal `HasOneOrManyThrough::__construct($farParent)` parameter, which is variance-exempt (a constructor is never invoked through a covariantly-widened reference).
- `BelongsTo::associate()` takes the related model, not the declaring model, so it is unaffected.
- Real Laravel has contravariant `TDeclaringModel` uses (`addEagerConstraints`, `match`, `getRelationExistenceQuery`, `newRelatedInstanceFor`), but they are either `protected` or absent from these (leaner) stubs. Verified empirically that Psalm still enforces the reflected `addEagerConstraints(array<TDeclaringModel>)` contravariant check on a precisely typed relation, so covariance does not mask it. The only theoretical masking would require an explicit widen-to-base followed by a contravariant call, a path the stub surface does not expose and which never flows through the stub return types.

Verdict: covariance is sound for `TDeclaringModel` in these stubs. The `Relation.phpstub` rationale comment records the invariant and tells a future stub sync to revisit if a non-constructor `TDeclaringModel` parameter is ever added.

### Laravel divergence

Laravel source uses `$this` and an invariant `@template TDeclaringModel`. Both `static` and the covariance are deliberate Psalm-7-specific divergences (Psalm does not late-static-substitute `$this`). Larastan keeps `TDeclaringModel` invariant because its own stubs carry `Builder<TDeclaringModel>` parameter positions and PHPStan substitutes `$this`/`static` differently, so the covariance should not be synced away to match it.

### Delta (17 real-world apps, base `ed1ce450` vs head)

| Movement | Count | Notes |
|---|---|---|
| `InvalidReturnStatement` removed | -2 | bookstack `Book::cover` / `Bookshelf::cover` |
| `InvalidReturnType` removed | -2 | same two methods |
| `MixedArgument` / `MixedArgumentTypeCoercion` added | +2 | filament, thread-scheduling noise (see below) |

**Removed (wins).** Both bookstack methods declare `@return BelongsTo<Image, Entity>` (the base class, via `HasCoverInterface`) while the concrete declaring model is `Book&static` / `Bookshelf&static`. `Book` and `Bookshelf` extend `Entity`, so under covariance `Book&static` is a subtype of `Entity` and the annotation is satisfied. Under the old invariant template Psalm wrongly demanded the declaring model be exactly `Entity`. These are sound false-positive removals (the declaring model is output-only, so a caller reading `getParent()` gets an `Entity`, which a `Book` is), not masked real errors. Confirmed by reading both sites.

**Added (not introduced by this PR).** The filament +2 are at `HasBulkActions.php:120-121`, a `Collection::reduce` closure unrelated to relation declaring-model templates. Re-running filament single-threaded (deterministic) gives byte-identical issues on base and head (both 13999 total), so the multi-threaded +2 is Psalm thread-scheduling noise, not an effect of this change.

**No disappeared real-positives.** Across all 17 apps the only removed issue types are the #913-class `InvalidReturnType` / `InvalidReturnStatement` false positives, both verified sound. The delta produced zero `MixedMethodCall` removals despite the issue title: real apps express #913 as `InvalidReturnType` on a declared relation method rather than as a raw-chain `MixedMethodCall`, so the bookstack wins are the real-world form of this bug.

### Verification

`composer test:type` 452/452, `composer test:unit` 785, `composer test:app` clean, `composer psalm` self-analysis clean at 100% coverage.

### Release

Recommended for v4.13.2: contained, false-positive-only, delta-clean, no breaking change. Does not warrant a dedicated relation-template release.

## Checklist
- [x] Tests cover the change (rewritten `BelongsToChainSelfTemplateCollapseTest.phpt` plus new `RelationStaticDeclaringModelTest.phpt`: final vs non-final, `hasMany()->where()`, a through-relation, a morph relation, and a child-model `static` resolution parallel to #1043)
- [x] Documentation is updated (covariance rationale in `Relation.phpstub` plus a one-line guard comment in each relation stub, and the stub-authoring rule)

## Known limitations (unmasked by this fix, tracked)

Removing the `mixed` collapse means relation-method return expressions are now type-checked instead of skipped. On strict codebases this surfaces two pre-existing Psalm 7 limitations (not plugin bugs). Both are pinned in `tests/Type/tests/Relation/RelationStubKnownLimitationsTest.phpt`, asserting the current output so they flip loudly when upstream lands:

1. **Pivot relations (`belongsToMany`/`morphToMany`/`morphedByMany`) raise `MissingTemplateParam` on chained calls.** These classes declare 4 templates with the last two (`TPivotModel`/`TAccessor`) defaulted; Psalm 7 does not honor `@template` defaults for a partial generic, so the stub's (and Laravel's own) under-filled `@return` trips. Laravel ships under-filled returns (3-of-4) that pass under PHPStan/Larastan, which honor the defaults. The return is deliberately NOT completed to 4 args: pinning `TAccessor='pivot'` false-positives legitimate `->as('name')` accessor overrides, since Psalm's `@psalm-this-out` generic narrowing is also inert. Tracked upstream: vimeo/psalm#5407 and PR vimeo/psalm#11790.
2. **`morphTo()` with a narrowed related type (`@return MorphTo<Concrete|..., self>`) raises `MoreSpecificReturnType`/`InvalidReturnType`.** The morph target is resolved at runtime via the morph map, so the stub can only honestly return `MorphTo<Model, static>`. Inherent to polymorphic relations; resolvable app-side (widen the annotation) or by a future morph-map-aware resolver, not by a stub or variance change.


alies-dev:
Folding in ideas from #917 (the earlier, half-fix attempt at #913) before closing it as superseded by this PR. Capturing what is worth keeping so it is not lost:

**1. `ModelRelationReturnTypeHandler` redundancy (follow-up worth a separate pass).**
`ModelRelationReturnTypeHandler` was introduced specifically to work around the same `$this`-in-template Psalm limitation for user-defined relation methods. Now that the factory stubs substitute `static` directly, parts of that handler may be redundant. Not in scope here (this PR is false-positive-only and delta-clean), but a candidate cleanup once this lands. Caveat: the handler also fills the trailing pivot/accessor template defaults that Psalm 7 does not honor (the `MissingTemplateParam` limitation pinned in `RelationStubKnownLimitationsTest`), so any simplification must confirm it is not the thing keeping those slots populated.

**2. Bare-intermediate assertion matrix per relation shape.**
This PR asserts the bare factory type at one site (`BelongsTo<Customer, Order>` on an unchained `belongsTo`), which is the load-bearing guard: a future regression that re-collapses the intermediate to `mixed` would still satisfy any *declared* return type, so the only thing that catches it is a direct `@psalm-check-type-exact` on the unchained factory call. #917 carried this guard across all four structurally distinct shapes (2-template `belongsTo`/`hasOne`/`hasMany`, 3-template through `hasManyThrough`, 4-template pivot `belongsToMany`, no-`TRelated` `morphTo`). Worth considering whether to extend the bare-intermediate exact checks to the through and pivot shapes here too, since those are the slots most likely to silently shift (`static` landing in the wrong template position).

**3. Why #917 looked complete but was not.**
#917's regression test used only `final` models. On a `final` model `static` resolves to the exact class, so the covariance half is never exercised and the test passed. The bug only surfaces on a non-final model where `static` resolves to `Model&static` and the invariant template demands equality. This PR's final-vs-non-final coverage (`RelationStaticDeclaringModelTest`) is exactly the gap #917 missed, and the reason the two coupled halves are both required.

Closing #917 in favor of this.

alies-dev:
Closing #870 in favor of this PR. Both target the same root defect (Psalm 7 not late-static-substituting the declaring-model template slot when a relation factory is chained inside a relation method), and both carry the identical second half: `@template-covariant TDeclaringModel` across the 14 relation classes.

The mechanisms diverge on the first half:

- #870 took the **handler** route: `ModelRelationReturnTypeHandler` inspects the call statement, and on `$this` dispatch emits `TNamedObject($bindingClass, is_static: true)`. That path depends on vimeo/psalm#11768 (so that `$this` parses in template position) and is a no-op on stock Psalm 7.
- This PR takes the **stub** route: `$this` -> `static` in the `HasRelationships.phpstub` return docblocks. No upstream dependency, works on the Psalm 7 we ship against today.

The stub route wins: smaller surface, no dependency on an unreleased upstream change, and no per-call handler cost on a hot path.

Two ideas from #870 worth carrying forward / double-checking here:

1. **Receiver-awareness.** #870 deliberately distinguished `$this` dispatch from external dispatch (`(new Model())->relation()`), attaching `&static` only on the former so that external-dispatch assertions stayed as `Relation<X, ConcreteClass>` without `&static`. The stub `static` resolves uniformly regardless of receiver, which is simpler, but it's worth an explicit type test that external/instance dispatch (`(new Order())->customer()`) still infers `BelongsTo<Customer, Order>` and does not pick up an unexpected `&static`. If `RelationStaticDeclaringModelTest.phpt` doesn't already pin the non-`$this` receiver case, adding it would lock in the behavior #870 was careful about.

2. **Backward-compat framing.** #870 verified its change was a no-op under stock Psalm 7 (no #11768). Inverse risk applies here: confirm the stub `static` does the right thing once #11768 *does* land in a Psalm release (it should, since `static` substitution is the path that already works) so we don't regress when upstream catches up.

Everything else in #870 is subsumed. Closing it.

PR #1057, opened 2026-06-13
Resolve docblock-only `@method scopeXxx` scopes on builders

### LINKED ISSUE #1054: Docblock @method scopeXxx scopes resolve to mixed (pseudo-methods not consulted)
**Describe the bug**

Scopes declared only as a class-level `@method` docblock annotation (no concrete `scopeXxx()` body, e.g. scopes injected at runtime by a package/trait and documented via `@method`) are not resolved: calling them on a builder returns `mixed`. The plugin resolves real `scopeXxx()` methods and `#[Scope]` methods, but does not consult `@method`-declared pseudo-methods for the `scope`-prefixed form.

**Code to reproduce**
```php

/**
 * @method static Builder<static> scopePublished()
 */
class Article extends Model {}

function probe(): void
{
    $r = Article::query()->published();
    /** @psalm-check-type-exact $r = Builder<Article> */   // actual: mixed
}
```

**Impacted Versions**
```text
psalm/plugin-laravel  dev-master @ 939d9b6f
vimeo/psalm           7.0.0-beta19
laravel/framework     v13.15.0
PHP 8.4
```

**Psalm config**
```xml
<?xml version="1.0"?>
<psalm errorLevel="3" findUnusedCode="false" xmlns="https://getpsalm.org/schema/config">
    <projectFiles><directory name="app"/></projectFiles>
    <plugins><pluginClass class="Psalm\LaravelPlugin\Plugin"/></plugins>
</psalm>
```

**Additional context**

Psalm stores class-level `@method` annotations as `pseudo_methods` / `pseudo_static_methods` on `ClassLikeStorage`. The scope detection (`BuilderScopeHandler::getScopeParams` / `hasScopeMethod`) could fall back to a `scope`-prefixed pseudo-method lookup (stripped params, same builder-return fabrication as real scopes) when neither a real `scopeXxx()` method nor a `#[Scope]` method exists. Larastan supports this (`BuilderHelper` consults `@method scopeXxx` tags). Used by packages and legacy codebases that annotate runtime-injected scopes.

alies-dev:
## Issue to Solve

Eloquent scopes documented only as a class level `@method scopeXxx()` tag (no concrete `scopeXxx()` body and no `#[Scope]` method) resolved to `mixed` when called on a Builder. For a model annotated `@method static Builder<static> scopePublished()`, `Article::query()->published()` inferred `mixed` instead of `Builder<Article>`.

## Related

Closes #1054

## Solution Description

`BuilderScopeHandler::getScopeParams()` gains a third fallback, after a real `scopeXxx()` method and a `#[Scope]` method: a new `resolvePseudoScopeParams()` reads the model's `pseudo_static_methods` / `pseudo_methods` from `ClassLikeStorage`, drops the leading `$query` that a real scope receives, and reuses the existing `self`/`static` param expansion. The return is fabricated as `Builder<TModel>`, exactly as for a real (often void) scope. This mirrors Larastan's `BuilderHelper`, which shifts the first `@method` scope param the same way. Because Psalm's `Populator` flattens trait and parent `@method` tags into a model's local pseudo maps, scopes documented on a `use`d trait (the package or trait case the issue describes) or on an abstract parent resolve through the same path. The static `Model::scope()` form is covered too, via the shared `hasScopeMethod()`.

Why trust the tag: a scope prefixed `@method scopeXxx()` with no backing real method is not dispatchable at runtime (`Model::hasNamedScope()` is `method_exists` gated, so the bare call throws `BadMethodCallException`). The plugin trusts the developer's `@method` assertion the way Psalm and PHPStan trust any pseudo method, accepting that a wrong tag types runtime fatal code. That is the deliberate Larastan parity trade off.

Known limitation: a `self` typed param on a tag inherited from a parent over narrows to the queried subclass, because Psalm records no declaring class for a static pseudo method (`declaring_pseudo_method_ids` is instance only), so the real scope `getAppearingMethodId` path has no equivalent. The common direct and trait merged forms are exact.

## Solution Verification

New type test `tests/Type/tests/Builder/PseudoMethodScopeTest.phpt` covers the builder instance and static model forms, the `$query` strip arity (too few, too many, wrong type), optional and variadic post `$query` params, trait merged and parent inherited resolution, and the instance vs static `@method` map distinction. Full suite green: `composer test:type` (452), `composer test:unit` (785), `composer psalm` (100% coverage, no errors), `composer cs`, `composer rector`.

## Checklist
- [x] Documentation is updated (handler docblocks document the resolution order, the trust the tag rationale and runtime caveat, and the parent `self` limitation)

### DIFF
--- a/src/Handlers/Eloquent/BuilderScopeHandler.php
+++ b/src/Handlers/Eloquent/BuilderScopeHandler.php
@@ -32,11 +32,17 @@ use Psalm\Type\Union;
  * When a method call on Builder doesn't match a real method, checks the model for:
  * 1. Legacy scopeXxx() methods (e.g., scopeActive → active())
  * 2. Methods with #[Scope] attribute (e.g., #[Scope] active() → active())
- * 3. Trait-declared builder methods (e.g., SoftDeletes::withTrashed → withTrashed())
+ * 3. Docblock-only scopes documented as a class-level `@method scopeXxx()` tag — no concrete
+ *    scopeXxx() body and no #[Scope] method. Resolved from the model's pseudo_static_methods/
+ *    pseudo_methods, after 1 and 2, by trusting the tag (Larastan parity; note the tag is NOT
+ *    runtime-dispatchable on its own). See #1054 and resolvePseudoScopeParams.
+ * 4. Trait-declared builder methods (e.g., SoftDeletes::withTrashed → withTrashed())
  *    These are registered as Builder macros at runtime via global scopes
  *    (e.g., SoftDeletingScope::extend), and declared via @method static returning
  *    Builder<static> on the model trait.
- * 4. @method PHPDoc scopes are handled natively by Psalm
+ * 5. Public-form `@method published()` scopes (the bare call name documented directly, not the
+ *    scope-prefixed form) are handled natively by Psalm — the call name matches the tag, so no
+ *    scope-prefix lookup or $query stripping applies.
  *
  * @internal
  */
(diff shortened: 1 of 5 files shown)


alies-dev:
I tested it on banch of real apps and packages — no real cases so far, clsing this as not needed

Issue #1031, opened 2026-06-11
InvalidArgument false positive: `self` param in trait-declared scope resolves to the custom builder class


alies-dev:
**Describe the bug**

When a model uses a custom Eloquent builder and a legacy scope is declared **in a trait** with a non-query parameter typehinted `self`, calling that scope on the builder instance emits a false-positive `InvalidArgument`. Psalm reports the expected type of the argument as the custom builder class instead of the model:

```
ERROR: InvalidArgument
Argument 1 of App\Builders\RankedPostBuilder::rankedabove expects App\Builders\RankedPostBuilder,
but App\Models\RankedPost provided (see https://psalm.dev/004)
```

The same scope declared directly on the model class (not via a trait) does not trigger the issue.

**Code to reproduce**

`app/Builders/RankedPostBuilder.php`:

```php
<?php

namespace App\Builders;

use App\Models\RankedPost;
use Illuminate\Database\Eloquent\Builder;

/** @extends Builder<RankedPost> */
final class RankedPostBuilder extends Builder
{
}
```

`app/Models/Concerns/ComparesRank.php`:

```php
<?php

namespace App\Models\Concerns;

use Illuminate\Database\Eloquent\Builder;

/** @psalm-require-extends \Illuminate\Database\Eloquent\Model */
trait ComparesRank
{
    /** @param Builder<self> $query */
    public function scopeRankedAbove(Builder $query, self $model): void
    {
        $query->whereKeyNot($model->getKey());
    }
}
```

`app/Models/RankedPost.php`:

```php
<?php

namespace App\Models;

use App\Builders\RankedPostBuilder;
use App\Models\Concerns\ComparesRank;
use Illuminate\Database\Eloquent\Builder;
use Illuminate\Database\Eloquent\Model;

final class RankedPost extends Model
{
    use ComparesRank;

    /** @param Builder<self> $query (same signature, but declared directly on the model: no issue) */
    public function scopeRankedBelow(Builder $query, self $model): void
    {
        $query->whereKeyNot($model->getKey());
    }

    public function newEloquentBuilder($query): RankedPostBuilder
    {
        return new RankedPostBuilder($query);
    }
}
```

Analyzed snippet:

```php
<?php

use App\Models\RankedPost;

$post = new RankedPost();

RankedPost::query()->rankedAbove($post); // InvalidArgument: expects RankedPostBuilder, RankedPost provided
[...]

PR #1061, opened 2026-06-14
Honor value-returning Eloquent scope return types

### LINKED ISSUE #1053: Value-returning scopes typed as Builder<Model>, ignoring the scope's declared return (callScope ?? $this)
**Describe the bug**

A local scope is not required to return the builder. Laravel's `Builder::callScope()` returns `$scope(...) ?? $this`, so a scope whose body `return`s a value (e.g. `->first()`, a count, a collection) propagates that value to the caller. The plugin types every forwarded scope call as `Builder<Model>`, ignoring the scope's declared return type. Downstream this produces false positives on the real return: a null check on a `?Model`-returning scope is flagged `TypeDoesNotContainNull`/redundant, and member access goes through the wrong type.

**Code to reproduce**
```php

class Post extends Model
{
    /** @param Builder<self> $query */
    public function scopeFirstPublished(Builder $query): ?self
    {
        return $query->where('published', true)->first();
    }
}

function probe(): void
{
    $r = Post::query()->firstPublished();
    /** @psalm-check-type-exact $r = Post|null */   // actual: Builder<Post>

    if ($r === null) { return; }                    // actual: TypeDoesNotContainNull (false positive)
}
```

**Impacted Versions**
```text
psalm/plugin-laravel  dev-master @ 939d9b6f
vimeo/psalm           7.0.0-beta19
laravel/framework     v13.15.0
PHP 8.4
```

**Psalm config**
```xml
<?xml version="1.0"?>
<psalm errorLevel="3" findUnusedCode="false" xmlns="https://getpsalm.org/schema/config">
    <projectFiles><directory name="app"/></projectFiles>
    <plugins><pluginClass class="Psalm\LaravelPlugin\Plugin"/></plugins>
</psalm>
```

**Additional context**

`Builder<Model>` is correct for the common void/builder-returning scope (the `?? $this` fallback). The bug is only for scopes whose declared return is non-void and not a builder. Proposed fix: when the scope's declared return type is non-void and not a `Builder` subtype, return `declared | Builder<Model>` (dropping `null` from the declared part, since a null scope result falls back to the builder). The same `?? $this` semantics affect `Builder::scopes(...)` (typed `static|mixed` upstream); a consistent decision should cover both. Larastan also fabricates `Builder` here, so this would be a precision improvement over both tools.

alies-dev:
## Issue to Solve

A local Eloquent scope is not required to return the builder. Laravel's `Builder::callScope()` evaluates `$result = $scope(...) ?? $this`, so a scope whose body `return`s a value (`->first()`, a count, a collection) propagates that value to the caller. Only a null, void, or builder result falls back to the builder.

The plugin typed **every** forwarded scope call as `Builder<Model>`, ignoring the scope's declared return. Downstream this produced false positives: a null check on a `?Model`-returning scope was flagged `TypeDoesNotContainNull` (redundant), and member access routed through the wrong type.

```php
class Post extends Model
{
    /** @param Builder<self> $query */
    public function scopeFirstPublished(Builder $query): ?self
    {
        return $query->where('published', true)->first();
    }
}

$r = Post::query()->firstPublished();
// before: Builder<Post>     after: Post|Builder<Post>
```

## Related

Closes #1053

## Solution Description

Adds `BuilderScopeHandler::forwardedScopeReturnType()`, modeling Laravel's `$scope(...) ?? $this` coalesce on a forwarded scope call's return type:

- nullable value scope (`?Post`) becomes `Post | <fallback>`. The null result is what `?? $this` swaps for the builder, so the inferred type is never null (a downstream `=== null` check is now correctly reported redundant, matching the runtime).
- non-null value scope (`int`) becomes the declared type alone (the `?? $this` branch is dead).
- void, builder-returning, or undeclared scope keeps the unchanged `<fallback>` (prior behavior preserved).

`<fallback>` is the `$this` of `callScope`, supplied by each of the four forwarded scope-call paths it is wired into:

| Path | Call form | Fallback |
| :--- | :--- | :--- |
| `BuilderScopeHandler` | `Post::query()->scope()` | `Builder<Post>` |
| `ModelMethodHandler` | `Post::scope()` (static) | `Builder<Post>` |
| `MethodForwardingHandler` | `$user->posts()->scope()` | the `Relation` (`Relation::__call` maps a query-identical result back to `$this`) |
| `CustomBuilderMethodHandler` | `Post::query()->scope()` on a custom builder | `CustomBuilder<Post>` |

`self`, `static`, and `$this` in the declared return are pinned first (`self` to the scope's composing class, `static` / `$this` to the queried model, with `final: true`), so `?self` resolves to `?Post` before null is dropped. This mirrors the parameter pinning of #1031. A raw-type early-out keeps the common void/fluent scope on its prior fast path. Direct calls (`$this->scope($q)`) are unchanged: they bypass `callScope` and already use the real declared return.

Verification: two new type tests (`ValueReturningScopeTest.phpt`, `RelationValueScopeTest.phpt`) cover all four paths, nullable, non-null scalar, and non-null object value scopes, `self`-vs-`static` pinning on an abstract parent, the custom-builder fallback, and the void/builder regressions. Value-scope fixtures were added to the `Customer`, `Vehicle`, and `AbstractDocument` archetype models. Full suite green (Type 455, Unit 789, app integration), Psalm self-analysis at 100% coverage, CS clean.

## Checklist
- [x] Tests cover the change (type tests in `tests/Type/`)

### DIFF
--- a/src/Handlers/Eloquent/BuilderScopeHandler.php
+++ b/src/Handlers/Eloquent/BuilderScopeHandler.php
@@ -321,7 +321,7 @@ final class BuilderScopeHandler implements MethodReturnTypeProviderInterface, Me
         }
 
         $scopeMethodId = self::resolveScopeMethodId($codebase, $modelClass, $methodName);
-        if ($scopeMethodId === null) {
+        if (!$scopeMethodId instanceof \Psalm\Internal\MethodIdentifier) {
             return self::$scopeParamsCache[$key] = null;
         }
 
@@ -412,7 +412,7 @@ final class BuilderScopeHandler implements MethodReturnTypeProviderInterface, Me
         Union $fallback,
     ): Union {
         $scopeMethodId = self::resolveScopeMethodId($codebase, $modelClass, $methodName);
-        if ($scopeMethodId === null) {
+        if (!$scopeMethodId instanceof \Psalm\Internal\MethodIdentifier) {
             return $fallback;
         }
 
@@ -428,7 +428,7 @@ final class BuilderScopeHandler implements MethodReturnTypeProviderInterface, Me
 
         // No declared return, or `void`/`never`: the scope surfaces nothing, so `?? $this`
         // always falls back to the builder/relation — the long-standing behavior.
-        if ($declared === null || $declared->isVoid() || $declared->isNever()) {
+        if (!$declared instanceof \Psalm\Type\Union || $declared->isVoid() || $declared->isNever()) {
             return $fallback;
         }
 
--- a/src/Handlers/Eloquent/BuilderScopeHandler.php
+++ b/src/Handlers/Eloquent/BuilderScopeHandler.php
@@ -22,6 +22,7 @@ use Psalm\Plugin\EventHandler\MethodParamsProviderInterface;
 use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
 use Psalm\StatementsSource;
 use Psalm\Storage\FunctionLikeParameter;
+use Psalm\Type;
 use Psalm\Type\Atomic\TGenericObject;
 use Psalm\Type\Atomic\TNamedObject;
 use Psalm\Type\Union;
@@ -204,7 +205,9 @@ final class BuilderScopeHandler implements MethodReturnTypeProviderInterface, Me
             // Codebase\Methods::getMethodParams.
             self::$pendingScopeModel[$methodName] = $modelClass;
 
-            return $builderReturn;
+            // A value-returning scope surfaces its declared return via Laravel's `?? $this`
+            // coalesce; a plain void/fluent scope keeps $builderReturn unchanged (issue #1053).
+            return self::forwardedScopeReturnType($codebase, $modelClass, $methodName, $builderReturn);
         }
 
         // Trait-declared builder methods (e.g., SoftDeletes::withTrashed) are registered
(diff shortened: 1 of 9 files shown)


alies-dev:
## psalm-delta benchmark

Psalm + plugin run on 18 real-world Laravel apps, base `9f1a1669` vs PR head `c8fc5c92`. **Net effect: effectively zero, no regressions.**

### Per-app delta

| App | +Add | −Rem | ΔNet | ΔCov | ΔTime |
|---|--:|--:|--:|--:|--:|
| filament | 2 | 0 | +2 | −0.00 | −1.04s |
| private-project-1 | 0 | 1 | −1 | +0.00 | +0.60s |
| _16 other apps_ (aureuserp, bagisto, bookstack, coolify, corcel, monica, pixelfed, solidtime, spatie-dashboard, tastyigniter, unit3d, vito, laravel-excel, laravel-backup, laravel-permission, laravel-socialite) | 0 | 0 | 0 | 0.00 | — |
| **Total / Weighted** | **2** | **1** | **+1** | **−0.00** | **−5.54s** |

### Issue-type movements

| Issue type | +Add | −Rem | ΔNet |
|---|--:|--:|--:|
| `MixedArgument` | 1 | 0 | +1 |
| `MixedArgumentTypeCoercion` | 1 | 0 | +1 |
| `UndefinedClass` | 0 | 1 | −1 |

### Analysis

Both filament "additions" land on `packages/tables/src/Concerns/HasBulkActions.php:120-121`, the known multi-threaded-analysis flaky spot that oscillates ±2 on the base ref alone, independent of this PR. The single `UndefinedClass` removal in the private app is unrelated noise (no connection to scope return types).

Value-returning local scopes are uncommon in these codebases, so the precise `$scope() ?? $this` typing introduces no measurable issue churn. Type coverage and wall time are flat. The benchmark confirms the change is low-risk with no real-world regressions.

PR #1072, opened 2026-06-15
Narrow `App::make()` facade calls to the resolved class

alies-dev:
## Issue to Solve

`App::make(Foo::class)` on the `Illuminate\Support\Facades\App` facade infers `mixed`, unlike `app(Foo::class)`, `resolve(Foo::class)`, and `$container->make(Foo::class)`, which already resolve to `Foo`. Laravel's facade declares `make()`, `makeWith()`, and `get()` as `object|mixed`, which collapses to `mixed`. Any list or return built from such a call gets polluted (for example a Nova resource's `actions(): list<Action>` becomes `list{..., mixed, ...}`, triggering `MixedReturnTypeCoercion` / `LessSpecificReturnStatement`).

## Related

None.

## Solution Description

The obvious fix (adding the facade to `ContainerHandler::getClassLikeNames()`) does not work and actually crashes Psalm. `App::make()` is a magic static method (dispatched through `Facade::__callStatic`), so in `AtomicStaticCallAnalyzer` the return-type provider for that branch runs before the arguments are analysed (argument types are `null`), and once a non-null type is returned Psalm calls `getMethodParams()`, which throws `UnexpectedValueException("Cannot get method params")` for a method that has no real parameter storage.

This PR adds a dedicated `AppFacadeMakeHandler` that:

- Narrows `make()`, `makeWith()`, and `get()` when the first argument is a class-string, mirroring the container's conditional return type `($abstract is class-string<TClass> ? TClass : mixed)`.
- Reads the class off the AST node (`Foo::class`, `static::class`, `self::class`), since the resolved argument type is unavailable at that analysis stage.
- Supplies parameter storage through a `MethodParamsProvider` so analysis does not abort.
- Covers the canonical facade plus its aliases (the global `\App`) via `FacadeMapProvider`.

Plain-string aliases (`'view'`, `'config'`) and `class-string<Foo>` variables keep Laravel's declared `object|mixed`, so existing behaviour does not regress.

Verified with a new type test (`tests/Type/tests/Facades/AppFacadeMakeTest.phpt`) covering the narrowing cases, named arguments, `static`/`self`, the global alias, and the guard cases (plain string stays `mixed`, arity errors). The full type suite, unit tests, the fresh-app integration test, and self-analysis (100% type coverage) all pass.

## Checklist

### DIFF
--- a/src/Handlers/Facades/AppFacadeMakeHandler.php
+++ b/src/Handlers/Facades/AppFacadeMakeHandler.php
@@ -1,0 +1,200 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Facades;
+
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
+use Psalm\Codebase;
+use Psalm\LaravelPlugin\Providers\ApplicationProvider;
+use Psalm\LaravelPlugin\Providers\FacadeMapProvider;
+use Psalm\Plugin\EventHandler\Event\MethodParamsProviderEvent;
+use Psalm\Plugin\EventHandler\Event\MethodReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\MethodParamsProviderInterface;
+use Psalm\Plugin\EventHandler\MethodReturnTypeProviderInterface;
+use Psalm\StatementsSource;
+use Psalm\Storage\FunctionLikeParameter;
+use Psalm\Type;
+use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Union;
+
+/**
+ * Narrows `App::make()`, `App::makeWith()` and `App::get()` on the `Illuminate\Support\Facades\App`
+ * facade to the resolved class, mirroring the container's own conditional return type
+ * `($abstract is class-string<TClass> ? TClass : mixed)`.
+ *
+ * Why a handler and not a stub: Laravel declares these as `object|mixed` (collapses to `mixed`). A
+ * real stub method is shadowed by Laravel's reflected magic-method tag on static calls, and a
+ * class-docblock `@method` override can't express templates/conditional returns in Psalm 7.
+ *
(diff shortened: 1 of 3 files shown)


alies-dev:
## psalm-delta: real-world app benchmark

Ran Psalm + plugin on 18 real-world Laravel apps, base (`f4077aad`) vs this PR (`ad44cc2d`), delta only.

**Result: net −198 issues, no regressions, no crashes.**

17 public apps (aureuserp, bagisto, bookstack, coolify, corcel, monica, pixelfed, solidtime, spatie-dashboard, tastyigniter, unit3d, vito, laravel-excel, filament, laravel-backup, laravel-permission, laravel-socialite) show **zero delta**: none call `App::make(Class::class)` in a typed context. One private codebase (reported as `private-project-1`), which uses the pattern heavily, accounts for the whole movement.

| App | +Add | −Rem | ΔNet | ΔCov |
|---|--:|--:|--:|--:|
| 17 public apps (each) | 0 | 0 | 0 | 0.00 |
| private-project-1 | 24 | 218 | **−198** | +0.16 |
| **Total / Weighted** | **24** | **218** | **−198** | **+0.04** |

### Removed (−218): the intended cascade
Narrowing `App::make(Foo::class)` from `mixed` to `Foo` clears the downstream mixed errors it previously caused.

| Issue type | ΔNet |
|---|--:|
| `MixedMethodCall` | −112 |
| `MixedAssignment` | −47 |
| `MixedReturnStatement` | −24 |
| `MixedArgument` | −16 |
| `MixedArgumentTypeCoercion` | −4 |
| `LessSpecificReturnStatement` | −3 |
| `MixedArrayAccess` | −3 |
| `MixedReturnTypeCoercion` | −2 |

### Added (+24): findings previously hidden behind `mixed`
Once values carry a concrete type, stricter checks fire on real type info. These are surfaced findings, not plugin regressions.

| Issue type | ΔNet | note |
|---|--:|---|
| `MissingPureAnnotation` | +13 | info level, purity hygiene now visible on typed objects |
| `RedundantCondition` | +3 | e.g. an `instanceof Foo` guard now provably redundant |
| `ArgumentTypeCoercion` | +3 | narrowed type reaches a stricter parameter |
| `InvalidOperand` / `PossiblyInvalidArgument` / `RedundantCast` | +1 each | downstream of the now concrete type |

Type coverage rises (+0.16% on the affected app). No app crashed, timings stayed flat.

PR #1073, opened 2026-06-15
Report undefined methods on base `Builder<Model>` from `$class::query()`

### LINKED ISSUE #1070: False negative: undefined scope/method on dynamic `$class::query()` (class-string<Model>) not reported
## Summary

When a model class is held in a `class-string<Model>` variable and queried via `$fqcn::query()->someScope()`, the plugin does **not** report a call to an undefined scope/method on the builder. The same call written against a concrete model **is** correctly reported. This let a renamed scope reach production as a runtime `BadMethodCallException`.

## Minimal repro

```php
use App\Models\Post; // any model with a custom Eloquent builder

final class Repro
{
    /** @param class-string<\Illuminate\Database\Eloquent\Model> $fqcn */
    public function dynamic(string $fqcn): void
    {
        assert(is_subclass_of($fqcn, Model::class));
        $fqcn::query()->thisScopeDoesNotExist()->get(); // (1) NO error reported
    }

    public function concrete(): void
    {
        Post::query()->thisScopeDoesNotExist()->get();   // (2) UndefinedMagicMethod (correct)
    }
}
```

## Actual vs expected

| Call | Reported? |
|------|-----------|
| (2) concrete `Post::query()->thisScopeDoesNotExist()` | ✅ `UndefinedMagicMethod ... PostBuilder::thisscopedoesnotexist does not exist` |
| (1) dynamic `$fqcn::query()->thisScopeDoesNotExist()` | ❌ no error |

Expected: (1) should also be reported (or at least not silently accept any method name), since at runtime `$fqcn::query()` returns the concrete custom builder, which throws `BadMethodCallException` for the missing method.

## Root cause (as far as I traced it)

For the concrete call, the plugin resolves `Post::query()` to the concrete custom builder `PostBuilder<Post>`; `CustomBuilderMethodHandler::doesScopeMethodExistOnBuilder()` / the model→builder map then drive existence checks and Psalm emits `UndefinedMagicMethod` for a genuinely-missing scope.

For the dynamic call, `$fqcn::query()` (where `$fqcn` is `class-string<Model>`, even after `is_subclass_of` narrowing) resolves only to the base `Builder<Model>` template type, not the concrete custom builder. The base `Builder` has `__call` plus `@mixin Query\Builder`, so any method name is accepted permissively and the existence providers (keyed on the concrete builder→model map) never fire. The undefined scope/method is silently swallowed.

## Real-world impact

A scope was migrated from `scopeContainsTemplate` to a `#[Scope] whereContainsTemplate` method (rename). A caller iterating `class-string` model holders and calling `$fqcn::query()->containsTemplate($x)` kept compiling clean under Psalm and shipped, throwing `Call to undefined method ...EloquentBuilder::containsTemplate()` in production. Concrete-model call sites of the same scope would have been caught.

## Versions

- psalm/plugin-laravel: dev-master
- vimeo/psalm: 7.0.0-beta19
- PHP: 8.4

alies-dev:
## Issue to Solve

`$class::query()->method()` where `$class` is a `class-string<Model>` resolves to the base `Builder<Model>`. Because the `Builder` stub carries `@mixin \Illuminate\Database\Query\Builder` (and `Query\Builder` has `__call`), Psalm accepted any method name permissively, so an undefined or renamed scope passed analysis and only failed at runtime with `BadMethodCallException`. PHPStan/Larastan already catch this; Psalm did not.

## Related

Closes #1070

## Solution Description

Adds an always-on `AfterExpressionAnalysis` rule (`UndefinedBuilderMethodHandler`) that reports Psalm's **built-in `UndefinedMagicMethod`** when a method on a base `Builder<Model>` is none of: a real `Eloquent\Builder` / `Query\Builder` method, a registered macro, or a dynamic `where{Column}` form. This mirrors Laravel's `Builder::__call` / `Query\Builder::__call` dispatch, so it reports exactly the calls that throw `BadMethodCallException` at runtime — and it is the **same issue Psalm already raises for the concrete custom-builder case** (`WorkOrder::query()->fake()`), so both paths surface identically and honor existing `UndefinedMagicMethod` configuration. No new issue type is introduced.

Reporting is deliberately restricted to a direct `$class::query()->method()` receiver. A bare `Builder $q` parameter (filters, pipelines, `whereHas(..., fn (Builder $q) => ...)` closures) is type-identical to `Builder<Model>` but usually backs a concrete model at runtime, so flagging it would be a false positive. Concrete-model builders (`Builder<Post>`) and models with custom builders are also out of scope (the latter already report undefined calls via the builder class). Trade-off: a variable-assigned builder (`$q = $class::query(); $q->...`) or a longer chain (`query()->where()->...`) is not reported.

Verified with a new type test (1 positive plus 6 negative cases: real builder/query/dynamic-where methods, a registered macro, a bare `Builder<Model>` parameter, a variable-assigned builder, and a concrete-model builder), the full type suite (458 tests), self-analysis at 100% type coverage, and the fresh-app integration test.

## Checklist
- [x] Tests cover the change (type test in `tests/Type/tests/Builder/UndefinedBuilderMethodTest.phpt`)

### DIFF
--- a/src/Handlers/Rules/UndefinedBuilderMethodHandler.php
+++ b/src/Handlers/Rules/UndefinedBuilderMethodHandler.php
@@ -1,0 +1,210 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Rules;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Query\Builder as QueryBuilder;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Identifier;
+use Psalm\Codebase;
+use Psalm\CodeLocation;
+use Psalm\Internal\MethodIdentifier;
+use Psalm\Issue\UndefinedMagicMethod;
+use Psalm\IssueBuffer;
+use Psalm\LaravelPlugin\Util\DynamicWhereResolver;
+use Psalm\Plugin\EventHandler\AfterExpressionAnalysisInterface;
+use Psalm\Plugin\EventHandler\Event\AfterExpressionAnalysisEvent;
+use Psalm\Type\Atomic\TGenericObject;
+use Psalm\Type\Atomic\TNamedObject;
+use Psalm\Type\Union;
+
+/**
+ * Flags undefined method/scope calls on `$class::query()` when `$class` is a bare
+ * `class-string<\Illuminate\Database\Eloquent\Model>`.
+ *
+ * Such a `query()` resolves to the base `Builder<Model>`, whose stub
+ * `@mixin \Illuminate\Database\Query\Builder` (carrying `__call`) makes Psalm accept ANY
+ * method name permissively — so a genuinely-undefined scope/method is silently swallowed
+ * and surfaces only as a downstream `mixed`. At runtime the same call reaches
+ * `Builder::__call`, which forwards to `Query\Builder::__call`, finds no scope/macro/
+ * where-clause to route to (the base `Model` declares no scopes), and throws
+ * `BadMethodCallException`. This was the motivating bug in #1070: a renamed scope reached
+ * production via `$fqcn::query()->oldScope()`.
+ *
+ * The handler reports Psalm's built-in {@see \Psalm\Issue\UndefinedMagicMethod} — the same
+ * issue Psalm already raises for the concrete custom-builder case
+ * (`WorkOrder::query()->fake()`) — so both paths surface identically and honor existing
+ * `UndefinedMagicMethod` issue configuration.
+ *
+ * **Why the receiver is restricted to a direct `::query()` call.** A bare `Builder $q`
+ * parameter (idiomatic in filters, pipelines, and `whereHas(..., fn (Builder $q) => ...)`
+ * closures) is *type-identical* to `Builder<Model>` — Psalm fills the missing template
+ * param with the `Model` bound and there is no type-level discriminator. Such a variable
+ * usually stands in for a concrete model at runtime, where the scope is perfectly valid, so
+ * flagging it would be a false positive. Requiring the receiver to be `$class::query()`
+ * narrows reporting to the genuine `class-string<Model>` case from #1070 and leaves bare
+ * `Builder` receivers alone. The trade-off: a variable-assigned builder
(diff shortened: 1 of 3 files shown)


alies-dev:
## psalm-delta: base `3f7d0f49` → PR `e9215d01`

Ran the plugin on 18 real-world Laravel apps on both refs (base = current `master`, head = this PR), reporting deltas only.

### Per-app delta

| aureuserp | 0 | 0 | +0 | +0.00 | +0.99s |
| bagisto | 0 | 0 | +0 | +0.00 | +0.38s |
| bookstack | 0 | 0 | +0 | +0.00 | +0.73s |
| coolify | 0 | 0 | +0 | -0.00 | +0.49s |
| corcel | 0 | 0 | +0 | +0.00 | -0.05s |
| monica | 0 | 0 | +0 | +0.00 | +0.10s |
| pixelfed | 0 | 0 | +0 | +0.00 | +0.52s |
| solidtime | 0 | 0 | +0 | +0.00 | +0.38s |
| spatie-dashboard | 0 | 0 | +0 | +0.00 | +0.30s |
| tastyigniter | 0 | 0 | +0 | +0.00 | +0.35s |
| unit3d | 0 | 0 | +0 | +0.00 | -0.13s |
| vito | 0 | 0 | +0 | +0.00 | -0.05s |
| laravel-excel | 0 | 0 | +0 | +0.00 | -0.51s |
| filament | 0 | 2 | -2 | +0.00 | -0.31s |
| laravel-backup | 0 | 0 | +0 | +0.00 | -0.04s |
| laravel-permission | 0 | 0 | +0 | +0.00 | +0.19s |
| laravel-socialite | 0 | 0 | +0 | +0.00 | +0.06s |
| private-project-1 | 0 | 0 | +0 | +0.00 | -1.68s |
| **Total / Weighted** | **0** | **2** | **-2** | **+0.00** | **+1.72s** |

### Issue type movements

| MixedArgument | 0 | 1 | -1 |
| MixedArgumentTypeCoercion | 0 | 1 | -1 |

### Interpretation

- **Zero issues introduced** across all 18 apps. The provenance-restricted rule (only `$class::query()->method()` on a bare `class-string<Model>`) does not fire on these codebases, so there is no false-positive regression from making it always-on.
- The `filament` `-2` is known multi-thread flakiness on `HasBulkActions.php` (this app flaps ±2 on the base ref alone). It is unrelated to this PR: the rule is purely additive and only emits `UndefinedMagicMethod`, so it cannot remove `MixedArgument` / `MixedArgumentTypeCoercion`.
- Type coverage unchanged; runtime delta is within noise (+1.72s total across 18 apps).

Issue #750, opened 2026-04-17
`app(static::class, ...)` returns mixed — ContainerResolver misses template class-strings


alies-dev:
## Description

`app(static::class, [...])` and `resolve(static::class, [...])` are common Laravel patterns for constructing instances through the container with constructor arguments. The plugin currently returns `mixed` for these calls, which cascades into a large number of downstream `MixedAssignment`, `MixedMethodCall`, and `MixedArgument` errors.

## Root cause

[`ContainerResolver::resolvePsalmTypeFromApplicationContainerViaArgs()`](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Util/ContainerResolver.php#L55-L85) only handles the case where the first argument is a literal string:

```php
if ($firstArgType && $firstArgType->isSingleStringLiteral()) {
    $abstract = $firstArgType->getSingleStringLiteral()->value;
    // ...
}
return null;
```

For `app(static::class, ...)`, the arg type is `class-string<static>` (a template-bound `TClassString`), not a literal string, so the branch is skipped. `ContainerHandler::getFunctionReturnType()` then falls back to `Type::getMixed()`.

## Reproduction

Audited [filamentphp/filament](https://github.com/filamentphp/filament) with the plugin (v4.8.2, Psalm 7 beta, `errorLevel=1`).

Representative snippets from `packages/actions/src/Action.php`:

```php
$static = app(static::class, ['label' => $label]);  // MixedAssignment
$static->configure();                                // MixedMethodCall
```

```php
$static = app(static::class, ['actions' => $actions]);  // MixedAssignment
$static->configure();                                    // MixedMethodCall
```

## Impact

In Filament alone, this pattern accounts for roughly 141 cascading errors (86 `MixedAssignment` + 55 `MixedMethodCall` on `\$static->configure()`), before counting further `MixedArgument` propagation at the next call site.

The pattern is common across the Laravel ecosystem whenever a class uses a static builder/factory method with container resolution, which is idiomatic in Filament, Livewire, Nova, and many Spatie packages.

## Expected behavior

When the first argument is a `TClassString` with a single class constraint (literal class-string or `class-string<T>` where `T` is known), return a `TNamedObject` of that class.

## Proposed fix

[...]

PR #1076, opened 2026-06-16
Infer `app(Foo::class)` as `Foo` when the booted app cannot resolve it

### LINKED ISSUE #757: `app(Foo::class)` / `resolve(Foo::class)` returns mixed when `Foo` is not bound in the Testbench app
## Description

When the first argument to `app()` / `resolve()` / `make()` is a literal class-string referring to a class the plugin's internal Testbench app does not know how to instantiate, the plugin falls back to `mixed`. This cascades into many downstream `MixedAssignment`, `MixedPropertyFetch`, `MixedMethodCall`, and `MixedArgument` errors.

Note: this is distinct from #750 (`static::class`, now handled in [`resolveFromClassString`](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Util/ContainerResolver.php#L115-L134)). This issue is about the literal-string branch in [`resolveFromLiteralString`](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Util/ContainerResolver.php#L84-L103).

## Root cause

[`ContainerResolver::resolveFromLiteralString()`](https://github.com/psalm/psalm-plugin-laravel/blob/master/src/Util/ContainerResolver.php#L84-L103) only returns a type if it can successfully `make()` the abstract through the plugin's booted Laravel app:

```php
private static function resolveFromLiteralString(string $abstract): ?Union
{
    $concrete = self::resolveFromApplicationContainer($abstract);

    if ($concrete === null) {
        return null;   // -> Type::getMixed()
    }
    // ...
}
```

The plugin's booted app is Orchestra Testbench, which does not register the analysed package's service providers. Any class that requires a binding, a non-public constructor, or provider-supplied dependencies will throw in `make()`:

```php
// src/Util/ContainerResolver.php:34-39
try {
    $concrete = ApplicationProvider::getApp()->make($abstract);
    // ...
} catch (\Throwable) {
    return null;
}
```

When that happens, the literal `Foo::class` argument is dropped and the caller sees `mixed`, even though `Foo` is a real, loadable class.

## Reproduction

Audited [spatie/laravel-backup](https://github.com/spatie/laravel-backup) with the plugin (v4.8.2, Psalm 7 beta, `errorLevel=1`).

`src/Tasks/Backup/Zip.php`:

```php
class Zip
{
    protected Config $config;

    public function __construct(protected string $pathToZip)
    {
        $this->zipFile = new ZipArchive;
        $this->config = app(Config::class);   // MixedAssignment
        // ...
    }

    public static function createForManifest(Manifest $manifest, string $pathToZip): self
    {
        $config = app(Config::class);         // MixedAssignment

        $relativePath = $config->backup->source->files->relativePath   // 4x MixedPropertyFetch
            ? rtrim($config->backup->source->files->relativePath, DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR
            : false;
        // ...
    }
}
```

`Spatie\Backup\Config\Config` has a `protected __construct(...)` and no service-provider binding inside the plugin's Testbench app, so `ApplicationProvider::getApp()->make(Config::class)` throws and the plugin returns `mixed`.

Similar sites in the same package: `src/Notifications/BaseNotification.php:23`, `src/Notifications/Notifiable.php:40`. A single `$config = app(Config::class)` on line 28 of `Zip.php` cascades into ~20 downstream errors in that file alone.

## Proposed fix

When container resolution fails but the abstract is itself a loadable class name, fall back to treating `app($abstract)` as returning an instance of that class. This mirrors Laravel's runtime behaviour for unbound classes (the container instantiates them directly).

```php
private static function resolveFromLiteralString(string $abstract): ?Union
{
    $concrete = self::resolveFromApplicationContainer($abstract);

    if ($concrete !== null) {
        return class_exists($concrete)
            ? new Union([new TNamedObject($concrete)])
            : new Union([TLiteralString::make($concrete)]);
    }

    // Container resolution failed (class unbound, protected constructor,
    // missing provider dependencies, etc.). If the abstract is itself a real
    // class, Laravel's runtime returns an instance of it — so do the same.
    if (class_exists($abstract)) {
        return new Union([new TNamedObject($abstract)]);
    }

    return null;
}
```

This is symmetrical with the existing `resolveFromClassString` path (#750) which already returns `TNamedObject` for `class-string<Foo>` without going through the container.

## Impact

In `spatie/laravel-backup`: 5 direct occurrences, but roughly 25 cascading errors (most of `src/Tasks/Backup/Zip.php`). The pattern `app(SelfOwnedConfigClass::class)` is common in Laravel packages that carry their own configuration or service objects.

alies-dev:
## Issue to Solve

`app(Foo::class)` / `resolve(Foo::class)` / `make(Foo::class)` degraded to `mixed` whenever the plugin's booted app could not `make()` the abstract — e.g. a class with a non-public constructor or unresolvable constructor dependencies that no registered provider binds. A single `$x = app(Foo::class)` then cascaded into `MixedAssignment`, `MixedPropertyFetch`, `MixedArgument`, and `MixedReturnStatement` across every downstream use.

This is the literal class-string branch of `ContainerResolver`. Distinct from #750 (`static::class`, the `class-string<Foo>` atomic path), which already returns a `TNamedObject` without touching the container.

## Related

Fixes #757. Subset of umbrella #766.

## Solution Description

When container resolution fails but the abstract is itself a loadable class, fall back to `new Union([new TNamedObject($abstract)])`, mirroring Laravel's runtime: in a real app the owning provider is loaded, so `app(Foo::class)` returns a `Foo`. This is symmetrical with `resolveFromClassString()` (#750).

`class_exists()` is the load-bearing guard, and it is both safe and sufficient:

- `Container::build()` only throws "not instantiable" / "unresolvable dependency" **after** `new ReflectionClass($abstract)` has already succeeded, so the class is loaded and `class_exists()` is true exactly in the failure case we target.
- Interfaces are excluded (`class_exists` returns false), so an unresolvable contract stays `mixed` rather than being wrongly claimed to resolve to itself.
- The fallback only ever returns the abstract itself, a **supertype** of whatever the runtime would build, so it cannot introduce a false positive on a member that genuinely exists.

### Verification

- New type test `tests/Type/tests/ContainerUnboundClassTest.phpt` exercises `app()` / `resolve()` / `make()` against `Illuminate\Validation\Rules\Enum` (autoloadable, unbound, scalar-ctor — `make()` throws).
- The taint fixture `SafeRedisEvalHardcoded.phpt` previously needed a `@var PhpRedisConnection` hint on `app(PhpRedisConnection::class)`; the plugin now infers that type natively, so the obsolete hint is removed.
- Full type suite (474), unit suite, and self-analysis (100% coverage) all green.
- `psalm-delta` across 18 real-world apps (master vs this branch): **zero new issues introduced on every app**, **58 issues removed** (filament −54, coolify −4), all in the Mixed family (39 `MixedArgument`, 7 `MixedReturnStatement`, 4 `MixedReturnTypeCoercion`, 3 `MixedAssignment`, etc.). No regressions; slightly faster overall.

### Note on the original repro

The issue's `spatie/laravel-backup` `app(Config::class)` example is already resolved on current `master` by #942 (`ApplicationProvider::registerDiscoveredVendorProviders()` registers the analysed package's own provider, so the `Config` binding exists). This change is a **complementary** net for the cases provider registration cannot cover: self-owned concrete classes whose binding silently fails during analysis, or classes that are simply unbound.

## Checklist

### DIFF
--- a/src/Util/ContainerResolver.php
+++ b/src/Util/ContainerResolver.php
@@ -98,6 +98,33 @@ final class ContainerResolver
         $concrete = self::resolveFromApplicationContainer($abstract);
 
         if ($concrete === null) {
+            // Container resolution failed: either the abstract is unbound, or the
+            // plugin's booted app (Orchestra Testbench, when analysing a Laravel
+            // *package* rather than an app) lacks the provider that would register
+            // the binding, and the concrete class is not auto-wireable via reflection
+            // (protected/private constructor or provider-supplied dependencies).
+            // See #757 / umbrella #766.
+            //
+            // If the abstract is itself a loadable class name, mirror Laravel's
+            // runtime: in a real app the owning provider IS loaded, so
+            // `app(Foo::class)` returns a `Foo`. Returning the named object here is
+            // symmetrical with resolveFromClassString() (#750), which already returns
+            // a TNamedObject for `class-string<Foo>` without touching the container.
+            //
+            // `class_exists()` is both safe and sufficient as the guard: the typical
+            // failure (Container::build throwing "not instantiable" / "unresolvable
+            // dependency") only happens AFTER `new ReflectionClass($abstract)` has
+            // already succeeded, so the class is loaded and `class_exists()` is true.
+            // It also correctly excludes interfaces (returns false → mixed) — we never
+            // claim an unresolvable contract resolves to itself. We only ever return
+            // the abstract itself, a supertype of whatever the runtime would build, so
+            // this cannot introduce a false-positive on a member that genuinely exists.
+            if (\class_exists($abstract)) {
+                return new Union([
+                    new TNamedObject($abstract),
+                ]);
+            }
+
             return null;
         }
 
--- a/tests/Type/tests/ContainerUnboundClassTest.phpt
+++ b/tests/Type/tests/ContainerUnboundClassTest.phpt
@@ -1,0 +1,38 @@
+--FILE--
+<?php declare(strict_types=1);
+
+// Regression for #757 (subset of umbrella #766): `app(Foo::class)` /
+// `resolve(Foo::class)` / `make(Foo::class)` previously returned `mixed` whenever
+// the plugin's booted app could not `make()` the abstract — e.g. the class is
+// unbound and its constructor has unresolvable (scalar / provider-supplied)
+// dependencies, or a non-public constructor. The literal class-string branch now
+// falls back to a named object, mirroring Laravel's runtime: in a real app the
+// owning provider is loaded, so `app(Foo::class)` returns a Foo.
+//
+// NB: the fixture must be a *real autoloadable* class. `ContainerResolver` uses
+// PHP's `class_exists()` (the analysed project's autoloader is registered with
+// Psalm), so the fallback only fires for classes loadable in the Psalm process —
+// which every genuine package class (the issue's Spatie\Backup\Config\Config) is.
+// `Illuminate\Validation\Rules\Enum` is unbound in the booted app and its
+// constructor requires a scalar `$type`, so `make()` throws BindingResolutionException.
+
+use Illuminate\Validation\Rules\Enum;
+
+function appHelperResolvesUnboundClass(): Enum
+{
+    $rule = app(Enum::class);
+    /** @psalm-check-type-exact $rule = Illuminate\Validation\Rules\Enum */
+    return $rule;
+}
+
+function resolveHelperResolvesUnboundClass(): Enum
+{
+    return resolve(Enum::class);
+}
+
+function containerMakeResolvesUnboundClass(): Enum
+{
+    return app()->make(Enum::class);
+}
+?>
+--EXPECTF--
--- a/tests/Type/tests/TaintAnalysis/SafeRedisEvalHardcoded.phpt
+++ b/tests/Type/tests/TaintAnalysis/SafeRedisEvalHardcoded.phpt
@@ -4,7 +4,9 @@
 <?php declare(strict_types=1);
 
 function runHardcodedScript(): void {
-    /** @var \Illuminate\Redis\Connections\PhpRedisConnection $redis */
+    // The plugin infers `app(Foo::class)` as Foo even when the booted app cannot
+    // build it (PhpRedisConnection needs a $client ctor arg, unresolvable here),
+    // so the previously-required `@var` hint is now redundant (#757).
     $redis = app(\Illuminate\Redis\Connections\PhpRedisConnection::class);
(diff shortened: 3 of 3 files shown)


alies-dev:
## psalm-delta — master (`1f38abbe`) → `ed8c7733`

18 real-world Laravel apps, Psalm + plugin. Delta only.

| aureuserp | 0 | 0 | +0 | +0.00 | +0.98s |
| bagisto | 0 | 0 | +0 | +0.00 | +0.00s |
| bookstack | 0 | 0 | +0 | +0.00 | +0.63s |
| coolify | 0 | 4 | -4 | +0.00 | -0.18s |
| corcel | 0 | 0 | +0 | +0.00 | -0.21s |
| monica | 0 | 0 | +0 | +0.00 | -1.61s |
| pixelfed | 0 | 0 | +0 | +0.00 | -0.55s |
| solidtime | 0 | 0 | +0 | +0.00 | -1.73s |
| spatie-dashboard | 0 | 0 | +0 | +0.00 | -1.32s |
| tastyigniter | 0 | 0 | +0 | +0.00 | -0.32s |
| unit3d | 0 | 0 | +0 | +0.00 | -0.45s |
| vito | 0 | 0 | +0 | +0.00 | -0.07s |
| laravel-excel | 0 | 0 | +0 | +0.00 | +0.04s |
| filament | 0 | 54 | -54 | +0.07 | -0.45s |
| laravel-backup | 0 | 0 | +0 | +0.00 | -0.18s |
| laravel-permission | 0 | 0 | +0 | +0.00 | -0.37s |
| laravel-socialite | 0 | 0 | +0 | +0.00 | -0.62s |
| private-project-1 | 0 | 0 | +0 | +0.00 | -5.67s |
| **Total / Weighted** | **0** | **58** | **-58** | **+0.00** | **-12.08s** |

_ΔCov weighted by base `wall_seconds` (codebase-size proxy)._

### Issue type movements

| MixedArgument | 0 | 39 | -39 |
| MixedReturnStatement | 0 | 7 | -7 |
| MixedReturnTypeCoercion | 0 | 4 | -4 |
| MixedAssignment | 0 | 3 | -3 |
| MixedFunctionCall | 0 | 2 | -2 |
| MixedArgumentTypeCoercion | 0 | 1 | -1 |
| MixedMethodCall | 0 | 1 | -1 |
| MixedPropertyAssignment | 0 | 1 | -1 |

**Zero new issues on every app; 58 removed, all Mixed-family.** No regressions, slightly faster overall.

_Base `1f38abbe` is 2 test-only commits ahead of the fork point; neither touches plugin runtime, so the delta isolates this PR's behavioral impact._

Issue #1113, opened 2026-06-20
`SessionGuard.phpstub` strips all methods (user/attempt/logout/...), causing `UndefinedMethod` on `auth('web')->user()`


alies-dev:
**Describe the bug**

`stubs/common/Auth/SessionGuard.phpstub` is a *full class stub* that replaces the real `Illuminate\Auth\SessionGuard`, but it only declares a single method (`hashPasswordForCookie()`):

```php
// stubs/common/Auth/SessionGuard.phpstub
namespace Illuminate\Auth;

class SessionGuard implements StatefulGuard, SupportsBasicAuth
{
    public function hashPasswordForCookie($passwordHash) {}
}
```

Because a class phpstub *replaces* the reflected class, every other member of the real `SessionGuard` is dropped:

- `use GuardHelpers, Macroable;`
- every concrete method: `user()`, `id()`, `check()`/`guest()` (via `GuardHelpers`), `attempt()`, `login()`, `logout()`, `once()`, `validate()`, `loginUsingId()`, `attempting()`, `getLastAttempted()`, etc.

The stub is a concrete class that `implements StatefulGuard, SupportsBasicAuth` but does not define the interface methods, so any call to a stripped method on a value narrowed to `SessionGuard` reports `UndefinedMethod`. The plugin's auth handlers narrow `auth('web')` / `Auth::guard('web')` to `SessionGuard`, so this hits normal user code.

This is adjacent to (but distinct from) #979: that one was about the `Guard` contract lacking `logout()`; this one is the stub itself shadowing the whole concrete class.

**Code to reproduce**

```php
<?php
// session-driver guard ('web'), narrowed to SessionGuard by the plugin
return (bool) auth('web')->user()?->getAuthIdentifier();
// ERROR: UndefinedMethod - Method Illuminate\Auth\SessionGuard::user does not exist
```

Confirmation that the stub is the cause: adding `public function user() {}` to `SessionGuard.phpstub` makes the `UndefinedMethod` disappear (the call then resolves and only the downstream provider-model type remains). The same applies to `attempt()`, `logout()`, `id()`, and the other stripped methods.

**Expected**

`auth('web')->user()` (and the other `SessionGuard` methods) resolve, as they did before the class-level stub shadowed the real class.

**Impacted Versions**

```text
laravel/framework      13.16.1
psalm/plugin-laravel   4.14.0
vimeo/psalm            7.0.0-beta19
```
PHP 8.4.22.

**Additional context**

[...]

PR #1128, opened 2026-06-21
Fix false `UndefinedMethod` on `app('encrypter')` methods

### LINKED ISSUE #1126: Encrypter.phpstub strips all methods (getKey/getAllKeys/getPreviousKeys), causing UndefinedMethod on app('encrypter')->getKey()
**Describe the bug**

`stubs/common/Encryption/Encrypter.phpstub` is a *full class stub* that replaces the real `Illuminate\Encryption\Encrypter`, but it only declares the four taint-carrying methods (`encrypt`, `encryptString`, `decrypt`, `decryptString`):

```php
// stubs/common/Encryption/Encrypter.phpstub
namespace Illuminate\Encryption;

class Encrypter implements EncrypterContract, StringEncrypter
{
    public function encrypt($value, $serialize = true) {}
    public function encryptString($value) {}
    public function decrypt($payload, $unserialize = true) {}
    public function decryptString($payload) {}
}
```

Because a class phpstub *replaces* the reflected class, every other member of the real `Encrypter` is dropped, including `getKey()`, `getAllKeys()`, and `getPreviousKeys()`.

The plugin's container resolver narrows `app('encrypter')` / `resolve('encrypter')` / `make('encrypter')` to a bare `Encrypter` object **without naming the class**, so Psalm never scans the real vendor source and the stub becomes its sole definition. Unlike most Laravel service classes (`Cache\Repository`, `Session\Store`, `Database\Connection`, ...) the encrypter carries no `Macroable`/`__call` to mask the stripped methods, so the strip surfaces as a hard `UndefinedMethod`.

This is the same trap fixed for the auth guards in #1113, applied to a different class.

**Code to reproduce**

```php
<?php
// encrypter reached via container narrowing, never naming Illuminate\Encryption\Encrypter
app('encrypter')->getKey();
// ERROR: UndefinedMethod - Method Illuminate\Encryption\Encrypter::getKey does not exist
```

Confirmation that the stub is the cause: deleting `Encrypter.phpstub` makes the `UndefinedMethod` disappear (the real class is then scanned and all methods resolve). The same applies to `getAllKeys()` and `getPreviousKeys()`.

**Expected**

`app('encrypter')->getKey()` (and the other real `Encrypter` methods) resolve, while the encrypt/decrypt taint contract (escape secrets on encrypt, restore on decrypt) is preserved.

**Impacted Versions**

```text
laravel/framework      13.16.1
vimeo/psalm            7.0.0-beta19
```
PHP 8.4.22.

**Additional context**

Found while auditing all class-redeclaring stubs in `stubs/common` for the #1113 narrowing-shadow trap. Of 83 partial class-redeclaration stubs, `Encrypter` is the only one (besides the already-fixed guards) that is both reached provider-narrowed-without-naming and lacks a `Macroable`/`__call` to mask the strip. The stub exists only to attach taint annotations to the four crypt methods; a full class redeclaration is too broad for that. Fix: move the taint to a scan-phase handler (mirroring `GuardTaintHandler`) and delete the stub.

alies-dev:
## Issue to Solve

`stubs/common/Encryption/Encrypter.phpstub` was a full-class redeclaration that hosted only the four taint methods (`encrypt`, `encryptString`, `decrypt`, `decryptString`). A class `.phpstub` *replaces* the reflected class, so every other method of the real `Illuminate\Encryption\Encrypter` (`getKey`, `getAllKeys`, `getPreviousKeys`) was stripped.

The plugin's container resolver narrows `app('encrypter')` / `resolve('encrypter')` / `make('encrypter')` to a bare `Encrypter` **without naming the class**, so Psalm never scans the real vendor source and the stub becomes its sole definition. Unlike most Laravel service classes, `Encrypter` carries no `Macroable`/`__call` to mask the strip, so `app('encrypter')->getKey()` surfaced as a hard `UndefinedMethod`. Same trap fixed for the auth guards in #1113.

## Related

Closes #1126. Sibling of #1113 (same narrowing-shadow trap, applied to the guards).

## Solution Description

Move the taint off the stub and onto the real method storage, then delete the stub (mirrors `GuardTaintHandler` from #1113):

- New `EncrypterTaintHandler` (`AfterClassLikeVisitInterface`) sets, on the real `Encrypter` method storage, exactly the fields the deleted docblocks populated: `removed_taints` (escape, on `encrypt`/`encryptString`), `added_taints` (unescape, on `decrypt`/`decryptString`), and `return_source_params` (the `@psalm-flow ($value) -> return`). The instance-call taint path reads these back, so the escape/unescape/flow contract is reproduced verbatim.
- The real class is left intact, so `getKey()` and the other stripped methods resolve.

Verified:

- New subprocess regression test `EncrypterMethodResolutionTest` reaches `Encrypter` only via `app('encrypter')` and asserts no `UndefinedMethod` (and no `MixedMethodCall`, so a future narrowing regression back to `mixed` is also caught). Red without the fix (3 `UndefinedMethod`), green with it. A `.phpt` cannot reproduce this: single-file psalm-tester scans the narrowed class anyway (per #1113).
- The four existing `Safe*`/`Tainted*` Encrypter taint `.phpt`s still pass; the provider path (`app('encrypter')->decryptString($tainted)`) fires `TaintedHtml`/`TaintedUserSecret`/`TaintedSystemSecret`, and the escape suppresses the secret kinds.
- Full type and unit suites, `composer psalm` (100% coverage), `composer cs`, and `composer rector` are all clean.

### Audit

Found by auditing **all 83** class-redeclaration stubs in `stubs/common` for this trap. A stub leaks `UndefinedMethod` only when it is a partial redeclaration **and** reached provider-narrowed-without-naming **and** the class has no `Macroable`/`__call`/`@mixin` to mask the strip. Screen: `method_exists($fqcn, '__call')` (catches inherited `__call`) plus a `@mixin` check, yielding 17 candidates, each probed with an isolated `<projectFiles>` Psalm run.

`Encrypter` is the only at-risk stub (besides the already-fixed guards). Every other partial stub is safe:

| reason | examples |
|---|---|
| `Macroable`/`__call` masks the strip | `Cache\Repository`, `Session\Store`, `Database\Connection`, `Config\Repository`, `Filesystem\Filesystem`, `HashManager`, `Validator`, all `Relations\*` (`@mixin Builder`) |
| contract-typed receiver | `view()` to `Contracts\View\Factory`, `response()` to `Contracts\Routing\ResponseFactory` |
| merged via `@method`/subclass scan | `DB::connection()` to `Connection`, `Process::` to `PendingProcess`, `MessageBag`, `Migrator` |
| named directly | `Collection`, `Str`, `Stringable`, `HtmlString`, `Js`, the facades, `Attribute` |
| complete redeclaration | `ControllerMiddlewareOptions` |

## Checklist
- [x] Tests cover the change (subprocess unit test + the existing Encrypter taint type tests)
- [x] Documentation is updated (`docs/contributing/taint-analysis.md`: the narrowing-shadow exception and the `@psalm-taint-specialize` follow-up note)


alies-dev:
Note for the master to 3.x merge (Psalm 6): `EncrypterTaintHandler` relies on two `MethodStorage` fields beyond what `GuardTaintHandler` used, `added_taints` (unescape) and `return_source_params` (flow). Their read-back semantics were verified against Psalm 7 (`vendor` here is 7-beta19). Both are long-standing core taint fields, so Psalm 6 should behave identically, and the four Encrypter taint `.phpt`s will confirm it when the 3.x matrix runs. Flagging it as the one field-level assumption not directly exercised on the Psalm 6 side in this PR.

PR #1130, opened 2026-06-21
Resolve concrete-only `Application` methods on the contract

### LINKED ISSUE #1108: `UndefinedInterfaceMethod` for concrete-only Application methods (isProduction/isLocal) on contract-typed properties
### Summary

Calling methods that exist on the concrete `Illuminate\Foundation\Application` but are **not declared on the `Illuminate\Contracts\Foundation\Application` contract** raises `UndefinedInterfaceMethod` (psalm 181) whenever the receiver is typed as the contract. The most common offenders are `isProduction()` and `isLocal()`.

The plugin already maps the contract to the concrete app for **container resolution / return types** (`ApplicationInterfaceProvider` + `ContainerHandler`), but that mechanism is a `MethodReturnTypeProviderInterface`. It only fires for methods that *already exist* on the type. Method-existence on the contract itself is never widened, so 181 is emitted before any return-type provider runs.

### Reproduce

```php
use Illuminate\Console\Command;

final class DatabasePrune extends Command
{
    public function handle(): void
    {
        // Command::$laravel is typed `@var \Illuminate\Contracts\Foundation\Application`
        if ($this->laravel->isProduction()) { // UndefinedInterfaceMethod
            // ...
        }
    }
}
```

```
Method Illuminate\Contracts\Foundation\Application::isProduction does not exist (see https://psalm.dev/181)
        if ($this->laravel->isProduction()) {
```

Same error fires for any contract-typed receiver, e.g. `ServiceProvider::$app` and `app()` when its return type is the contract.

### Root cause

`Illuminate\Contracts\Foundation\Application` does not declare these methods; only the concrete `Illuminate\Foundation\Application` does. Confirmed against `laravel/framework` v12.62.0. Concrete-only methods include:

`isLocal`, `isProduction`, `detectEnvironment`, `environmentFile`, `environmentFilePath`, `environmentPath`, `useEnvironmentPath`, `loadEnvironmentFrom`, `afterLoadingEnvironment`.

Of these, `isLocal()` / `isProduction()` are idiomatic and very common in commands and providers.

The plugin stubs the **concrete** app (`stubs/common/Foundation/Application.phpstub`) but there is no stub for the **contract**, and `ContainerHandler` (a method-return provider) cannot synthesize missing methods.

For comparison, Larastan (PHPStan) does not surface this because it resolves the app to the concrete type via `ApplicationResolver`, so users hit a Psalm-vs-PHPStan divergence on otherwise-valid code.

### Why this is the plugin's job, not the user's

There is exactly one runtime implementation of `Illuminate\Contracts\Foundation\Application` (`Illuminate\Foundation\Application`); no third-party implementor ships these days (Lumen is archived and never implemented this contract). So narrowing loses no real safety here, unlike multi-driver contracts (`Cache\Repository`, `Filesystem`, `Queue`) where it would hide real bugs. The fix should stay scoped to the Application/Container contracts, not generalize to all contracts.

### Proposed fix

Add a stub for `Illuminate\Contracts\Foundation\Application` declaring the concrete-only environment/state methods (at minimum `isLocal()` and `isProduction()`), or inject them onto the contract's class-like storage in `ContainerHandler::afterClassLikeVisit()` (which already runs on these class-likes). Either approach makes method existence pass; the existing return-type providers then continue to narrow as today.

### Environment

- psalm/plugin-laravel: dev-master
- vimeo/psalm: 7.0.0-beta19
- laravel/framework: v12.62.0

alies-dev:
## Issue to Solve

Calling methods that exist only on the concrete `Illuminate\Foundation\Application` (e.g. `isProduction()`, `isLocal()`, the `environment*()` family) on a receiver typed as the `Illuminate\Contracts\Foundation\Application` contract raised `UndefinedInterfaceMethod` (psalm 181). The contract declares ~30 methods but not these 9 environment/state ones. Common receivers: `Command::$laravel`, `ServiceProvider::$app`, and `app()` when widened to the contract.

## Related

Fixes #1108

## Solution Description

After the codebase is populated, an `AfterCodebasePopulated` handler (`ApplicationContractMethodHandler`) registers the concrete app's own method ids on the contract's `declaring_method_ids` / `appearing_method_ids` resolution maps for the 9 concrete-only methods. Naive `methodExists()` reads `declaring_method_ids` directly, so the call resolves on contract-typed receivers, and return/param lookups follow the id to the concrete app's real storage. Signatures are delegated to the live concrete app, so they stay correct across Laravel 12-13 without restating them here.

The entries are deliberately **not** written to `$contract->methods`: the interface-compliance check iterates `$interface->methods` at analysis time, so writing there would force every class that `implements` the contract to declare these 9 methods (spurious `UnimplementedInterfaceMethod`). Keeping them out of `methods` makes the resolution invisible to compliance while still resolving the call.

Why not the alternatives:

- A `MethodExistenceProviderInterface` only suppresses 181 when the receiver declares `__call` (the magic-method path is `__call`-gated). The contract has no `__call`.
- A contract `.phpstub` would have to restate every signature (drifts across versions), and re-declaring the interface risks resetting its parent list unless the full `extends` clause is copied verbatim.

Verification: a red→green type test exercises all 9 methods on `Command::$laravel`, `ServiceProvider::$app`, and a contract-typed parameter (pinning `bool` / `string` returns and chaining off a fluent setter to prove `@return $this` is not `mixed`). A separate no-leak test asserts a genuinely-undefined method still raises `UndefinedInterfaceMethod`, guarding the scope. Full type suite, unit, app integration, and self-analysis (100% coverage) are green; `cs` / `rector` are clean.

Out-of-scope follow-up: `Application::path()` is also concrete-only and hits the same 181, while its seven path-siblings (`basePath()`, `configPath()`, ...) are all on the contract. Left out of this env/state-scoped fix as a candidate for a separate change.

## Checklist
- [x] Tests cover the change (type tests in `tests/Type/`)

### DIFF
--- a/src/Handlers/Application/ApplicationContractMethodHandler.php
+++ b/src/Handlers/Application/ApplicationContractMethodHandler.php
@@ -1,0 +1,130 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Application;
+
+use Illuminate\Contracts\Foundation\Application as ApplicationContract;
+use Psalm\Codebase;
+use Psalm\LaravelPlugin\Providers\ApplicationProvider;
+use Psalm\Plugin\EventHandler\AfterCodebasePopulatedInterface;
+use Psalm\Plugin\EventHandler\Event\AfterCodebasePopulatedEvent;
+use Psalm\Storage\ClassLikeStorage;
+
+/**
+ * Declares the environment/state methods that exist on the concrete
+ * {@see \Illuminate\Foundation\Application} but NOT on its
+ * {@see \Illuminate\Contracts\Foundation\Application} contract, so calling them
+ * on a contract-typed receiver (`Command::$laravel`, `ServiceProvider::$app`, or
+ * any property/parameter typed on the interface) no longer raises
+ * `UndefinedInterfaceMethod` (psalm 181).
+ *
+ * **Why a resolution-map write and not a `.phpstub`.** Stubbing the contract to
+ * add these methods would mean restating every signature in the stub (which drifts
+ * as Laravel evolves) and re-declaring the interface — which, unless the full
+ * `extends` clause is copied verbatim, resets the interface's parent list (see the
+ * "Class declarations wipe reflected metadata" note in the contributing docs).
+ * Pointing the contract's resolution maps at the live concrete app keeps signatures
+ * version-correct and leaves the reflected interface untouched.
+ *
+ * **Why not a method-existence provider.** `MethodExistenceProviderInterface` only
+ * suppresses 181 when the receiver declares `__call` (the analyzer routes a
+ * provider-confirmed-but-storage-less method through the magic-method handler,
+ * which is `__call`-gated — see `AtomicMethodCallAnalyzer`). The contract has no
+ * `__call`, so the method must exist for real in `declaring_method_ids`; only a
+ * resolution-map entry achieves that.
+ *
+ * **Why only the resolution maps, not `$contract->methods`.** Naive
+ * `methodExists()` consults `declaring_method_ids` directly, so adding entries
+ * there (pointed at the concrete app's own declarations) is enough to resolve the
+ * call. Deliberately NOT touching `$contract->methods`: the interface-compliance
+ * check iterates `$interface->methods` at analysis time, so writing there would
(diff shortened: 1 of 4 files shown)


alies-dev:
closed in favor of https://github.com/psalm/psalm-plugin-laravel/pull/1141

PR #1209, opened 2026-07-02
[v4.15] Add `<experimental>` config element to gate unstable rules and handlers

### LINKED ISSUE #1203: Add `<experimental>` config element to gate unstable rules and handlers
## Summary

Add an `<experimental>` config element that gates not-yet-stable rules and handlers behind an explicit opt-in. Only users who enable a feature by name (or enable everything) run it. This gives new rules a real-world testing period without exposing every user to churn, and gives us a documented lifecycle for promoting or withdrawing features.

Prior art: PHPStan's `bleedingEdge.neon` (single all-or-nothing switch backed by internal named feature toggles that flip to default on the next major). We combine both ideas: named per-feature opt-in plus an "all" switch.

**Base branch: `4.15`** (not `master`). The pilot feature (`ModelToArrayShapeHandler`, merged there via #1168) exists only on `4.15`.

## Configuration design

New XML element inside `<pluginClass>` in `psalm.xml`:

```xml
<!-- granular opt-in -->
<plugins>
        <experimental>
            <feature name="modelToArrayShape" />
        </experimental>
    </pluginClass>
</plugins>

<!-- or: enable every experimental feature (early-adopter / bleeding-edge mode) -->
<experimental all="true" />
```

Semantics:

| Input | Behavior |
|---|---|
| element absent | all experiments off (pure opt-in, no tri-state) |
| `<feature name="...">` with a known name | that feature enabled |
| `all="true"` | every feature in the enum enabled; `<feature>` children redundant but harmless |
| unknown `name` | throw `InvalidArgumentException` listing all valid names and linking to `docs/config.md` (matches the existing `columnFallback` typo-strictness in `PluginConfig::fromXml()`) |
| name of a graduated feature | `trigger_error(..., E_USER_DEPRECATED)` saying it is now stable and the element should be removed, then ignore. Never a hard error: a plugin upgrade must not break existing psalm runs. Precedent: the deprecated env-var path in `PluginConfig::resolveCachePath()` |
| name of a withdrawn feature | same mechanism, message says the feature was withdrawn, then ignore |
| `<experimental />` present but empty and no `all` attribute | deprecation-style notice that the element has no effect (user intent unclear, likely a mistake) |
| duplicate `<feature>` entries | dedupe silently |

### 1. `src/Config/ExperimentalFeature.php` (new)

String-backed enum, the single registry of live experiments. Each case carries a docblock: one-line description plus tracking issue/PR link.

```php
<?php

declare(strict_types=1);

namespace Psalm\LaravelPlugin\Config;

/**
 * Registry of experimental features that must be explicitly enabled via
 * the `<experimental>` element in psalm.xml.
 *
 * Lifecycle (see docs/contributing/README.md):
 *  - introduce: add a case here, default off
 *  - stabilize: remove the case, add its value to GRADUATED
 *  - withdraw:  remove the case, add its value to WITHDRAWN
 *
 * @internal
 */
enum ExperimentalFeature: string
{
    /** Infer array shapes for Model::toArray()/attributesToArray() (#923, PR #1168). */
    case ModelToArrayShape = 'modelToArrayShape';

    /**
     * Feature names that used to be experimental and are now stable/always-on.
     * Requesting one produces a deprecation notice, not an error.
     * Value = plugin version where it graduated.
     *
     * @var array<non-empty-string, non-empty-string>
     */
    public const GRADUATED = [];

    /**
     * Feature names that were withdrawn without stabilizing.
     * Requesting one produces a deprecation-style notice, not an error.
     * Value = short reason.
     *
     * @var array<non-empty-string, non-empty-string>
     */
    public const WITHDRAWN = [];
}
```

(If Psalm complains about constants on enums in this codebase style, move the two maps to a small `ExperimentalFeatureLifecycle` final class beside the enum. Prefer the enum-constant form first.)

### 2. `src/Config/PluginConfig.php`

This is a `final readonly` value object built once in `PluginConfig::fromXml(?\SimpleXMLElement $config)` and threaded into handlers. Existing parsing helpers to imitate: `xmlBoolAttr()`, `xmlOptionalBoolAttr()`, `xmlNameList()` (which already demonstrates the `iterable<\SimpleXMLElement>` iteration workaround for SimpleXMLElement dynamic properties and the throw-on-bad-attribute convention).

Add:

- constructor params + public fields:
  - `public bool $experimentalAll`
  - `/** @param list<ExperimentalFeature> $experimentalFeatures */ public array $experimentalFeatures`
- a private parser, roughly:

```php
/**
 * Parse <experimental all="true"> / <experimental><feature name="..."/></experimental>.
 *
 * @return array{all: bool, features: list<ExperimentalFeature>}
 */
private static function xmlExperimentalFeatures(?\SimpleXMLElement $config): array
```

  Parse order per `<feature name>` value: `GRADUATED`? → deprecation notice, skip. `WITHDRAWN`? → notice, skip. `ExperimentalFeature::tryFrom()` → null? throw with the full valid-name list (from `ExperimentalFeature::cases()`) and a pointer to `docs/config.md`. Bonus DX: include a nearest-match suggestion via `levenshtein()` when a close name exists.
  The `all` attribute reuses the `'true'|'false'` strict validation style of `xmlBoolAttr()` (reject anything else with `InvalidArgumentException`).
  Note the SimpleXMLElement gotcha documented in `xmlOptionalBoolAttr()`: accessing a missing child via dynamic property returns an empty proxy, so "element absent" checks need `isset()`-style handling; follow the existing helpers.

- a mutation-free helper:

```php
/** @psalm-mutation-free */
public function isExperimentEnabled(ExperimentalFeature $feature): bool
{
    return $this->experimentalAll || \in_array($feature, $this->experimentalFeatures, true);
}
```

### 3. `src/Plugin.php` — gating

`Plugin::registerHandlers()` gates conditional handlers like this (see the existing `reportImplicitQueryBuilderCalls` / `findMissingViews` blocks):

```php
if ($pluginConfig->isExperimentEnabled(ExperimentalFeature::ModelToArrayShape)) {
    require_once __DIR__ . '/Handlers/Eloquent/ModelToArrayShapeHandler.php';
    $registration->registerHooksFromClass(ModelToArrayShapeHandler::class);
}
```

**Critical repo rule: keep the paired `require_once` immediately before `registerHooksFromClass()`.** Every handler registration uses explicit `require_once` because the plugin must work from `psalm.phar` where the project's PSR-4 autoloader for `Psalm\LaravelPlugin\` may not be registered. Do not replace with `class_exists()` or autoload-only loading (see docs/contributing/README.md and issue #895).

Also emit one line per run when any experiment is active, via the `\Psalm\Progress\Progress` instance already available in `Plugin::__invoke()`:

```
Laravel plugin: experimental features enabled: modelToArrayShape
```

so CI logs record exactly which non-stable analysis ran.

### 4. Pilot feature: gate `ModelToArrayShapeHandler`

On `4.15`, `ModelToArrayShapeHandler` (Eloquent `toArray()`/`attributesToArray()` array-shape inference, merged via #1168) is registered unconditionally. Move its registration behind `ExperimentalFeature::ModelToArrayShape` as the first consumer. Update its type tests: tests that exercise the shape inference need the feature enabled in their psalm config fixture; add one test asserting the handler is inert (plain `array<...>` return, no shape) when the feature is off.

Future consumers (do NOT implement here, just so the shape is understood): the rules from PR #1167 (`UnknownModelAttribute`) and PR #1169 (`UnresolvableAppendedAttribute`) are expected to adopt the same gate in their own PRs once this lands.

### 5. `psalm-laravel diagnose`

`src/Cli/DiagnoseCommand.php` + `src/Cli/Diagnose/Diagnostics.php` / `Report.php`: add a section listing enabled experimental features (or "none, N available" with a pointer to `docs/config.md` when none are enabled). Follow the existing section structure in `Report`.

### 6. Documentation

- `docs/config.md`: new `## experimental` section following the existing per-option format (default, description, example). Must state the stability contract explicitly: experimental features may change behavior, change name, or be removed in ANY release; they are exempt from semver. Include a table of current features (name, one-line description, tracking issue) and both enable forms. Also add `<experimental>` to the "Full config example" block at the top of the file.
- `docs/contributing/README.md`: document the lifecycle — introduce (enum case, default off, docs entry) → stabilize (registration becomes unconditional or gets its own top-level flag; enum case removed; name added to `GRADUATED`; deprecation notice kept for at least one minor release, removed at the next major) → withdraw (case removed; name added to `WITHDRAWN`). Also note that this file's hook→handler tables must reflect the gated registration (repo convention: any `Plugin::registerHandlers()` change updates this doc in the same PR).
- Experimental rules' issue docs (here: none new; `ModelToArrayShape` has no issue doc since it emits no custom issue — skip) — but add the convention to the contributing doc: any experimental rule that emits a custom issue gets a top banner in `docs/issues/<Name>.md`: "Experimental — enable via `<experimental><feature name="..." /></experimental>`".

### 7. Tests

Extend `tests/Unit/PluginConfigTest.php` (existing patterns there show how `SimpleXMLElement` fixtures are built):

- absent element → `experimentalAll === false`, empty feature list, `isExperimentEnabled()` false for every case
- named parse → exactly the named features enabled, others off
- `all="true"` → `isExperimentEnabled()` true for every enum case
- `all="banana"` → `InvalidArgumentException`
- unknown feature name → `InvalidArgumentException`, message contains every valid name
- graduated name → deprecation triggered (PHPUnit: convert to exception or use `set_error_handler` capture, follow however the existing cache-path deprecation is tested, if it is), feature list unaffected
- withdrawn name → same
- empty `<experimental />` → notice, everything off
- duplicate `<feature>` entries → deduped

Gating test: the pilot's type tests (phpt via psalm-tester under `tests/Type/`) cover on/off behavior as described in step 4.

## Rejected alternatives (do not implement)

- One top-level element per experiment (`<findFooExperimental/>`): graduation forces a rename and breaks user configs.
- Env-var enable: XML is the single config surface; the last env knob (`PSALM_LARAVEL_PLUGIN_CACHE_PATH`) is already deprecated.
- Single boolean only (pure bleedingEdge): no way to trial one risky rule in isolation; `all="true"` already serves the early-adopter persona.

## Repo conventions the implementation must follow

- PHP 8.2+, PER Coding Style 3.0 — run `composer cs -- --show-progress=none --no-ansi -n` and `composer rector -- --no-progress-bar --no-ansi` before pushing (CI auto-commits style fixes otherwise).
- Self-analysis must stay at 100% type inference: `composer psalm -- --no-progress --stats` (no new mixed). No new `@psalm-suppress`, no baseline entries (there is no baseline in this repo).
- Test commands: `composer test:unit -- --no-progress`, `composer test:type -- --no-progress`, single files via `./vendor/bin/phpunit <path>`.
- Final pre-commit psalm gate must run WITHOUT `--no-suggestions` (CI fails on UnusedParam/MissingPureAnnotation/ClassMustBeFinal that the flag hides).
- Comments explain decisions for future contributors; concise engineering style.

## Acceptance criteria

- [ ] `ExperimentalFeature` enum + `GRADUATED`/`WITHDRAWN` maps exist with the `ModelToArrayShape` case
- [ ] `PluginConfig` parses both forms, validates strictly, exposes `isExperimentEnabled()`
- [ ] `ModelToArrayShapeHandler` registers only when enabled; `require_once` pairing preserved
- [ ] Progress line lists active experiments once per run
- [ ] `psalm-laravel diagnose` reports experimental feature status
- [ ] `docs/config.md` + `docs/contributing/README.md` updated (stability contract + lifecycle + hook→handler table)
- [ ] Unit tests for all parse/validation branches; type tests for pilot on/off
- [ ] `composer cs`, `composer rector`, full psalm self-analysis (100% typed, no suggestions flag), unit + type suites green

alies-dev:
## Issue to Solve

Not-yet-stable rules and handlers currently ship unconditionally: every user runs new analysis the moment it merges, with no opt-in period and no documented path to graduate or withdraw a feature. This adds a config-level gate so new analysis gets real-world testing without exposing every user to churn.

## Related

Closes #1203

## Solution Description

New `<experimental>` XML element inside `<pluginClass>` in `psalm.xml`:

```xml
<!-- granular opt-in -->
<experimental>
    <feature name="modelToArrayShape" />
</experimental>

<!-- or: enable every experimental feature -->
<experimental all="true" />
```

- `src/Config/ExperimentalFeature.php`: string-backed enum, the single registry of live experiments (currently just `ModelToArrayShape`), plus `GRADUATED`/`WITHDRAWN` maps for the deprecation-notice lifecycle described in `docs/contributing/README.md`.
- `src/Config/PluginConfig.php`: parses both forms, validates strictly (unknown name throws `InvalidArgumentException` with a Levenshtein "did you mean" hint and the full valid-name list; invalid `all=` value throws), and exposes `isExperimentEnabled()` / `activeExperimentalFeatures()`.
- `src/Plugin.php`: gates `ModelToArrayShapeHandler` (the pilot feature, merged via #1168) behind `ExperimentalFeature::ModelToArrayShape`, keeping the required `require_once` pairing. Prints one line per run naming active experiments (suppressed under `--no-progress`), and forwards graduated/withdrawn/childless-element notices as `Progress::warning()` calls.
- `psalm-laravel diagnose` reports which experimental features are active, or "none (N available, see docs/config.md)".
- `docs/config.md` gained an `## experimental` section (stability contract, both enable forms, current feature list); `docs/contributing/README.md` gained the introduce/stabilize/withdraw lifecycle.

**Deviation from the issue's design sketch:** the issue proposed `trigger_error(..., E_USER_DEPRECATED)` for graduated/withdrawn/childless-element notices. That crashes real `vendor/bin/psalm` runs — Psalm's CLI `ErrorHandler` converts every PHP error/warning/deprecation into a thrown exception during a run, and `Config::initializePlugins()` re-wraps it as a fatal `ConfigException`. Notices are instead collected on `PluginConfig` and surfaced via `Plugin::reportActiveExperiments()`'s `Progress::warning()` calls. Covered by a regression test running the real Psalm binary (`tests/Type/tests/ExperimentalEmptyElementDoesNotCrashTest.phpt`).

**Known follow-up, deliberately out of scope:** `PluginConfig::resolveCachePath()` has a pre-existing `trigger_error(E_USER_DEPRECATED)` call for the deprecated `PSALM_LARAVEL_PLUGIN_CACHE_PATH` env var, with the identical crash bug. Left untouched here since it's unrelated to the experimental-config feature; worth a follow-up issue.

Verified with `composer psalm` (100% type coverage, no suggestions flag), `composer test:unit`, `composer test:type`, `composer cs`, `composer rector` — all clean after rebasing onto current `origin/4.15`.

## Checklist
- [x] Tests cover the change (type tests in `tests/Type/` for the pilot feature's on/off gating and the crash regression; unit tests in `tests/Unit/` for `PluginConfig`, `Diagnostics`, `IssueUrlGenerator`)
- [x] Documentation is updated (`docs/config.md`, `docs/contributing/README.md`)

---
<a href="https://app.blacksmith.sh/psalm/codesmith/psalm-plugin-laravel/pr/1209"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785573504&installation_id=141955579&pr_number=1209&repository=psalm%2Fpsalm-plugin-laravel&return_to=https%3A%2F%2Fgithub.com%2Fpsalm%2Fpsalm-plugin-laravel%2Fpull%2F1209&signature=2bc452560083b8c8889648dc07ba9d582882464c84e828b4bb17dbcb1bfb8ce6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


alies-dev:
Closing this PR as superseded by a simpler design that will be implemented fresh from the latest `4.15` branch.

The release policy has changed:

- `ModelToArrayShapeHandler` is now considered stable for v4.15 and will be registered unconditionally.
- Experimental configuration will be a single boolean: `<experimental value="true" />`.
- The boolean will only promote the default severity of experimental plugin issues (`UnknownModelAttribute` and `UndefinedModelRelation`) from `info` to `error`.
- Explicit Psalm `issueHandlers` configuration will always win.
- `UnresolvableAppendedModelAttribute` remains a stable error.

That means the named-feature grammar in this PR—`<feature>`, `all="true"`, `<exclude>`, feature retirement maps, and per-feature diagnostics—is no longer the intended architecture. Rewriting this 1,499-line diff incrementally would make the replacement harder to review than a clean implementation.

Useful findings from this PR will still carry forward, especially strict XML parsing, the `psalm.xml.dist` diagnose fallback, real-Psalm regression coverage, and the discovery that `trigger_error()` during plugin initialization can abort analysis.

#1203 remains open and will be closed by the replacement PR.

Issue #733, opened 2026-04-12
False positive TaintedSql on firstWhere() array form: @psalm-taint-sink sql on $column hits array values


alies-dev:
## Description

`Builder::firstWhere()` has `@psalm-taint-sink sql $column` in its stub, intended to catch column-name injection (`firstWhere($request->input('col'))`). However, when called with an associative array, the **values** inside the array also go into `$column`, which causes a false positive `TaintedSql` even though Eloquent parameterizes all values.

## Stub (Eloquent/Builder.stubphp)

```php
/**
 * @psalm-taint-sink sql $column   ← hits array values too
 * @psalm-taint-escape sql
 * @psalm-flow ($operator, $value) -> return
 * @psalm-taint-specialize
 */
public function firstWhere($column, $operator = null, $value = null, $boolean = 'and') {}
```

## Reproduction

```php
$sid = $request->input('id');   // tainted

// Flagged as TaintedSql — but this is a safe parameterized query
UserToken::firstWhere([
    'driver_id' => $socialite->getId(),  // tainted (Http\Client\Response source)
    'driver'    => $driver,
]);

// Also flagged
Conversation::where(['status_id' => $sid, 'to_id' => $dm->from_id])->delete();
```

Found in real apps:
- `app/Actions/AttemptToAuthenticateSocialite.php` (monica)
- `app/Http/Controllers/DirectMessageController.php:401,407` (pixelfed)
- `app/Http/Controllers/Groups/GroupsCommentController.php:346` (pixelfed)
- `app/Http/Controllers/Groups/GroupsPostController.php:351` (pixelfed)

## Root Cause

When `firstWhere(['key' => $taintedValue])` is called:

- The array (including its tainted values) goes into the `$column` parameter
- `@psalm-taint-sink sql $column` fires because `$column` received tainted data
- Eloquent's actual SQL execution is fully parameterized — no injection risk

Compare with `where()`, which has **no** `@psalm-taint-sink sql $column` and never false-positives on the array form:

```php
// where() stub — correct, no false positives on array form
* @psalm-taint-escape sql
* @psalm-flow ($value) -> return
* @psalm-taint-specialize
```

## Fix

Remove `@psalm-taint-sink sql $column` from `firstWhere()`, making it consistent with `where()`:

```php
/**
 * @psalm-taint-escape sql
 * @psalm-flow ($operator, $value) -> return
 * @psalm-taint-specialize
 */
[...]

PR #1218, opened 2026-07-03
fix: stop `where()` array form flagging false `TaintedSql`

### LINKED ISSUE #734: False positive TaintedSql: @psalm-taint-sink sql $column on where()/orWhere()/having() fires on array VALUES, not just column names
## Description

`where()`, `orWhere()`, `whereNot()`, `orWhereNot()`, `having()`, and `orHaving()` all have `@psalm-taint-sink sql $column` in `stubs/common/Database/Query/Builder.stubphp`. This annotation was added to catch column-name injection (`where($request->input('col'), ...)`), but it also fires when these methods are called with the **array form** — where `$column` receives a `['column' => $value]` map and the VALUES inside the array are tainted. The values are parameterized by Eloquent and are safe, so these are false positives.

## Affected stubs (Query/Builder.stubphp)

```php
// where(), orWhere(), whereNot(), orWhereNot(), having(), orHaving()
// All have the same pattern:
* @psalm-taint-sink sql $column   ← fires on VALUES inside ['col' => $tainted]
* @psalm-taint-escape sql
* @psalm-flow ($operator, $value) -> return
```

Note: `$column` is typed as `\Closure|string|array` — the array overload puts VALUES into `$column`.

## Reproduction

```php
$sid = $request->input('id');   // tainted

// ❌ Incorrectly flagged as TaintedSql — safe, Eloquent parameterizes all values
Conversation::where([
    'status_id' => $sid,
    'to_id'     => $dm->from_id,
    'from_id'   => $dm->to_id,
])->delete();

GroupLike::where([
    'group_id'   => $gid,
    'profile_id' => $pid,
    'comment_id' => $sid,
])->first();
```

Found in real apps:
- `app/Http/Controllers/DirectMessageController.php:401,407` (pixelfed)
- `app/Http/Controllers/Groups/GroupsCommentController.php:346` (pixelfed)
- `app/Http/Controllers/Groups/GroupsPostController.php:351` (pixelfed)

## Root cause

When `where(['status_id' => $tainted])` is called, the entire array — including its tainted VALUES — lands in the `$column` parameter. `@psalm-taint-sink sql $column` then fires `TaintedSql` because `$column` received a tainted argument. But the VALUES in the array are bound as parameterized placeholders by Eloquent/PDO — there is no SQL injection risk.

Contrast with `whereRaw($sql)` where the whole string IS dangerous: that correctly has `@psalm-taint-sink sql $sql` and should not change.

## Relationship to #733

This is the same root cause as #733 (`firstWhere()` false positive) but affects the base Query Builder `where()` family. Both issues stem from `@psalm-taint-sink sql $column` firing on array VALUES that are safely parameterized.

## Proposed fix

Remove `@psalm-taint-sink sql $column` from the six affected methods:

```php
// Before
* @psalm-taint-sink sql $column   ← remove this line
* @psalm-taint-escape sql
* @psalm-flow ($operator, $value) -> return

// After
* @psalm-taint-escape sql
* @psalm-flow ($operator, $value) -> return
```

Apply to: `where()`, `orWhere()`, `whereNot()`, `orWhereNot()`, `having()`, `orHaving()`.

Keep `@psalm-taint-sink sql $column` only on `orderBy()` and other methods where a raw column identifier string genuinely matters (since `orderBy()` does NOT parameterize its column argument, unlike the WHERE/HAVING methods which use PDO binding for values).

**Why it's safe to remove:** Column names in WHERE clauses are always developer-written strings or Eloquent expression objects — not user input. User data goes into the VALUE positions (`$operator`/`$value` params or array values), which are correctly handled via `@psalm-flow ($operator, $value) -> return` + `@psalm-taint-escape sql`. The only genuine SQL identifier injection risk is in raw SQL methods (`whereRaw`, `selectRaw`, `orderByRaw`, `orderBy`) which have their own correct sink annotations.

alies-dev:
## Issue to Solve

`@psalm-taint-sink sql $column` on the `where()` / `having()` family fires on the **keyed-map** form:

```php
$query->where(['status_id' => $request->input('status')]); // false TaintedSql
```

Laravel routes the array form through `Builder::addArrayOfWheres()`, which for a string key calls `where($key, '=', $value)` — the literal **key** is the column identifier and each **value** is PDO-bound. A tainted value there is never interpolated, so the report is a false positive. Confirmed in pixelfed (`DirectMessageController`, `GroupsCommentController`, `GroupsPostController`) and monica. The string form `where($request->input('column'))` is a genuine identifier-injection vector and must keep reporting.

## Related

Closes #734. References #733 (`firstWhere`, closed as a duplicate of #734).

## Solution Description

Psalm cannot express a type-scoped sink (`@psalm-taint-sink` is unconditional), and `AddRemoveTaintsEvent` carries no method id — so a stub-only change cannot tell the where-family (values bound) apart from `whereAll()` (values *are* columns).

The stubs are therefore **unchanged**. A new `WhereColumnTaintHandler` implements `RemoveTaintsInterface` and removes the `sql` taint from an argument **only when its type is a sealed keyed-map literal** — a `TKeyedArray` with all-string keys and no `...<string,mixed>` fallback (`['col' => $v]`). That shape is unique to the where-family value-binding form:

- **map** `['col' => $v]` → stripped → no false positive
- **string** `where($request->input('col'))` → still reports (real injection)
- **whole array** `where($request->all())` (`array<string,mixed>`, user-controlled keys) → still reports
- **list** `whereAll(['a','b'])` (integer keys) → still reports

### Why RemoveTaints and not a per-call-site sink

An earlier draft added a per-call-site sink, which required `@psalm-taint-specialize` on the stubs. That **silently breaks `@psalm-flow` propagation of non-SQL taint** (html/shell/secret) through the `$value` positions — the failure mode documented in `docs/contributing/taint-analysis.md`, reproduced here by an A/B test. Stripping the taint at the argument needs no specialize, so the flow is preserved (guarded by `TaintedShellWhereValueFlowPreserved.phpt`).

### Verification

New taint type-tests under `tests/Type/tests/TaintAnalysis/`:
- map → clean: `SafeSqlWhereArrayValues`, `SafeSqlEloquentWhereArrayValues`, `SafeSqlFirstWhereArrayValues`, `SafeSqlOrWhereNotArrayValues`
- still reports: `TaintedSqlWhereWholeArrayInput` (`$request->all()`), `TaintedSqlFirstWhereColumnSink`
- flow preserved: `TaintedShellWhereValueFlowPreserved` (a non-SQL taint through a where value)
- documented limitation: `SafeSqlWhereAllMapKnownLimitation`

Existing string-form (`TaintedSqlWhereColumnSink`, `…Eloquent…`) and `whereAll`/`whereAny`/`whereNone` tests are unchanged and green. Full `test:type`, `test:unit`, `psalm` (100% coverage), `cs`, `rector` all pass.

### Known limitation

`whereAll(['col' => $tainted])` — a keyed **map** passed to `whereAll` (whose values become columns) — is **not** flagged (the map strip cannot distinguish it from a where value-map). This is malformed usage (`whereAll` takes a list) that also trips a `MixedArgumentTypeCoercion`; the realistic list form is covered. Documented in `SafeSqlWhereAllMapKnownLimitation.phpt`.

### Upstream layer

The truly correct fix is upstream: Psalm should support type-scoped sinks (e.g. `@psalm-taint-sink sql $column string`) or method-id + argument-offset context on `AddRemoveTaintsEvent`. Until then this handler is an interim, auto-retiring workaround (same pattern as `ConditionableWhenHandler`). A vimeo/psalm feature request will be filed and linked here.

### Backporting to 3.x (Psalm 6)

The same false positive exists on `3.x`. The handler is portable — only the `RemoveTaintsInterface` contract differs between Psalm 6 and 7. Copy `WhereColumnTaintHandler` onto `3.x` and change **only** the `removeTaints` method (the `isBoundValueMap` helper and every import/API — `getExpr`/`ArrayItem`, `StatementsAnalyzer::$node_data`, `Union::isSingle()`/`getSingleAtomic()`, `TKeyedArray::$properties`/`$fallback_params` — exist identically in Psalm 6.16):

```php
// Psalm 7 (this PR / master / 4.x): returns an int bitmask
public static function removeTaints(AddRemoveTaintsEvent $event): int
{
    // …guards…
    if (!$type instanceof Union || !self::isBoundValueMap($type)) {
        return 0;
    }
    return TaintKind::INPUT_SQL;
}

// Psalm 6 (3.x): returns list<string>
public static function removeTaints(AddRemoveTaintsEvent $event): array
{
    // …guards… (return [] instead of 0)
    if (!$type instanceof Union || !self::isBoundValueMap($type)) {
        return [];
    }
    return [TaintKind::INPUT_SQL]; // TaintKind::INPUT_SQL is the string 'sql' on Psalm 6
}
```

A single class cannot satisfy both (PHP fatals on the `array` vs `int` return-type mismatch), so the backport is a separate PR on `3.x`, consistent with the plugin's 3.x/4.x split. Register it the same way (paired `require_once` + `registerHooksFromClass`) and port the phpts (`TaintKind` strings match, so the `--EXPECTF--` output is identical). Note: use `$atomic->fallback_params` (not the Psalm-7-only `isSealed()`) so the helper stays byte-identical across versions.

## Checklist
- [x] Tests cover the change (8 type tests in `tests/Type/tests/TaintAnalysis/`)
- [x] Documentation is updated (`docs/contributing/taint-analysis.md`)

---
<a href="https://app.blacksmith.sh/psalm/codesmith/psalm-plugin-laravel/pr/1218"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785681352&installation_id=141955579&pr_number=1218&repository=psalm%2Fpsalm-plugin-laravel&return_to=https%3A%2F%2Fgithub.com%2Fpsalm%2Fpsalm-plugin-laravel%2Fpull%2F1218&signature=51e4d22dfd9fc0db6afc03c90bd8064b85c8f4a60eb6d06a1d7707614b407544"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -75,3 +75,7 @@ Unparseable = "Unparseable"
 # word here to cover every compound, current and future.
 Encrypter = "Encrypter"
 encrypter = "encrypter"
+# `addArrayOfWheres` is Laravel's real `Illuminate\Database\Query\Builder` method; it
+# appears in WhereColumnTaintHandler's docblock and the where-family taint docs/tests.
+# typos tokenizes the `Wheres` sub-word and flags it as a typo of "Whereas".
+Wheres = "Wheres"
--- a/docs/contributing/taint-analysis.md
+++ b/docs/contributing/taint-analysis.md
@@ -315,6 +315,8 @@ Both `$operator` and `$value` appear in `@psalm-flow` because in the **2-argumen
 
 The same pattern applies to `orWhere()`, `whereNot()`, `orWhereNot()`, `having()`, and `orHaving()`.
 
+The keyed-**map** array form `where(['col' => $value])` is a false positive under the plain sink: `Builder::addArrayOfWheres()` binds each value and uses only the (literal) key as the column, so a tainted value is never interpolated (#734/#733). `Psalm\LaravelPlugin\Handlers\Eloquent\WhereColumnTaintHandler` removes the `sql` taint for exactly that shape (a `TKeyedArray` with all-string keys) — see its docblock. Do **not** "fix" it by dropping the sink (the string form is a real vector) or by adding `@psalm-taint-specialize` to these stubs — the latter silently breaks the non-SQL `@psalm-flow` on the value positions (see the specialize note below).
+
 ### Pattern for find-family methods
 
 ```php
--- a/src/Handlers/Eloquent/WhereColumnTaintHandler.php
+++ b/src/Handlers/Eloquent/WhereColumnTaintHandler.php
@@ -1,0 +1,124 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Eloquent;
+
+use PhpParser\Node\ArrayItem;
+use Psalm\Internal\Analyzer\StatementsAnalyzer;
+use Psalm\Plugin\EventHandler\Event\AddRemoveTaintsEvent;
+use Psalm\Plugin\EventHandler\RemoveTaintsInterface;
+use Psalm\Type\Atomic\TKeyedArray;
+use Psalm\Type\TaintKind;
+use Psalm\Type\Union;
+
+/**
+ * Removes the `sql` taint from a where-family `$column` argument when it is a keyed-MAP literal —
+ * `where(['status_id' => $userValue])` — killing the false positive #734 / #733 without weakening the
+ * string-form sink.
+ *
+ * The stubs keep the plain `@psalm-taint-sink sql $column` on the where/having family (see
+ * `stubs/common/Database/Query/Builder.phpstub` etc.). That is correct for the string form
+ * `where($column)`, which interpolates `$column` as a raw identifier — a real injection vector. But
+ * the array form routes through `Builder::addArrayOfWheres()`, which for a string key calls
+ * `where($key, '=', $value)`: the literal KEY is the column and each VALUE is PDO-bound. A tainted
+ * value in that map is never interpolated, so the blanket sink over-reported (confirmed in pixelfed /
+ * monica). This handler subtracts exactly that case.
+ *
+ * ## Why a map, not any array
+ *
+ * The gate is a keyed array with ALL STRING keys — the value-binding shape `['col' => $v]`. Two
+ * shapes are deliberately NOT stripped, because for them the array elements ARE column identifiers:
+ *
+ *  - a LIST literal (integer keys) — `whereAll(['a', 'b'])` / `where([[$col, '=', $v]])`. Laravel's
+ *    `whereAll`/`whereAny`/`whereNone` iterate the array's VALUES as columns, and
+ *    `addArrayOfWheres`' numeric-key branch spreads a nested `[[$col, …]]` so element[0] is a raw
+ *    column. Integer keys therefore keep the sink.
+ *  - a generic `array<string, mixed>` — `where($request->all())` — whose keys are unknown and can be
+ *    user-controlled column names. Not a `TKeyedArray`, so it keeps the sink.
+ *
+ * ## Why RemoveTaints and not a per-call sink handler
+ *
+ * A `@psalm-taint-sink` annotation is unconditional — it cannot depend on the argument's type — and a
+ * per-call-site sink handler would need either fragile node-id replication or a stub
+ * `@psalm-taint-specialize`, which silently breaks `@psalm-flow` propagation of non-SQL taint through
+ * the value positions (the documented failure mode in `docs/contributing/taint-analysis.md`). Removing
+ * the taint at the argument, keyed off the argument's own type, needs neither: the keyed-string-map
+ * shape is unique to the where-family value-binding form, so although {@see removeTaints} is
+ * method-agnostic (the event carries no method id), in real code only a where-family map matches —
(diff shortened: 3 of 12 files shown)


alies-dev:
Superseded by #1221. This PR scopes the `sql`-taint strip by argument *type* (any sealed all-string-key array), which is method-agnostic and produces false negatives: `having(['col' => $t])` (no `addArrayOfWheres` branch, so the array compiles as a raw identifier), `whereAll(['col' => $t])` (values are columns), and a map that reaches a raw sink elsewhere (`$m = ['col' => $t]; whereRaw($m)` / returned then interpolated). #1221 records the where-family first-argument nodes and strips only those, keeping the taint everywhere else. Closing in favor of #1221.

PR #1236, opened 2026-07-11
Resolve concrete-only contract methods via container bindings

### LINKED ISSUE #1230: UndefinedInterfaceMethod for `Cache::driver()->flexible()`: narrow `CacheManager` returns to concrete `Illuminate\Cache\Repository`
**Describe the bug**

`Cache::driver()->flexible(...)` (and the same call via `Cache::store()` or an injected `CacheManager`) is reported as `UndefinedInterfaceMethod`:

```
Method Illuminate\Contracts\Cache\Repository::flexible does not exist (see https://psalm.dev/181)
```

Root cause: `flexible()` (added in Laravel 11.23, SWR caching) is declared only on the concrete `Illuminate\Cache\Repository`, not on the `Illuminate\Contracts\Cache\Repository` interface. `CacheManager::store()` / `driver()` / `memo()` docblock the contract as their return type, so Psalm resolves the call against the interface and fails.

At runtime these methods always return the concrete `Illuminate\Cache\Repository`: every store, including custom `Cache::extend()` drivers, is wrapped by `CacheManager::repository()`. Narrowing the manager's return type to the concrete class is therefore sound and gives access to the full concrete surface (`flexible()`, `tags()`, Macroable) without lying about the interface.

**Proposed fix**

Add a `stubs/common/Cache/CacheManager.phpstub` overriding the return type of `store()`, `driver()` and `memo()` to `\Illuminate\Cache\Repository`. The `Cache` facade path already resolves to `CacheManager` via `FacadeMapProvider`, so the facade call site is fixed by the same stub.

Alternatives considered and rejected:
- Adding `@method flexible(...)` to a contract stub: lies about the interface; a custom `Contracts\Cache\Repository` implementation genuinely lacks the method (false negative).
- Upstream `laravel/framework` change: adding `flexible()` to the contract is a BC break; a docblock-only `@return` narrowing on `CacheManager` may be declined because `Cache::swap()` / container rebinding can inject any contract implementation.

Stub-authoring caveats (per `docs/contributing/`): copy the `CacheManager` class header verbatim (a bare re-declaration strips implements/traits), and keep the version gate in mind: `flexible()` exists since Laravel 11.23, so the `Repository::flexible()` signature itself is fine in `stubs/common/` only if the plugin's supported Laravel floor is >= 11.23; otherwise it belongs in a version dir.

Related: #701 typed these calls down to `Illuminate\Contracts\Cache\Repository`, which is exactly the level that misses concrete-only methods like `flexible()`.

**Code to reproduce**

```php
<?php

use Illuminate\Support\Facades\Cache;

/** @return array<string> */
function loadThings(): array
{
    return Cache::driver('file')->flexible('things', [5 * 60, 60 * 60], static function (): array {
        return ['a', 'b'];
    });
}
```

**Impacted Versions**

```text
laravel/framework    v12.62.0
psalm/plugin-laravel dev-master
vimeo/psalm          7.0.0-beta19
```

**Psalm config**

```xml
<?xml version="1.0"?>
<psalm errorLevel="4" xmlns="https://getpsalm.org/schema/config">
    <projectFiles>
        <directory name="app" />
    </projectFiles>
    <plugins>
        <pluginClass class="Psalm\LaravelPlugin\Plugin"/>
    </plugins>
</psalm>
```

**Additional context**

The `memo()` driver wraps stores in `MemoizedStore` but still returns the concrete `Illuminate\Cache\Repository`, so it takes the same narrowing.

alies-dev:
## Issue to Solve

`Cache::driver()->flexible(...)` (and any concrete-only method called on a contract-typed receiver) raises `UndefinedInterfaceMethod`: `CacheManager::store()/driver()/memo()` docblock the `Contracts\Cache\Repository` interface while the runtime value is the concrete `Illuminate\Cache\Repository`. The same shape recurs across the framework (it is the #1108 problem generalized): the plugin's `ContractMethodBridgeHandler` already solved it, but only for an allow-list containing exactly `Contracts\Foundation\Application`.

## Related

Closes #1230
Refs #1108

Alternative to #1233, which fixes the same report by narrowing the three `CacheManager` returns to the concrete class. This PR instead adopts Larastan's general approach (its `ContractsMethodsExtension`): keep the honest contract types and make concrete-only methods resolve on the contract. The two are complementary rather than conflicting (narrowing gives more precise types at the three call sites; bridging covers every contract), but either alone fixes #1230. Maintainer picks: merge one or both.

## Solution Description

`ContractMethodBridgeHandler` now walks every scanned `Illuminate\Contracts\*` interface at `AfterCodebasePopulated` and bridges missing public non-magic methods from whatever the booted container actually resolves for that contract (`make()` in a `Throwable` guard; unresolvable contracts are skipped). Bridging still writes only `declaring_method_ids`/`appearing_method_ids`, never `methods[]`, so implementors are not forced to declare bridged methods. Because resolution is per-app, a project that rebinds a contract to a custom implementation bridges to its own concrete instead of a hardcoded framework class.

Two policy boundaries, both recorded in `docs/contributing/decisions.md`:

- `Contracts\Filesystem\Filesystem`, `Contracts\Filesystem\Cloud`, and `Contracts\Queue\Queue` remain excluded: their resolved concrete class itself varies by configured driver (`AwsS3V3Adapter` adds `temporaryUrl()`/`getClient()` that other disks lack; `RedisQueue`/`SqsQueue`/`DatabaseQueue` diverge similarly), so bridging the default driver's surface would lie at the class-hierarchy level for every other driver. This mirrors the existing `StorageHandler` policy for `disk()`/`drive()`. Cache is different: every driver constructs the same `Illuminate\Cache\Repository`, only the wrapped `Store` varies.
- Registration order in `Plugin.php` became load-bearing: bridging turns `AuthManager`'s `@mixin Guard` methods into real methods, rerouting Psalm's dispatch onto the `Guard`/`StatefulGuard` contracts. `AuthMethodHandler` therefore now registers on those contracts too (keeping `authenticate()`/`getUser()` narrowed to the config-derived user model), and `GuardHandler` must register first so its chain-aware `user()`/`loginUsingId()` logic keeps winning.

Verified by: the #1230 repro as a type test (`Cache::driver('file')->flexible(...)`, a `Contracts\Cache\Repository`-typed param calling `flexible()`/`tags()`, `cache()->driver()->flexible(...)`); a rewritten scope test pinning that `Contracts\Container\Container::wrap()` now bridges (intentional policy flip) while user-land interfaces outside `Illuminate\Contracts` never bridge; a new exclusion test pinning that `temporaryUrl()` on the Filesystem contract and a queue-driver-specific method still raise `UndefinedInterfaceMethod`; updated unit tests for the resolution seam. Full type suite (561), unit suite (829), `composer psalm`, `composer cs:check`, and the fresh-Laravel-app integration run (`composer test:app`) are all green. `flexible()` signatures were checked against Laravel v12.14.0 (the supported floor) and v13.19.0.

## Checklist
- [x] Documentation is updated (docs/contributing/decisions.md)

---
<a href="https://app.blacksmith.sh/psalm/codesmith/psalm-plugin-laravel/pr/1236"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786384245&installation_id=141955579&pr_number=1236&repository=psalm%2Fpsalm-plugin-laravel&return_to=https%3A%2F%2Fgithub.com%2Fpsalm%2Fpsalm-plugin-laravel%2Fpull%2F1236&signature=c7e9f71086a758070f0dcd4d1cfbe3689010fd69b16c4fd3768705b3d7826486"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


alies-dev:
Closing this alternative approach in favor of producer-specific return narrowing.

The global contract bridge widens every value typed as an Illuminate contract based on the app's default binding. That is unsound for driver-varying or custom implementations (for example Auth guards, hashers, queues, and filesystems) and can hide real UndefinedInterfaceMethod errors. It also eagerly resolves every contract during analysis.

We will instead narrow only known producer calls where the framework runtime guarantees a concrete result, while keeping bare contract values contract-faithful. #1233 covers Cache; follow-up work will cover other verified producers such as PasswordBrokerManager::broker(), View factory results, and Query Builder pagination.

This exploration established the broader design boundary.

PR #1240, opened 2026-07-12
Narrow stable Laravel producer returns to concrete implementations

alies-dev:
## Issue to Solve

`UndefinedInterfaceMethod` false positives where a known Laravel producer declares a **contract** return, but its framework implementation always constructs one stable **concrete**. Calling a concrete-only method on the result was flagged even though it is correct at runtime:

```php
Password::broker()->createToken($user);   // createToken() is on the concrete, not the contract
view('welcome')->fragment('x');           // fragment() is on Illuminate\View\View
DB::table('users')->paginate()->linkCollection();
trans()->has('some.key');
```

This continues the pattern established for `Cache::driver()->flexible()` in #1230 / #1233 (merged as `CacheManagerReturnTypeHandler`).

## Related

Conceptual follow-on to #1230 / #1233 (no dedicated tracking issue).

## Solution Description

Narrow **by provenance, never by conformance**. A value is narrowed only when it originates from a known producer expression; a bare contract-typed value keeps the contract-only surface.

| Producer | Declared contract | Concrete result | Source evidence (Laravel 12.x + 13.x) |
|---|---|---|---|
| `PasswordBrokerManager::broker()` + `Password` facade / root alias | `Contracts\Auth\PasswordBroker` | `Auth\Passwords\PasswordBroker` | `PasswordBrokerManager::resolve()` hard-constructs `new PasswordBroker(...)`; no broker-driver extension API |
| `View\Factory::make()` / `file()` / `first()` + `View` facade / alias | `Contracts\View\View` | `View\View` | `Factory::viewInstance()` hard-constructs `new View(...)` |
| `view()` (0 args) | `Contracts\View\Factory` | app's resolved factory class | helper branches on `func_num_args() === 0` |
| `view('name')` (≥1 arg) | `Contracts\View\View` | `View\View` (only for the stock `Illuminate\View\Factory`) | same helper |
| `Query\Builder::paginate()` | `Pagination\LengthAwarePaginator` (ungenerified) | `LengthAwarePaginator<int, stdClass>` | `paginate()` builds the concrete via `paginator()`; Query rows hydrate as `stdClass` |
| `Query\Builder::simplePaginate()` | `Contracts\Pagination\Paginator` | `Paginator<int, stdClass>` | same |
| `Query\Builder::cursorPaginate()` | `Contracts\Pagination\CursorPaginator` | `CursorPaginator<int, stdClass>` | same |
| `trans()` (0 args) | `Contracts\Translation\Translator` | app's resolved translator class | helper returns `app('translator')` on `is_null($key)` |

**Mechanism.** `ProducerReturnTypeHandler` (new) holds the reviewed producer→concrete registry for the Password and View families. It discriminates a real producer-method call from a facade `@method static` pseudo-call via `methodExists` (a handler is required because facade pseudo-tags shadow any stub-added method), gates the pseudo path on the canonical facade's `pseudo_static_methods` so Psalm 7 cannot fatal with "Cannot get method params", and drops narrowing if the declared return no longer names the expected contract (drift guard). `view()` / `trans()` narrow through their existing handlers by argument count. Query Builder pagination is stub-typed.

**Soundness boundaries**

- Providers register only producer classes, canonical facades, and root aliases — never a contract FQCN; no `ClassLikeStorage` mutation. Bare contract-typed parameters, properties, mocks, and custom implementations keep the contract-only surface (negative tests prove `UndefinedInterfaceMethod` still fires for all four contracts).
- Container-dependent helpers narrow only to the binding resolved at plugin boot; a non-standard / unscanned binding falls back to the contract.
- Disclosed bounded trade-off: producer internals like `Factory::viewInstance()` are `protected`, so a subclass overriding the construction but inheriting `make()` still narrows to the stock concrete. Worst case flips a contract-level false positive into a false negative, never a new false positive (same accepted trade-off as the Cache manager).
- Deliberately excluded: driver-variant surfaces (Auth guards, Hash, Queue, Broadcast, Redis, Filesystem), arbitrary container contracts, and any bridge-every-contract behavior.

**Verification**

- `composer test:type` — 574 pass (11 new PHPTs: positive narrowing + negative soundness per family, `MissingView`/`MissingTranslation`-still-fire ordering regressions, leading-vs-trailing-spread, `view(null)` / `trans(null)` argument-count boundary).
- `composer test:unit` — 887 pass (4 new: `view()` argument-count branch decision incl. non-standard factory subclass fallback).
- `composer test:app`, `composer psalm` (100% type coverage), `composer cs`, `composer rector` — all clean.
- Four review rounds; the one correctness item (a throwing `view` binding must degrade this feature, not disable the plugin, and its warning must not misdescribe the cause) is fixed.

## Checklist
- [x] Tests cover the change (type tests in `tests/Type/` and unit tests in `tests/Unit/`)
- [x] Documentation is updated (`docs/contributing/decisions.md` — Producer Return Narrowing policy)

---
<a href="https://app.blacksmith.sh/psalm/codesmith/psalm-plugin-laravel/pr/1240"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786440514&installation_id=141955579&pr_number=1240&repository=psalm%2Fpsalm-plugin-laravel&return_to=https%3A%2F%2Fgithub.com%2Fpsalm%2Fpsalm-plugin-laravel%2Fpull%2F1240&signature=feb85561d59e4d39acc1b7a042a983d220b761931a18ff2bcfae8efc86f412ed"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/contributing/decisions.md
+++ b/docs/contributing/decisions.md
@@ -249,6 +249,21 @@ Bug fixes (where the previous type was demonstrably wrong) are exempt.
 
 **Why:** A method named `posts()` that returns a `HasMany` relation should always be treated as a relationship property, even if a migration column named `posts` also exists. Similarly, an accessor `getFullNameAttribute()` should take priority over a `full_name` column. The order reflects specificity: relationships and accessors are explicit code the developer wrote; columns are inferred from migrations and serve as the fallback.
 
+## Producer Return Narrowing
+
+### Provenance may narrow, conformance never widens
+
+**Decision:** A stable producer (a framework method that hard-constructs one concrete with no supported extension point) narrows its declared contract return to that concrete. `ProducerReturnTypeHandler` holds the reviewed mapping (`PasswordBrokerManager::broker()`, `View\Factory::make()/file()/first()`, plus facades and root aliases). `view()`/`trans()` narrow via their existing handlers: zero-arg forms narrow to whatever class the binding resolved to at plugin boot (a scanned subclass counts); the argument-supplied `view('name')` form additionally requires the stock `Illuminate\View\Factory`. `Query\Builder` pagination is stub-narrowed (`stdClass` rows).
+
+**Why:** Laravel declares contracts where runtime always yields one concrete, so concrete-only calls like `Password::broker()->createToken()` report false `UndefinedInterfaceMethod`.
+
+**Boundaries:**
+- Key on the producing expression, never the contract: contract FQCNs are never registered, interface storage never modified. Bare contract-typed values (params, properties, mocks, custom impls) keep the contract-only surface.
+- Drift guard: each rule requires the declared return to still name the expected contract, else it disables itself.
+- New mappings need source verification across supported Laravel versions plus positive and negative tests.
+- Disclosed gap: producer internals like `Factory::viewInstance()` are protected, so a producer subclass inheriting the mapped method but overriding the construction still narrows to the stock concrete. Accepted (matches the Cache manager precedent): worst case turns a contract-level false positive into a false negative, never a new false positive.
+- Excluded: driver-variant surfaces (Auth guards, Hash, Queue, Broadcast, Redis, Filesystem) where the concrete depends on runtime config.
+
 ## Third-Party Package Support
 
 ### Plugin covers Laravel framework only, not third-party packages
--- a/docs/contributing/decisions.md
+++ b/docs/contributing/decisions.md
@@ -261,7 +261,7 @@ Bug fixes (where the previous type was demonstrably wrong) are exempt.
 - Key on the producing expression, never the contract: contract FQCNs are never registered, interface storage never modified. Bare contract-typed values (params, properties, mocks, custom impls) keep the contract-only surface.
 - Drift guard: each rule requires the declared return to still name the expected contract, else it disables itself.
 - New mappings need source verification across supported Laravel versions plus positive and negative tests.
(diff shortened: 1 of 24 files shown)


alies-dev:
Non-blocking cleanup done (4bc98f7d):
- 'view' binding resolved once in `__invoke()` and threaded into both init paths (no more double-resolve when `view.finder` is absent).
- decisions.md corrected: zero-arg `view()`/`trans()` narrow to the resolved binding only when it is the framework class or a subclass, not "whatever class" it resolved to.
- `CustomView::badge()` is now actually called, so the `__call` masking of the false-positive direction is asserted rather than just described.

Gates green: type 574, unit 888, Psalm 100% coverage, cs, rector.

PR #1271, opened 2026-07-15
Narrow Eloquent model `getKey()` return types

### LINKED ISSUE #1270: `getKey()` return not narrowed to model's known primary-key type (int/string)
## Problem

`$model->getKey()` always infers `int|string`, even when the model's primary-key type is provably `string` (or `int`). The `ModelMetadataRegistry` already knows the precise type, but nothing wires it into the `getKey()` return.

```php
class UlidModel extends Model { use HasUlids; }

$m = new UlidModel();
$m->getKey(); // inferred: int|string   — expected: string
```

## Cause

- `stubs/common/Database/Eloquent/Model.phpstub:132` hardcodes `getKey(): int|string`.
- `ModelMetadata` stores a `PrimaryKeyType` enum (`Integer`/`String`), computed from `getKeyType()` **after trait overrides** (`HasUuids`/`HasUlids` force `string`) — see `ModelMetadataRegistryBuilder` (`getKeyType()` read, `PrimaryKeyType` mapping).
- No `MethodReturnTypeProvider` overrides `getKey()`. `ModelMethodHandler` (the `Model::class` provider) returns `null` for `getKey`, so the call falls through to the stub's `int|string`.
- `PrimaryKeyType` is consumed today only for collection/relation keying (`ModelPropertyResolver::resolveKeyType`), never for the `getKey()` method itself.

## Proposed fix

Add a method return-type provider (or extend `ModelMethodHandler`) for `getKey()` that reads `ModelMetadata::primaryKeyType()` for the receiver model and narrows to `int` or `string`. Fall back to `int|string` when the model is unknown to the registry.

`getKeyName()` is already `string` in the stub, so no change needed there. Consider whether `getKeyType()` itself (returns `'int'|'string'` literal) is also worth narrowing.

## Notes

- Registry-dependent (`ModelMetadataRegistry` is now on `master`).
- Soundness: for a non-final model, the receiver may be a subclass that re-declares `$keyType`; narrowing should be keyed on the statically-known model class, matching how other registry consumers resolve metadata.

## Labels

eloquent, type-narrowing, priority:low

alies-dev:
## Issue to Solve

`Model::getKey()` keeps the broad `int|string` stub type even when model metadata knows the primary-key type.

## Related

Fixes #1270.

## Solution Description

Use complete primary-key metadata to return `int` or `string` for the inherited framework implementation. Unknown, incomplete, keyless, and custom `getKey()` models continue to use the existing fallback or declared return type.

Added focused type coverage for integer, string, UUID, ULID, custom-key, union, keyless, unknown, and overridden models, plus a unit check for incomplete metadata.

## Checklist

---
<a href="https://app.blacksmith.sh/psalm/codesmith/psalm-plugin-laravel/pr/1271"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786732790&installation_id=141955579&pr_number=1271&repository=psalm%2Fpsalm-plugin-laravel&return_to=https%3A%2F%2Fgithub.com%2Fpsalm%2Fpsalm-plugin-laravel%2Fpull%2F1271&signature=7258b0fc0aca1be3882fae41e3dc97967ca18c88164497aa77691603078cb839"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


alies-dev:
it's a wrong way to implement it

PR #1288, opened 2026-07-17
Fix: pluck() and aggregate narrowing on custom subclasses, raw aliases, and schema-typed columns

### LINKED ISSUE #1286: `pluck($value, $key)` loses key narrowing when the value column is a raw select alias (not a `@property`)
**Describe the bug**

`BuilderPluckHandler` bails out entirely when the *value* column passed to `pluck($value, $key)` is not a known `@property` on the model. This is the standard aggregate pattern where the value column is a raw select alias (`COUNT(*) AS cnt`), which can never be a `@property`. The bail-out also discards the *key* narrowing, even though the key column **is** a declared `@property` and could be narrowed independently.

The whole call falls back to the stub type `Collection<array-key, mixed>` instead of `Collection<KeyType, mixed>`.

**Code to reproduce**

```php

/**
 * @property int $id
 * @property class-string $handler_class
 */
final class Task extends Model
{
}

function aggregate(): void
{
    $result = Task::query()
        ->selectRaw('handler_class, COUNT(*) AS cnt')
        ->groupBy('handler_class')
        ->pluck('cnt', 'handler_class');
    /** @psalm-trace $result */
    $result->isEmpty();
}
```

Actual:

```
Trace: $result: Illuminate\Support\Collection<array-key, mixed>
```

Expected:

```
Trace: $result: Illuminate\Support\Collection<class-string, mixed>
```

The value type staying `mixed` is fine (inferring a raw SQL alias type would require SQL parsing, out of scope). But the key type should still narrow via the existing `resolveKeyType()` path.

**Root cause**

`ModelPropertyResolver::resolvePluckReturnType()` resolves the value column first and returns `null` when the lookup fails:

```php
$propertyType = self::resolvePropertyType($codebase, $modelClass, $columnName);
if (!$propertyType instanceof \Psalm\Type\Union) {
    return null; // <- discards achievable key narrowing
}
```

Instead of `return null`, it could fall back to `Type::getMixed()` for the value type and continue to key resolution, returning `Collection<KeyType, mixed>`.

Verified with an instrumented handler: the event fires with the correct model template param, and the bail happens exactly at the value-column `@property` lookup (`no @property 'cnt' on Task`).

**Practical impact**

For a method documented as `@return Collection<class-string<Handler>, int<0, max>>`, the inferred `Collection<array-key, mixed>` triggers `MixedReturnTypeCoercion` on both type params. With key narrowing preserved, only the value side remains `mixed`, and a `->map(static fn (mixed $v): int => (int) $v)` can eliminate it cleanly while the key type stays typed.

**Impacted Versions**

```text
psalm/plugin-laravel                dev-master 86a83a7
vimeo/psalm                         7.0.0-beta19
laravel/framework                   v13.20.0
```

**Additional context**

Follow-up to #486 / #967 (which introduced value/key narrowing from `@property` annotations). Sibling issue for custom Builder subclasses: filed separately.

### LINKED ISSUE #1287: `pluck()` not narrowed on custom Builder subclasses (`@extends Builder<TModel>`)
**Describe the bug**

`Builder::pluck()` narrowing from `@property` annotations (#486, #967) does not fire when the call is made on a **custom Builder subclass** declared with `/** @extends Builder<TModel> */`. The result falls back to the stub type `Collection<array-key, mixed>` even when both the value and key columns are declared `@property`s on the model.

This is the standard custom-builder pattern: a model overrides `newEloquentBuilder()` and its `query()` returns the dedicated builder class.

**Code to reproduce**

```php

/**
 * @property int $id
 * @property class-string $handler_class
 */
final class Task extends Model
{
}

/** @extends Builder<Task> */
final class TaskBuilder extends Builder
{
}

function fromCustomBuilder(TaskBuilder $builder): void
{
    $result = $builder->pluck('id', 'handler_class');
    /** @psalm-trace $result */
    $result->isEmpty();
}
```

Actual:

```
Trace: $result: Illuminate\Support\Collection<array-key, mixed>
```

Expected (same as on a plain `Builder<Task>`):

```
Trace: $result: Illuminate\Support\Collection<class-string, int>
```

**Root cause**

Verified with an instrumented `BuilderPluckHandler`: the event fires (declaring class is `Illuminate\Database\Eloquent\Builder`), but

1. `$event->getTemplateTypeParameters()` is empty, because the subclass itself is not generic; and
2. the LHS fallback (`ModelPropertyResolver::extractModelFromLhsType()`) only inspects `TGenericObject` atomics, while the LHS type here is a plain `TNamedObject` (`TaskBuilder`).

So `resolvePluckReturnType()` bails with no model class resolvable.

A possible fix: when the LHS is a non-generic `TNamedObject` subclass of `Builder`, resolve `TModel` from the classlike storage's `template_extended_params[Builder::class]['TModel']` (populated by the `@extends Builder<Task>` docblock).

**Impacted Versions**

```text
psalm/plugin-laravel                dev-master 86a83a7
vimeo/psalm                         7.0.0-beta19
laravel/framework                   v13.20.0
```

**Additional context**

Sibling issue: value-column raw select alias discards key narrowing (filed separately). Real-world code frequently hits both at once, e.g. an aggregate `pluck('cnt', 'known_property')` on a custom builder.

### LINKED ISSUE #1293: `pluck()` ignores migration-schema and cast-derived column types on annotation-less models
**Describe the bug**

`pluck()` narrowing only ever consults `@property`/`@property-read` PHPDoc annotations. It never falls back to migration-schema or cast-derived column types the way `ModelPropertyHandler::getPropertyType()` (ordinary `$model->column` reads) and `BuilderAggregateHandler` (`sum()`/`min()`/`max()`/`avg()`) already do. On an annotation-less model — the common case for apps without an IDE-helper docblock generator — `pluck($column)` falls back to `mixed` for the value even when the column's type is fully known from its migration.

**Code to reproduce**

```php
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

// No @property annotations — the common case without an IDE-helper docblock.
class OauthSetting extends Model
{
    protected $fillable = ['provider'];
}

// database/migrations/..._create_oauth_settings_table.php
Schema::create('oauth_settings', function (Blueprint $table) {
    $table->id();
    $table->string('provider')->unique(); // non-nullable, uncast
    $table->timestamps();
});

function pluckColumn(): void
{
    $providers = OauthSetting::query()->pluck('provider');
    /** @psalm-trace $providers */
}

function readProperty(OauthSetting $o): void
{
    $p = $o->provider;
    /** @psalm-trace $p */
}
```

Actual:

```
Trace: $providers: Illuminate\Support\Collection<int, mixed>
Trace: $p: string
```

Expected: both should agree, since they resolve the same column on the same model —

```
Trace: $providers: Illuminate\Support\Collection<int, string>
Trace: $p: string
```

Verified live against a real app (coolify, `oauth_settings` migration, `OauthSetting` model has no `@property` docblock): the property read (`readProperty`) already resolves `$p: string` correctly today via migration-schema inference. The pluck call on the identical column, in the identical run, still returns `mixed`. Same schema data, two different consultation paths, only one of them wired to it.

**Root cause**

`ModelPropertyResolver::resolvePropertyType()` (used by both `BuilderPluckHandler` and `CollectionPluckHandler` via `resolvePluckReturnType()` and `resolveKeyType()`) is a one-liner:

```php
public static function resolvePropertyType(Codebase $codebase, string $modelClass, string $propertyName): ?Union
{
    $classStorage = $codebase->classlike_storage_provider->get($modelClass);
    return $classStorage->pseudo_property_get_types['$' . $propertyName] ?? null;
}
```

`pseudo_property_get_types` is Psalm's own storage for real `@property`/`@property-read` PHPDoc — it is never populated from migration data. Migration-inferred columns land in a completely separate registry (`ModelMetadataRegistry`, warmed by `ModelMetadataRegistryBuilder` from the plugin's own `SchemaAggregator`), consulted only through the `PropertyTypeProviderEvent` hook `ModelPropertyHandler::getPropertyType()` — which fires for `$model->property` AST nodes, not for a column name passed as a string-literal method argument like `pluck('provider')`.

The plugin already has the right tool for this: `ModelPropertyHandler::resolveColumnType(Codebase $codebase, string $fqClasslikeName, string $columnName): ?Union` is a public method that mirrors `getPropertyType()`'s full priority chain (`@property` → cast override → migration schema) for exactly this situation — a caller holding a known column-name string rather than a property-access AST node. `BuilderAggregateHandler` already depends on it for `sum()`/`min()`/`max()`/`avg()` column resolution. Pluck's handlers never adopted it; they still call the narrower `resolvePropertyType()`.

**Possible fix**

Swap the resolver call at pluck's two column-name lookups to the registry-aware one:

- Value column, in `resolvePluckReturnType()`: `self::resolvePropertyType(...)` → `ModelPropertyHandler::resolveColumnType(...)`
- Key column, in `resolveKeyType()`: same swap.

`resolveColumnType()` already checks `pseudo_property_get_types` first internally, so it's a strict superset — no `??` chaining needed, straight replacement. Bonus: `resolveKeyType()`'s existing `UnionTypeComparator::isContainedBy($codebase, $keyPropertyType, Type::getArrayKey())` guard means `pluck($value, $key)` would also start narrowing `TKey` to a migration-derived key column type (when it's a subset of `array-key`), not just the value.

Two edge cases worth locking in with test fixtures rather than assuming they just work:
- A `casts`-mapped column used as the pluck key: does the enum/cast type survive the `array-key` containment check, or fall back correctly?
- A nullable migration column: `?string` correctly fails the `array-key` containment check for key narrowing (null isn't a valid array key) — worth a regression test, since it's easy to accidentally narrow to an invalid `TKey`.

**Impacted Versions**

```text
psalm/plugin-laravel                dev-master 6909e6dd
vimeo/psalm                         7.0.0-beta19
laravel/framework                   v13.19.0
```

**Additional context**

Found while investigating a `/psalm-delta` CI comment on #1288 (`pluck()` narrowing on custom Builder subclasses and raw select alias values): coolify's `MixedArgumentTypeCoercion` count moved +1 net, and the single genuinely-new instance (`database/seeders/OauthSettingSeeder.php`, `$providers->diff($allProviders->pluck('provider'))`) traced back to exactly this gap — `pluck('provider')` on `OauthSetting`, which has no `@property` docblock. That +1 is not a #1288 regression (both value and key components #1288 infers are individually more correct than before); it's this pre-existing, orthogonal limitation surfacing because #1288 makes the pluck-without-`$key` key type more precise (`int` instead of `array-key`), which happens to stop matching the expected parameter's own `array-key` type by coincidence and exposes the value side's long-standing `mixed`. Will not be fixed as part of #1288 — that PR is narrowly scoped to `@property`-based narrowing on custom Builder subclasses and raw-alias value columns, and this bug predates it (also reproduces on master, unaffected by #1288's changes).

### LINKED ISSUE #1294: `sum()`/`avg()`/`min()`/`max()` not narrowed on custom Builder subclasses (`@extends Builder<TModel>`)
**Describe the bug**

`BuilderAggregateHandler` (narrows `sum()`/`avg()`/`average()`/`min()`/`max()` return types from the aggregated column's resolved type) has the same model-resolution gap that #1287 fixed for `pluck()`: it does not narrow when the call is made on a **custom Builder subclass** declared with `/** @extends Builder<TModel> */`. The result falls back to the stub's default return type even when the model declares a resolvable `@property`/cast/schema type for the aggregated column.

This is the standard custom-builder pattern (`#[UseEloquentBuilder]` or `newEloquentBuilder()` override) — the same pattern #1287's reproduction used.

**Code to reproduce**

```php
use Illuminate\Database\Eloquent\Attributes\UseEloquentBuilder;

/**
 * @property int $id
 * @property int $amount
 */
final class Invoice extends Model
{
}

/** @extends Builder<Invoice> */
final class InvoiceBuilder extends Builder
{
}

function aggregate(): void
{
    $result = Invoice::query()->sum('amount');
    /** @psalm-trace $result */
}
```

Actual: falls back to the stub's default `sum()` return type (not narrowed to `int`).

Expected (same as on a plain `Builder<Invoice>`): narrows to `int`, per the existing behavior for non-custom builders.

**Root cause**

`BuilderAggregateHandler::resolveModelClass()` (`src/Handlers/Eloquent/BuilderAggregateHandler.php:149-182`) resolves the model from `$event->getTemplateTypeParameters()` first, then falls back to inspecting the call's LHS type — but that fallback only inspects `TGenericObject` atomics:

```php
foreach ($lhsType->getAtomicTypes() as $atomic) {
    if (!$atomic instanceof Type\Atomic\TGenericObject) {
        continue;
    }

    $model = ModelPropertyResolver::extractModelFromUnion($atomic->type_params[0] ?? null);
    if ($model !== null) {
        return $model;
    }
}
```

For a non-generic custom builder (`final class InvoiceBuilder extends Builder<Invoice> {}`, no own `@template`), the LHS type is a plain `TNamedObject`, not `TGenericObject`, so this loop never matches — exactly the same shape #1287 diagnosed for `BuilderPluckHandler`/`ModelPropertyResolver::extractModelFromLhsType()`.

Fix: reuse `ModelPropertyResolver::extractModelFromLhsBuilderExtends()` (added in #1287), which resolves `TModel` from the classlike storage's `template_extended_params[Builder::class]['TModel']` binding, as an additional fallback in `resolveModelClass()`.

**Impacted Versions**

```text
psalm/psalm-plugin-api               0.1.0
vimeo/psalm                          7.0.0-beta19
laravel/framework                    13.19.0
```

**Additional context**

Sibling gap to #1287 (fixed for `pluck()`) and #1286 (raw-alias value column). Filed while implementing the #1287/#1286 PR after noticing the same LHS-resolution shape duplicated in `BuilderAggregateHandler`.

### LINKED ISSUE #1295: `pluck()` not narrowed on non-generic custom Collection subclasses (`@extends Collection<TKey, TModel>`)
**Describe the bug**

`CollectionPluckHandler` has the same model-resolution gap that #1287 fixed for `Builder::pluck()`, but on the `Collection` side: it does not narrow `pluck()` when called on a **non-generic custom Collection subclass** declared with `/** @extends Collection<int, TModel> */` (no own `@template`). The result falls back to the stub type `Collection<array-key, mixed>` even when the model declares a resolvable `@property` for the plucked column.

This is the standard custom-collection pattern (`#[CollectedBy]` or `newCollection()` override) — the `Collection`-side analogue of the `Builder`-side pattern #1287 fixed.

**Code to reproduce**

```php
use Illuminate\Database\Eloquent\Collection;

/**
 * @property int $id
 * @property class-string $handler_class
 */
final class Task extends Model
{
}

/** @extends Collection<int, Task> */
final class TaskCollection extends Collection
{
}

function fromCustomCollection(TaskCollection $collection): void
{
    $result = $collection->pluck('id', 'handler_class');
    /** @psalm-trace $result */
    $result->isEmpty();
}
```

Actual:

```
Trace: $result: Illuminate\Support\Collection<array-key, mixed>
```

Expected (same as on a plain `Collection<int, Task>`):

```
Trace: $result: Illuminate\Support\Collection<class-string, int>
```

**Root cause**

`ModelPropertyResolver::resolvePluckReturnType()` (shared by `BuilderPluckHandler` and `CollectionPluckHandler`) resolves the model from `$event->getTemplateTypeParameters()` first, then falls back to `extractModelFromLhsType()`, which only inspects `TGenericObject` atomics — the same shape #1287 diagnosed for `Builder`. For a non-generic custom `Collection` subclass, the LHS type is a plain `TNamedObject`, so this fallback never matches, and #1287's `extractModelFromLhsBuilderExtends()` fallback is scoped to `template_extended_params[Builder::class]`, which a `Collection` subclass's storage never has — so it correctly contributes nothing there either. No fallback currently covers this shape on the `Collection` side.

**Verified empirically** (instrumented dump of `template_extended_params` for a fixture shaped exactly like the reproduction above — `App\Collections\InvoiceCollection`, `final class InvoiceCollection extends Collection {}` with `@extends Collection<int, Invoice>`, no own `@template`):

```
template_extended_params = [
    'Illuminate\Database\Eloquent\Collection' => ['TKey' => int, 'TModel' => App\Models\Invoice],
    'Illuminate\Support\Collection'           => ['TKey' => int, 'TValue' => App\Models\Invoice],
]
```

Both ancestor entries are present and both resolve to the concrete model, flattened through the two-level `Eloquent\Collection extends Support\Collection` hierarchy exactly as Psalm's `Populator::extendTemplateParams()` does for `Builder`. A fix should consult `template_extended_params[\Illuminate\Database\Eloquent\Collection::class]['TModel']` first (matches how users actually write `@extends Collection<int, Task>`, since that `Collection` import resolves to `Illuminate\Database\Eloquent\Collection` in the standard pattern) and fall back to `template_extended_params[\Illuminate\Support\Collection::class]['TValue']` (covers a custom collection extending `Illuminate\Support\Collection` directly, without going through the Eloquent subclass), mirroring `extractModelFromLhsBuilderExtends()`'s shape but keyed on the `Collection` hierarchy instead of `Builder`.

**Impacted Versions**

```text
psalm/psalm-plugin-api               0.1.0
vimeo/psalm                          7.0.0-beta19
laravel/framework                    13.19.0
```

**Additional context**

Sibling gap to #1287 (fixed for `Builder`). Filed while implementing the #1287/#1286 PR — the `test_pluck_on_non_builder_custom_collection_unaffected` regression guard added there documents this exact gap as intentionally out of scope for that PR.

alies-dev:
## Issue to Solve

Five sibling gaps in `pluck()`/aggregate return-type narrowing that real-world code frequently hits together (e.g. an aggregate `pluck('cnt', 'known_property')` on a custom builder over an annotation-less model):

1. Narrowing from `@property` annotations does not fire when `pluck()` is called on a custom Builder subclass declared with `/** @extends Builder<TModel> */` (the standard `newEloquentBuilder()` pattern). The event's template parameters are empty because the subclass itself is not generic, and the LHS fallback only inspected `TGenericObject` atomics, so a plain `TNamedObject` LHS bailed out to the stub `Collection<array-key, mixed>`.
2. When the *value* column is not a known `@property` (typically a raw select alias like `COUNT(*) AS cnt`), the resolver bailed out entirely, discarding the *key* narrowing that was still achievable from the key column's `@property`.
3. `pluck()` ignored migration-schema and cast-derived column types entirely: on an annotation-less model whose column type the plugin's own schema inference already resolves for `$model->column` reads, `pluck('column')` stayed `mixed` (surfaced by the delta report on this PR: coolify's `OauthSettingSeeder`).
4. `sum()`/`avg()`/`min()`/`max()` had the same custom-Builder-subclass gap as (1).
5. Non-generic custom Collection subclasses (`InvoiceCollection extends Collection<int, Invoice>`) had the same gap as (1) for `Collection::pluck()`.

## Related

Fixes #1287
Fixes #1286
Fixes #1293
Fixes #1294
Fixes #1295

## Solution Description

**#1287** — `ModelPropertyResolver` gains an LHS fallback: when the LHS is a non-generic `TNamedObject`, resolve `TModel` from the classlike storage's `template_extended_params[Builder::class]['TModel']`, which Psalm's Populator populates from the `@extends Builder<Task>` docblock and flattens across deeper inheritance chains. This mirrors the established `template_extended_params[Factory::class]` pattern in `FactoryCountTypeProvider`. The fallback is naturally scoped: only classes with `Builder` as a generic ancestor have that storage key. An unsubstituted template binding (generic custom builder) safely yields no match rather than a wrong one.

**#1286** — instead of returning `null` when the value column has no resolvable type, the resolver falls back to `mixed` for the value type and continues to key resolution, returning `Collection<KeyType, mixed>`. When *both* value and key are unresolvable it still returns `null` (defers to the stub). As a side effect, single-argument `pluck('unknown_column')` now infers `Collection<int, mixed>` instead of the stub's `Collection<array-key, mixed>` — positional pluck always produces sequential int keys, so this is a strict precision gain; three pre-existing tests were updated to the more precise expectation. `resolveKeyType()` now returns `?Union` so the caller can distinguish "narrowed" from "defaulted to array-key".

**#1293** — both pluck column-resolution call sites (value and key) now use `ModelPropertyHandler::resolveColumnType()` (priority chain `@property` → casts → migration schema, already used by the aggregate handler) instead of the `@property`-only lookup. This matches runtime semantics: `Builder::pluck()` hydrates through the model when the column has a cast or accessor, so cast/schema types are the correct source. Key narrowing now also works from schema types; an enum-cast or nullable column used as key correctly falls back to `array-key` via the existing containment guard.

**#1294** — `extractModelFromLhsBuilderExtends()` is now public and wired as the final fallback in `BuilderAggregateHandler::resolveModelClass()`, reusing the #1287 mechanism for aggregates.

**#1295** — new `extractModelFromLhsCollectionExtends()` resolves the model for non-generic custom Collection subclasses from `template_extended_params` (checks `Eloquent\Collection['TModel']`, then `Support\Collection['TValue']`; both verified empirically to be present and flattened by Psalm's Populator). Naturally inert for Builder LHS the same way the Builder fallback is inert for Collections — neither hierarchy's storage carries the other's key.

Two low-value limitations are deliberately documented in code comments instead of fixed: union-LHS resolution is first-match (a union of builders over *different* models would narrow against the first resolvable atom), and argument extraction is positional (PHP named arguments would misindex). Both are vanishingly rare in real code.

Verified with type tests in `tests/Type/tests/Builder/` (`BuilderPluckCustomBuilderTest.phpt`, `BuilderPluckRawAliasTest.phpt`, `BuilderAggregateCustomBuilderTest.phpt`, `CollectionPluckCustomSubclassTest.phpt`) covering direct non-generic subclasses via the real `#[UseEloquentBuilder]` pattern, deeper inheritance chains, generic-instantiation and non-Builder regression guards, union LHS, self-referential unsubstituted templates, raw-alias value with known/unknown/no key, the combined custom-builder + raw-alias case, and `Collection::pluck()` sharing. Migration-schema and cast behavior (including enum-cast and nullable key columns) is covered by `tests/Unit/.../ModelPropertyResolverPluckSchemaTest.php` with a synthetic metadata registry, since the phpt harness does not run migrations. Full type suite, unit suite, Psalm self-analysis (`--find-dead-code --find-unused-psalm-suppress`), php-cs-fixer, and the phpt-conventions check all pass; each fix commit is independently green (bisectable).

## Checklist

### DIFF
--- a/src/Handlers/Eloquent/BuilderAggregateHandler.php
+++ b/src/Handlers/Eloquent/BuilderAggregateHandler.php
@@ -144,6 +144,13 @@ final class BuilderAggregateHandler implements MethodReturnTypeProviderInterface
      * LHS fallback handles @mixin chains where event template params arrive
      * unsubstituted (mirrors {@see ModelPropertyResolver::resolvePluckReturnType}).
      *
+     * A non-generic custom Builder subclass (`final class TaskBuilder extends
+     * Builder<Task> {}`, wired via `#[UseEloquentBuilder]`) is neither generic itself
+     * (so the template-params loop above finds nothing) nor a TGenericObject LHS (so
+     * the loop just below finds nothing either) — the same gap issue #1287 fixed for
+     * `pluck()`. Falls back to the identical `template_extended_params[Builder::class]`
+     * lookup {@see ModelPropertyResolver::extractModelFromLhsBuilderExtends()} (issue #1294).
+     *
      * @return class-string<Model>|null
      */
     private static function resolveModelClass(MethodReturnTypeProviderEvent $event): ?string
@@ -178,7 +185,7 @@ final class BuilderAggregateHandler implements MethodReturnTypeProviderInterface
             }
         }
 
-        return null;
+        return ModelPropertyResolver::extractModelFromLhsBuilderExtends($lhsType, $event->getSource()->getCodebase());
     }
 
     /**
--- a/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
+++ b/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
@@ -8,11 +8,12 @@ use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
 use Psalm\Codebase;
+use Psalm\Internal\Type\Comparator\UnionTypeComparator;
 // UnionTypeComparator is in Psalm\Internal\* but is the established convention for
 // type-containment checks in plugins (Psalm's own bundled providers use it directly,
 // and several other handlers in this codebase already depend on Psalm\Internal\*).
 // Re-verify when bumping Psalm to a new minor/major version.
-use Psalm\Internal\Type\Comparator\UnionTypeComparator;
+use Psalm\LaravelPlugin\Handlers\Eloquent\ModelPropertyHandler;
 use Psalm\NodeTypeProvider;
 use Psalm\Type;
 use Psalm\Type\Atomic\TGenericObject;
@@ -142,7 +143,11 @@ final class ModelPropertyResolver
         // columns (`selectRaw('COUNT(*) AS cnt')`) have no model annotation to resolve.
         // Don't bail on that alone — fall back to mixed for the value and still attempt
         // to narrow the key below, since the two axes are independent.
-        $resolvedPropertyType = self::resolvePropertyType($codebase, $modelClass, $columnName);
+        //
+        // resolveColumnType() (not the lower-level resolvePropertyType() below) so a
+        // column with no @property still narrows from migration schema / casts, mirroring
+        // ordinary `$model->column` reads and BuilderAggregateHandler's column resolution.
+        $resolvedPropertyType = ModelPropertyHandler::resolveColumnType($codebase, $modelClass, $columnName);
         $valueResolved = $resolvedPropertyType instanceof \Psalm\Type\Union;
         $propertyType = $resolvedPropertyType ?? Type::getMixed();
 
@@ -289,7 +294,9 @@ final class ModelPropertyResolver
             $keyColumn = $keyArg->value->value;
         }
 
-        $keyPropertyType = self::resolvePropertyType($codebase, $modelClass, $keyColumn);
+        // Registry-aware resolution (see resolvePluckReturnType() above): a key column
+        // with no @property still narrows from migration schema / casts.
+        $keyPropertyType = ModelPropertyHandler::resolveColumnType($codebase, $modelClass, $keyColumn);
         if (!$keyPropertyType instanceof Union) {
             return null;
         }
--- a/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
+++ b/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Psalm\LaravelPlugin\Handlers\Eloquent\Support;
 
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
 use Psalm\Codebase;
@@ -110,7 +111,23 @@ final class ModelPropertyResolver
         // concrete model. Fall back to inspecting the call's LHS expression, whose
         // type carries the concrete generic arguments.
         if ($modelClass === null && $lhsExpr instanceof \PhpParser\Node\Expr) {
-            $modelClass = self::extractModelFromLhsType($nodeTypeProvider->getType($lhsExpr), $modelTemplateIndex);
+            $lhsType = $nodeTypeProvider->getType($lhsExpr);
+            $modelClass = self::extractModelFromLhsType($lhsType, $modelTemplateIndex);
+
+            // Custom Builder subclasses declared as `/** @extends Builder<Task> */
+            // final class TaskBuilder extends Builder {}` are not themselves generic,
+            // so the LHS type above is a plain TNamedObject (no type_params for the
+            // fallback just above to read) and the event's template parameters are
+            // empty. Resolve TModel from the classlike storage's
+            // template_extended_params[Builder::class]['TModel'], which Psalm
+            // populates from the @extends docblock — the same mechanism
+            // ModelFactoryMethodTypeProvider/FactoryCountTypeProvider use for
+            // Factory<TModel>. Scoped to Builder: this only ever contributes a result
+            // when $modelTemplateIndex is Builder's own (0), since a Collection
+            // subclass's storage has no Builder entry in template_extended_params.
+            if ($modelClass === null) {
+                $modelClass = self::extractModelFromLhsBuilderExtends($lhsType, $codebase);
+            }
         }
 
         if ($modelClass === null) {
@@ -168,6 +185,55 @@ final class ModelPropertyResolver
 
             $modelType = $atomic->type_params[$modelTemplateIndex] ?? null;
             $modelClass = self::extractModelFromUnion($modelType);
+            if ($modelClass !== null) {
+                return $modelClass;
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * Resolve TModel for a custom, non-generic Builder subclass from its classlike
+     * storage's `@extends Builder<TModel>` binding.
+     *
+     * `ClassLikeStorageProvider::get()` lowercases internally, so the FQCN is passed
+     * through untouched. `template_extended_params` is keyed on the canonical class
+     * name (`$parent_storage->name`), which matches `Builder::class` for any
+     * vendor-stable Laravel install, and flattens through intermediate,
+     * non-generic ancestors (e.g. `TaskBuilder extends BaseTaskBuilder extends
+     * Builder`), so deeper inheritance chains resolve the same way.
+     *
+     * A generic custom builder (`@template T of Model` / `@extends Builder<T>`),
+     * used as `TaskBuilder<Task>`, is a TGenericObject and already handled by
+     * {@see self::extractModelFromLhsType()}; when reached from here instead, its
+     * own TModel binding resolves to the unsubstituted template `T`, which
+     * {@see self::extractModelFromUnion()} does not recognize as a Model, so this
+     * method correctly yields null for that case rather than a wrong binding.
+     *
+     * @return class-string<Model>|null
+     * @psalm-mutation-free
+     */
+    private static function extractModelFromLhsBuilderExtends(?Union $lhsType, Codebase $codebase): ?string
+    {
+        if (!$lhsType instanceof Union) {
+            return null;
+        }
+
+        foreach ($lhsType->getAtomicTypes() as $atomic) {
+            if (!$atomic instanceof TNamedObject) {
+                continue;
+            }
+
+            try {
+                $classStorage = $codebase->classlike_storage_provider->get($atomic->value);
+            } catch (\InvalidArgumentException) {
+                continue;
+            }
+
+            $modelClass = self::extractModelFromUnion(
+                $classStorage->template_extended_params[Builder::class]['TModel'] ?? null,
+            );
             if ($modelClass !== null) {
                 return $modelClass;
             }
--- a/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
+++ b/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
@@ -53,23 +53,6 @@ final class ModelPropertyResolver
     }
 
     /**
-     * Look up the type of a model property from @property / @property-read PHPDoc annotations.
-     *
-     * @param class-string<Model> $modelClass
-     * @psalm-mutation-free
-     */
-    public static function resolvePropertyType(Codebase $codebase, string $modelClass, string $propertyName): ?Union
-    {
-        try {
-            $classStorage = $codebase->classlike_storage_provider->get($modelClass);
-        } catch (\InvalidArgumentException) {
-            return null;
-        }
-
-        return $classStorage->pseudo_property_get_types['$' . $propertyName] ?? null;
-    }
-
-    /**
      * Build a typed Collection return type for pluck().
      *
      * Resolves the model property type from @property annotations and determines
@@ -161,9 +144,9 @@ final class ModelPropertyResolver
         // Don't bail on that alone — fall back to mixed for the value and still attempt
         // to narrow the key below, since the two axes are independent.
         //
-        // resolveColumnType() (not the lower-level resolvePropertyType() below) so a
-        // column with no @property still narrows from migration schema / casts, mirroring
-        // ordinary `$model->column` reads and BuilderAggregateHandler's column resolution.
+        // resolveColumnType() so a column with no @property still narrows from
+        // migration schema / casts, mirroring ordinary `$model->column` reads and
+        // BuilderAggregateHandler's column resolution.
         $resolvedPropertyType = ModelPropertyHandler::resolveColumnType($codebase, $modelClass, $columnName);
         $valueResolved = $resolvedPropertyType instanceof \Psalm\Type\Union;
         $propertyType = $resolvedPropertyType ?? Type::getMixed();
--- a/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
+++ b/src/Handlers/Eloquent/Support/ModelPropertyResolver.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Psalm\LaravelPlugin\Handlers\Eloquent\Support;
 
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
 use Psalm\Codebase;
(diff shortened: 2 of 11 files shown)


alies-dev:
## Before / after (release-notes material)

Given the standard custom-builder setup:

```php
/**
 * @property int $id
 * @property class-string $handler_class
 */
final class Task extends Model
{
}

/** @extends Builder<Task> */
final class TaskBuilder extends Builder
{
}
```

**`pluck()` on a custom Builder subclass** (#1287):

```php
$map = Task::query()->pluck('id', 'handler_class'); // Task::query(): TaskBuilder
// before: Collection<array-key, mixed>
// after:  Collection<class-string, int>
```

**Key narrowing survives a raw select alias value** (#1286):

```php
$counts = Task::query()
    ->selectRaw('handler_class, COUNT(*) AS cnt')
    ->groupBy('handler_class')
    ->pluck('cnt', 'handler_class');
// before: Collection<array-key, mixed>
// after:  Collection<class-string, mixed>
```

**`pluck()` now uses migration-schema and cast types** (#1293) — works on models with no `@property` annotations at all:

```php
// migration: $table->string('provider')->unique();
final class OauthSetting extends Model {} // no docblock

$providers = OauthSetting::pluck('provider');
// before: Collection<array-key, mixed>
// after:  Collection<int, string>
```

**Aggregates on a custom Builder subclass** (#1294):

```php
// @property float $total_amount on Invoice, InvoiceBuilder via #[UseEloquentBuilder]
$sum = Invoice::query()->sum('total_amount');
// before: mixed
// after:  float
```

**`pluck()` on a custom Collection subclass** (#1295):

```php
/** @extends Collection<int, Invoice> */
final class InvoiceCollection extends Collection
{
}

$numbers = Invoice::all()->pluck('invoice_number'); // Invoice::all(): InvoiceCollection
// before: Collection<array-key, mixed>
// after:  Collection<int, string>
```

Issue #1304, opened 2026-07-23
Taint Analysis: `Request::offsetGet()` (ArrayAccess) is not a taint source


alies-dev:
`Illuminate\Http\Request` implements `ArrayAccess`, and `offsetGet()` returns user input:

```php
// Illuminate\Http\Request
public function offsetGet($offset): mixed
{
    return $this->__get($offset);
}
```

It delegates straight to `__get()`, which resolves to `Arr::get($this->all(), ...)`. So `$request['term']` is the same user-controlled value as `$request->input('term')`, but carries no taint source.

**Effect.** These are the same value; only the first two are tracked:

```php
DB::table('t')->whereRaw((string) $request->input('term'));  // TaintedSql
DB::table('t')->whereRaw((string) $request->term);           // TaintedSql (since #1301)
DB::table('t')->whereRaw((string) $request['term']);         // silent
```

Verified empirically while exploring #1301: the array-access form is silent both before and after that change. `stubs/common/Http/Request.phpstub` declares no `offsetGet`.

**Why it was not folded into #1301.** #1301 was scoped to `__get` alone so its blast radius (every `$request->foo` in every codebase, across all taint kinds) could be measured on its own. `offsetGet` is a separate reachability path and deserves its own before/after measurement.

**Scope.**
- `@psalm-taint-source input` on an `offsetGet` declaration in `stubs/common/Http/Request.phpstub`.
- Confirm whether the delegation to `__get` means the source fires once or twice when both are annotated (the stub declares each method independently; Psalm sees no body).
- phpt coverage mirroring the `TaintedSqlRequestMagicProperty` / `TaintedHtmlRequestMagicProperty` pair added in #1301.
- Same sealed-subclass boundary applies: userland `extends Request` subclasses are sealed under Psalm 7's `sealAllProperties` default, but `ArrayAccess` is a real interface method rather than a magic-property fetch, so the sealing interaction differs and should be checked rather than assumed.

Note: `__isset()` was checked at the same time and needs nothing. It returns `bool`, so no data flows to a sink.

alies-dev:
not supported syntax, low ROI

alies-dev:
[...]

Issue #1300, opened 2026-07-22
TaintedSql inconsistency


Seb33300:
**Describe the bug**
The same logic behaves differently depending on how the code is written.

**Code to reproduce**
```php
// No error
$term = request('term');
Model::select('id')->whereLike('name', "%{$term}%")->get();

// ERROR: TaintedSql
$term = request('term');
Model::select('id')->where([['name', 'LIKE', "%{$term}%"]])->get();

// No error
$term = request->term;
Model::select('id')->where([['name', 'LIKE', "%{$term}%"]])->get();
```

**Impacted Versions**
Output of the
```shell
composer show | grep -E 'psalm|laravel/'
```
```text
laravel/framework                             13.20.0      The Laravel Framework.
laravel/pint                                  1.29.3       An opinionated code formatter for PHP.
laravel/prompts                               0.3.21       Add beautiful and user-friendly forms to your comm...
laravel/sanctum                               4.3.2        Laravel Sanctum provides a featherweight authentic...
laravel/serializable-closure                  2.0.13       Laravel Serializable Closure provides an easy and ...
laravel/tinker                                3.0.2        Powerful REPL for the Laravel framework.
laravel/ui                                    4.6.3        Laravel UI utilities and presets.
psalm/plugin-laravel                          4.15.1       Psalm plugin for Laravel
psalm/psalm-plugin-api                        0.1.0        Stable, semantically versioned API for Psalm plugins
vimeo/psalm                                   7.0.0-beta19 A static analysis tool for finding errors in PHP a...
```

**Psalm config**
```xml
<?xml version="1.0"?>
<psalm
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="https://getpsalm.org/schema/config"
    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
    errorLevel="8"
    findUnusedCode="false"
    ensureOverrideAttribute="false"
    runTaintAnalysis="true"
>
    <projectFiles>
        <directory name="app"/>
        <directory name="bootstrap"/>
        <directory name="config"/>
        <directory name="database"/>
        <directory name="routes"/>
        <directory name="database/factories"/>
        <!-- composer require psalm/plugin-phpunit psalm/plugin-mockery for the full tests support -->
        <!-- <directory name="tests"/>-->


        <file name="public/index.php"/>
        <file name="artisan"/>
[...]

Issue #1134, opened 2026-06-21
False positive: file_get_contents() on an uploaded file (UploadedFile) reports TaintedFile + TaintedSSRF


alies-dev:
## Summary

Reading an uploaded file with `file_get_contents($request->file('x'))` — an extremely common, safe Laravel idiom — raises two false-positive taint criticals: `TaintedFile` and `TaintedSSRF`.

## Reproduction

Found while auditing [Bottelet/DaybydayCRM](https://github.com/Bottelet/DaybydayCRM).

```php
// controller
$file = $request->file('file');                 // Illuminate\Http\UploadedFile
$fileData = $fileSystem->upload($folder, $filename, $file);

// service
public function upload($folder, $filename, $file): array
{
    $file_content = file_get_contents($file);   // <-- TaintedFile + TaintedSSRF reported here
    ...
}
```

Taint trace (abbreviated):
```
Illuminate\Http\Request::file   (stubs/common/Http/Request.phpstub:54, @psalm-taint-source input)
  -> $file = $request->file('file')
  -> upload($folder, $filename, $file)
  -> file_get_contents($file)     sink
```

## Why it's a false positive

`file()` returns an `UploadedFile`. Passing it to `file_get_contents()` coerces it via
`SplFileInfo::__toString()` to `getPathname()` — the PHP-generated temp upload path in
`upload_tmp_dir`. That path is:

- not attacker-controllable (PHP assigns a random temp name; the request cannot set it),
- never a URL (`file()` only reads the uploaded-files bag; a text field returns `null`, not a string), so SSRF via `file_get_contents` is structurally impossible here,
- not redirectable to arbitrary files (no LFI).

The attacker controls the file **content** and the **client original name**, neither of which is the path argument to `file_get_contents`. So the path-based sinks (`TaintedFile`, `TaintedSSRF`) should not fire.

## Root cause

`stubs/common/Http/Request.phpstub` marks `file()` with a coarse `@psalm-taint-source input`. That taints the whole `UploadedFile`, including its `__toString`/path coercion, so any path sink it reaches lights up. The content is user input; the temp **path** is not.

## Suggested fix

Model `UploadedFile` taint more precisely so the path coercion does not feed file/SSRF sinks while the content stays tainted. Options:

[...]

PR #1305, opened 2026-07-23
Taint Analysis: source `Request::__get()` as user input

### LINKED ISSUE #1301: Request::__get() is not a taint source
`Illuminate\Http\Request::__get()` reads user input but carries no `@psalm-taint-source`, so every dynamic property read on a request is invisible to taint analysis.

```php
public function __get($key)
{
    return Arr::get($this->all(), $key, fn () => $this->route($key));
}
```

`$this->all()` is request input. Every sibling accessor on the request is already a source in `stubs/common/Http/Request.phpstub` (`input()`, `query()`, `post()`, `get()`, `only()`, …); `__get` is the one hole.

**Effect.** These two are the same value, but only the first is tracked:

```php
DB::table('t')->whereRaw((string) $request->input('term'));  // TaintedSql
DB::table('t')->whereRaw((string) $request->term);           // silent
```

This is case C of #1300 — the reporter's third snippet was silent for this reason, not because the `where()` shape was safe.

**Why it is not folded into #1300.** Adding the source is one annotation, but its blast radius is every `request()->foo` in every user codebase, across all taint kinds (`html`, `sql`, `shell`, `ssrf`, …) at once. That needs its own psalm-delta run over the real-app corpus before merging, plus a decision on `route()`: the fallback arm returns route parameters, which are also user-controlled but are already sourced separately, so the annotation must not double-source them.

**Scope.**
- `@psalm-taint-source input` on `__get` in `stubs/common/Http/Request.phpstub` (the stub currently does not declare the method at all).
- Check `Illuminate\Http\Request::__isset` and the `ArrayAccess` `offsetGet` path for the same gap.
- phpt coverage mirroring the existing `TaintedHtmlRequestFluent.phpt` family.
- psalm-delta against the app corpus; report the per-issue-type movement in the PR.

alies-dev:
Fixes #1301.

Scope: T2 — new taint source; `__get` only. Src diff 20 stub lines + 9 comment lines; 3 new phpts.

## The gap

`Illuminate\Http\Request::__get()` returns user input:

```php
public function __get($key)
{
    return Arr::get($this->all(), $key, fn () => $this->route($key));
}
```

Every sibling accessor is already a source in `stubs/common/Http/Request.phpstub` (`input()`, `query()`, `post()`, `get()`, `only()`, `header()`, `route()`, `ip()`, `path()`, ~20 in all). `__get` was the one hole, and the stub did not declare it at all. So these two spellings of the same value analysed differently:

```php
DB::table('t')->whereRaw((string) $request->input('term'));  // TaintedSql
DB::table('t')->whereRaw((string) $request->term);           // silent
```

This is case C of #1300 — the reporter's third snippet was silent for this reason, not because the `where()` shape was safe.

## Coverage boundary

The source fires through the `__get` call node Psalm synthesises for an undeclared property fetch. That gives it a precise reach, which is worth stating because it is narrower than "every request property everywhere":

| Receiver | Covered | Why |
|---|---|---|
| `Illuminate\Http\Request` (injected, `request()`, `Request::instance()`) | **yes** | no property provider claims it; Psalm synthesises the `__get` call |
| userland `class X extends Request` | no | sealed (Psalm 7 `sealAllProperties` defaults true); the sealed path discards the `__get` return type. Already emitted `UndefinedMagicPropertyFetch` at baseline |
| `FormRequest` property, undeclared **and** with a presence-guaranteeing rule | n/a — already covered | `FormRequestPropertyHandler::doesPropertyExist()` claims the fetch before Psalm reaches `__get`; `ValidationTaintHandler` sources it with rule escapes |
| `FormRequest` property with no rule, a non-presence-guaranteeing rule (`sometimes|email`), or a real/`@property` declaration | no | not claimed by the provider (`doResolve()` refuses declared fields and requires `guaranteesPresence()`), so it falls toward `__get` and is then dropped by sealing; emits `UndefinedThisPropertyFetch` |

The two mechanisms are **disjoint by construction** — the provider either claims a property (handler sources it) or defers (sealed-drop). Confirmed by probe: a validated FormRequest property yields exactly one `TaintedHtml`, not two. No double-sourcing, and the `route()` fallback arm does not double-emit either (`$request->term` and `$request->route('term')` are two expressions, two correct findings).

Kind is bare `input`, matching all ~20 siblings.

## Interaction with #1302

#1302 (merged, `b9168f19`) strips the SQL sink element-wise from PDO-bound positions of `where()` array literals. This PR makes those positions reachable from a magic-property source for the first time, so #1300's exact third snippet is now a real interaction:

```php
$term = (string) request()->term;
Model::query()->select('id')->where([['name', 'LIKE', "%{$term}%"]])->get();  // must stay silent
```

`SafeSqlEloquentWhereNestedConditionMagicProperty.phpt` pins it. The strip keys on `spl_object_id($item)` + ordinal on the `ArrayItem`, downstream of and agnostic to the source expression shape, so no new wrong-report shape is reachable. Verified non-vacuous: moving the same tainted value to ordinal 0 (the raw-column position) still fires `TaintedSql`.

Merge order mattered here — landing this before #1302 would have converted #1300's snippet 3 from a silent false negative into an active false positive.

## Accepted imprecision

- **Sealed userland subclasses** drop the source (table above). Psalm discards the `__get` return type, and its taint nodes with it, on the sealed path. Arguably an upstream Psalm issue; out of scope here.
- **`Request::all()` merges `allFiles()`**, so `(string) $request->photo` can yield `TaintedFile`/`TaintedSSRF` on an `UploadedFile` whose only string form is a server-controlled temp path — the false-positive class `Request::file()` was carved out for in #1134. Verified pre-existing: `input()`'s and `file()`'s stub bodies are byte-identical between `origin/master` and this branch (the only delta is the `__get` addition), and on both, `$request->avatar` and `$request->input('avatar')` report identically while `$request->file('avatar')` reports neither. So the #1134 carve-out on `file()` still holds; this is a new spelling of the hole `input()` already had (baseline rule).
- **Guarded and defaulted reads are not sourced** — `$request->term ?? ''` and a read behind `isset($request->term)` stay silent, because Psalm's coalesce/isset analysis does not synthesise the `__get` call the annotation attaches to. Pre-existing: all three forms were silent before this PR, which strictly adds coverage for the bare read. Tracked as #1309.
- **Inline `validate()` escapes do not reach magic reads.** After `$request->validate(['email' => 'required|email'])`, `redirect()->to($request->email)` reports `TaintedSSRF` + `TaintedHeader` while `redirect()->to($request->input('email'))` correctly reports only `TaintedSSRF` — `ValidatedFieldReadResolver::fromPropertyFetch()` consults `FormRequest` rules only, so the inline-collected header escape never reaches the property spelling. Unlike the two above, this inconsistency **is** introduced here (the read was previously silent, so nothing could disagree). Fixing it requires the receiver-side inline-validate machinery from item 3 of #1239, which was built and deliberately dropped on complexity — out of contract for a stub-annotation PR. Tracked as #1310.

## Reviewer notes

- The stub docblock and the refreshed test comment credit `doesPropertyExist()` (with `isPropertyVisible()`) with bypassing `__get`, not `getPropertyType()` — the type provider only supplies the Union after existence/visibility have already gated the fetch. All three gate on the same `resolveRuleForProperty`.
- The new `Safe*` guard is a batch canary: it stays green in-suite even if the strip were reverted, the same disclosed limitation as its `SafeSqlWhereNestedConditionValues.phpt` sibling. The isolated control run is what actually proves it.
- This commit falsified two comments in `ValidationTaintHandler` that were true at the parent commit ("no `Request::__get` stub source exists", "or never existed"). Both refreshed here.

## Note for anyone revisiting receiver-side `validate()` narrowing

Type inference is unchanged by this PR: on a plain `Request`, `$request->filter` was `mixed` before and is `mixed` after — only the taint edge is new. Inline `$request->validate([...])` narrows its **return value** (`ValidatedTypeHandler::resolveInlineValidate()`), but receiver-side narrowing of subsequent `$request->foo` reads was item 3 of #1239 and was dropped.

If that work is ever revisited: **`__get` now belongs to the same exclusion set as `input()`**. The #1239 post-mortem established that overriding `input()`'s return type from a provider drops its stub `@psalm-taint-source` entirely, with no re-emission mechanism on the inline-validate path (the FormRequest path has one via `ValidatedFieldReadResolver::methodSourcesInput()`) — it silently killed taint on 7 existing phpts until a full-suite run caught it. As of this PR `__get` carries a stub source too, so narrowing its return type the same way would drop it the same way. `integer()`/`boolean()` remain safe to narrow, as they carry no source.

## Tests

- `TaintedSqlRequestMagicProperty.phpt` — `whereRaw((string) $request->filter)` → `TaintedSql`.
- `TaintedHtmlRequestMagicProperty.phpt` — `echo (string) $request->term` → `TaintedHtml` + `TaintedTextWithQuotes`.
- `SafeSqlEloquentWhereNestedConditionMagicProperty.phpt` — the #1302 interaction guard above.
- `SafeFormRequestSelfPropertyEmailEscapesHeader.phpt` is the seam guard proving the stub never leaks into provider-claimed fetches; green, so rule escapes survive.

Phpts use plain `\Illuminate\Http\Request` deliberately: a userland subclass receiver is silent (sealed) and would make a positive assertion born-red.

## Follow-ups

- #1309 — guarded/defaulted magic reads (`isset`, `??`) are still not sourced.
- #1310 — inline `validate()` escapes do not reach magic-property reads (the one inconsistency this PR introduces).
- #1304 — `Request::offsetGet()` (ArrayAccess) has the same gap; `$request['term']` is still silent. Kept separate so each source's blast radius is measured on its own.
- `__isset()` was checked and needs nothing: returns `bool`, no flow to a sink.

### DIFF
--- a/src/Handlers/Validation/ValidationTaintHandler.php
+++ b/src/Handlers/Validation/ValidationTaintHandler.php
@@ -24,8 +24,10 @@ use Psalm\Plugin\EventHandler\RemoveTaintsInterface;
  *
  *   1. addTaints re-introduces the source dropped by a type override
  *      ({@see ValidatedTypeHandler} / {@see FormRequestPropertyHandler}). For
- *      `$req->email` no `Request::__get` stub source exists and a provider type
- *      bypasses `__get`, so the re-source is the only thing tainting the read.
+ *      `$req->email` (undeclared, rule guarantees presence),
+ *      FormRequestPropertyHandler's `doesPropertyExist()` claims it before
+ *      Psalm ever reaches `Request::__get`, so that stub's own source can't
+ *      apply; the re-source is the only thing tainting the read.
  *   2. removeTaints applies the rule's per-field escape (e.g. `email` → safe
  *      for header/cookie).
  *
@@ -62,7 +64,10 @@ final class ValidationTaintHandler implements
 
     /**
      * Re-source the taint a validated read carries, when the stub source was
-     * dropped by a type override (or never existed, as on `Request::__get`).
+     * dropped by a type override, or never reached at all because
+     * FormRequestPropertyHandler's `doesPropertyExist()` claims an undeclared,
+     * presence-guaranteed property before Psalm gets to `Request::__get`
+     * (e.g. `$req->email`).
      */
     #[\Override]
     public static function addTaints(AddRemoveTaintsEvent $event): int
--- a/stubs/common/Http/Request.phpstub
+++ b/stubs/common/Http/Request.phpstub
@@ -252,4 +252,24 @@ class Request extends SymfonyRequest implements Arrayable, ArrayAccess
      * @psalm-taint-source input
      */
     public function getPassword() {}
+
+    /**
+     * Get an input element from the request (magic property access, e.g. `$request->term`).
+     *
+     * Fires via Psalm's synthesized __get call for undeclared property fetches on a plain
+     * Illuminate\Http\Request or request(). Userland subclasses are sealed (Psalm 7 defaults
+     * sealAllProperties to true), so this source is dropped there.
+     *
+     * An undeclared FormRequest property whose rule guarantees presence is
+     * claimed by FormRequestPropertyHandler's doesPropertyExist() (paired with
+     * isPropertyVisible()) before Psalm ever reaches __get; ValidationTaintHandler
+     * re-sources that read separately. The two paths are disjoint by
+     * construction, so this stub never double-sources a validated field.
+     *
+     * @param string $key
+     * @return mixed
+     *
+     * @psalm-taint-source input
+     */
+    public function __get($key) {}
 }
--- a/tests/Type/tests/TaintAnalysis/SafeFormRequestSelfPropertyEmailEscapesHeader.phpt
+++ b/tests/Type/tests/TaintAnalysis/SafeFormRequestSelfPropertyEmailEscapesHeader.phpt
@@ -12,9 +12,12 @@ use Illuminate\Foundation\Http\FormRequest;
  * a FormRequest narrow via FormRequestPropertyHandler, and the paired taint
  * paths in ValidationTaintHandler must:
  *
- *   1. Re-source ALL_INPUT (no stub `@psalm-taint-source` exists on `__get`,
- *      so the property handler returning a Union would otherwise leak the
- *      input pool with no taint at all).
+ *   1. Re-source ALL_INPUT. FormRequestPropertyHandler's `doesPropertyExist()`
+ *      claims the fetch before Psalm ever reaches `Request::__get`, so that
+ *      stub's own `@psalm-taint-source` (added for bare Request reads in
+ *      #1301) never applies here. Without the manual re-source, the
+ *      property handler's Union would leak the input pool with no taint
+ *      at all.
  *   2. Apply the rule's `removedTaints` mask so the same per-rule escape
  *      that fires on `$this->input('email')` also fires on `$this->email`.
  *
--- a/tests/Type/tests/TaintAnalysis/SafeSqlEloquentWhereNestedConditionMagicProperty.phpt
+++ b/tests/Type/tests/TaintAnalysis/SafeSqlEloquentWhereNestedConditionMagicProperty.phpt
@@ -1,0 +1,26 @@
+--ARGS--
+--no-progress --no-diff --config=./tests/Type/psalm.xml --taint-analysis
+--FILE--
+<?php declare(strict_types=1);
+
+class NestedConditionMagicPropertyArticle extends \Illuminate\Database\Eloquent\Model {}
+
+/**
+ * Interaction guard for #1301 x #1302, pinning #1300's exact third snippet: `request->term`
+ * (a magic-property read) as the where()-array source instead of `request('term')` /
+ * `$request->input('term')`. It reported no error at the time only because `__get` carried
+ * no taint source at all — #1301 gives it one, so the ordinal-2 PDO-bound strip added by
+ * #1302 must still apply when the source is a magic-property read, not just a method call.
+ *
+ * Same canary caveat as SafeSqlWhereNestedConditionValues.phpt: teeth-verified in an
+ * isolated psalm process — both this shape (silent) and the same chain with the tainted
+ * value moved to the ordinal-0 column position (still reports TaintedSql) — batch-swallowed
+ * in the suite.
+ */
+function safeEloquentNestedConditionMagicPropertyWhere(): void {
+    $term = (string) request()->term;
+
+    NestedConditionMagicPropertyArticle::query()->select('id')->where([['name', 'LIKE', "%{$term}%"]])->get();
+}
+?>
+--EXPECTF--
--- a/tests/Type/tests/TaintAnalysis/TaintedHtmlRequestMagicProperty.phpt
+++ b/tests/Type/tests/TaintAnalysis/TaintedHtmlRequestMagicProperty.phpt
@@ -1,0 +1,12 @@
+--ARGS--
+--no-progress --no-diff --config=./tests/Type/psalm.xml --taint-analysis
+--FILE--
+<?php declare(strict_types=1);
+
+function renderBreadcrumb(\Illuminate\Http\Request $request) {
+    echo (string) $request->term;
+}
+?>
+--EXPECTF--
+%ATaintedHtml on line %d: Detected tainted HTML
+%ATaintedTextWithQuotes on line %d: Detected tainted text with possible quotes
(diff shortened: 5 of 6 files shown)


alies-dev:
## psalm-delta: real-app corpus

Base `b9168f19` → PR arm. Measured on `eee2828d`; the shipped `8699ae55` differs only by docblock/comment wording, so the runs are analysis-identical.

**Coverage: 18/18 apps.** No app went unmeasured.

### Taint movement

| Issue type | Δ |
|---|---|
| `TaintedShell` | +6 |
| `TaintedHeader` | +2 |
| `TaintedHtml` | +1 |
| `TaintedTextWithQuotes` | +1 |
| **Total** | **+10 added, 0 removed** |

All 10 land in a single app (coolify). **Zero taint movement in the other 17.** Nothing was removed anywhere, which is the expected shape for a source-adding change: it cannot silence an existing finding.

All 10 trace to just two adjacent statements in one controller, both of the form `$model->attr = $request->attr;` on a plain type-hinted `Illuminate\Http\Request` — a request magic-property read assigned to a model attribute that later reaches shell/header/html sinks downstream. Master already tracks the identical value when spelled `$request->input('attr')`; these were invisible only because of the `__get` hole this PR closes. They look like true positives on the taint model, not noise.

### Non-taint movement: zero

Worth recording how this was established, because the first pass looked alarming. Extraction initially reported thousands of "removed" issues per app. That was an artifact, not a regression, from two parsing bugs:

1. `issues.json` carries every severity, while the captured stdout prints only `ERROR`-severity lines — so `info` issues (`ClassMustBeFinal`, `MissingPureAnnotation`, …) appeared to vanish on the arm parsed from text.
2. Message-text mismatches from the doc-link suffix domain and empty-placeholder whitespace.

With both fixed, non-taint movement is effectively zero. No evidence of any non-taint regression.

### Design-model checks

Two predictions from the design were checked directly against the findings, since either failing would invalidate the documented coverage boundary:

- **`@property`-tagged Request subclass receivers** — expected not to move (pseudo-property path, never reaches `__get`). **Zero hits.**
- **Unruled / `sometimes`-ruled FormRequest fields** — expected to stay silent (userland subclasses are sealed). **Zero hits.**

Both confirm the sealed-subclass model documented in the PR body.

### Methodology caveat

The corpus driver aborts the batch when any app exits nonzero, and Psalm exits 2 whenever it finds issues, so the PR arm never wrote `issues.json` for 9 apps. Those were recovered by parsing the captured stdout, verified byte-exact against Psalm's own `N errors found` tally per app (e.g. 22438 = 22438) to rule out truncation. Δtime is omitted: cache warmth across arms makes it meaningless here.

alies-dev:
## External review (2 rounds) + probe results

Two passes with different priming, since the priming changes what gets found. Round 1 was given the full known-issues list and asked three targeted questions; round 2 was an unprimed structured-checklist pass over the same diff. Every claim below was verified by probe before disposition — none were taken on assertion.

**Round 1 — 4 findings.**

| Finding | Verdict | Disposition |
|---|---|---|
| Comments overstate the provider boundary | Confirmed against `doResolve()` L181/L189 | **Fixed here** (`0e077b31`) |
| Whole-argument `where()` strip lacks the `isLaravelBuilder()` guard the element-wise path has | Confirmed | #1306 — false negative, predates #1302 |
| Flat/whole-map paths lack the raw-capable-object gate the nested path has | Confirmed | #1306 |
| Hoisted array literal reinstates the bound-value FP | Confirmed; #1302 documented it as a known limit | #1307 |

The first finding was the same defect class the internal gate blocked on: our comments claimed any *rule-covered* FormRequest property is claimed by the provider, when `doResolve()` also refuses declared/`@property` fields and requires `guaranteesPresence()`. Both copies and the coverage table now say "undeclared **and** presence-guaranteeing".

**Round 2 — 3 findings, all confirmed by probe.**

- `$request->term ?? ''` and reads behind `isset($request->term)` stay silent; only the bare read fires. Pre-existing — all three were silent before this PR. → #1309
- Inline `validate()` escapes never reach magic reads: `redirect()->to($request->email)` reports `TaintedSSRF` + `TaintedHeader` where `input('email')` correctly reports only `TaintedSSRF`. **This one is introduced here** and is the single inconsistency this PR adds. Fixing it needs the receiver-side machinery from item 3 of #1239, built and deliberately dropped on complexity — out of contract for a stub PR. → #1310
- UploadedFile paths: re-found independently, which corroborates the internal reviewer rather than contradicting them. Settled by diffing stub bodies between `origin/master` and this branch: `input()` and `file()` are byte-identical across the two, so master's behaviour for those methods *is* this branch's. `$request->avatar` and `$request->input('avatar')` report identically on both; `$request->file('avatar')` reports neither, so the #1134 carve-out holds. Pre-existing.

Worth recording that the two reviewers were complementary rather than redundant: the internal gate was asked whether the new source creates a wrong *report* and correctly said no; the external pass found the opposite direction — a *missing* report — in the same handler. Neither question subsumes the other.

PR #1311, opened 2026-07-23
Taint Analysis: gate whole-argument sql-taint strip on Laravel builder receiver

### LINKED ISSUE #1306: Taint Analysis: `where()` whole-argument and flat-map strip paths lack the guards the element-wise path has (false negatives)
Surfaced by an external review of #1305 and verified against source. Both are **false negatives**: SQL taint is stripped where Laravel would still emit the value raw, so a genuine injection goes unreported.

#1302 gave the element-wise strip path two guards that the other two strip paths never received.

### 1. The whole-argument strip is not scoped to a Laravel builder

`beforeExpressionAnalysis()` records the first argument for **any** receiver:

```php
self::$whereColumnArgumentIds[\spl_object_id($argument)] = true;   // WhereColumnTaintHandler.php:221
```

Element-wise removal then checks the receiver:

```php
if (!self::isLaravelBuilder($position['receiver'], $statements_source, $event)) {   // :291
    return 0;
}
```

`removeWholeArgumentTaints()` (:255-271) checks only argument-id membership and `isBoundValueMap($type)` — no receiver check. So a project's own non-builder `where()` loses the taint:

```php
final class ReportQuery
{
    public function where(array $parts): void
    {
        DB::unprepared($parts['sql']);          // injection, now unreported
    }
}

(new ReportQuery())->where(['sql' => (string) $request->term]);
```

The comment at :223-229 already spells out this exact risk ("a project's own `ReportQuery::where(array $parts)` that interpolates `$parts[0]` into raw SQL would otherwise have its list element stripped") and stores the receiver for the element-wise path only. The whole-argument path predates #1302 (the `isBoundValueMap` type-level path from the #734 / #1218 lineage), so this is not a #1302 regression — it is the older path never getting the same treatment.

Fix: store and verify the receiver for whole-argument records too.

### 2. Keyed-map values are stripped even when Laravel emits them raw

For a non-numeric string key, `scalar_gated` is false (:376), making removal unconditional (:295); the non-literal path likewise examines only keys (:523). But `addBinding()` is skipped for an `ExpressionContract`, whose `getValue()` the grammar emits verbatim. The nested path already recognises precisely this and keeps the sink (:420); the flat and whole-map paths need an equivalent raw-capable-object gate.

Reachable via `where(['name' => $expr])` where `$expr` is a user-influenced `ExpressionContract`. Narrower than item 1, but the same shape of hole.

### Scope

- Add the receiver check to the whole-argument path.
- Add a raw-capable-object gate to the flat and whole-map paths, mirroring :420.
- phpt coverage for both, using a non-builder receiver and an `ExpressionContract` value respectively. Mind the batch-swallow artifact when picking sink methods (see `docs/contributing/taint-analysis.md`).

alies-dev:
## Issue to Solve

`WhereColumnTaintHandler` records the first argument of every `where()`/`orWhere()`/`firstWhere()` call for the whole-argument sql-taint strip without looking at the receiver, while the element-wise strip verifies the receiver via `isLaravelBuilder()`. A non-builder object with a `where()` method that interpolates array values into raw SQL therefore had its argument's sql taint silently removed: a real false negative (the same shape renamed to `filter()` reports `TaintedSql`).

## Related

Fixes #1306

Scope note: #1304 (`Request::offsetGet()` taint source) was investigated in the same cycle and dropped. A stub-side `@psalm-taint-source` on `offsetGet()` cannot work: Psalm's `ArrayFetchAnalyzer` synthesizes the `offsetGet()` call in a cloned node-data set, copies only the return type, and discards the taint edge. Wrong layer for this plugin; details are on the closed issue.

## Solution Description

Contract: T1, taint fix, budget 60 src lines (actual ~27 insertions in `src/`), 1 reviewer, no corpus delta run (measured low prevalence; behaviour proven by phpts).

- `$whereColumnArgumentIds` now stores the receiver expression (`array<int, Expr|null>`) instead of `true`. The receiver still cannot be type-resolved at record time (`AfterExpressionAnalysis` fires before the strip), so it is stored and checked at strip time with the existing `isLaravelBuilder()`, mirroring the element-wise path.
- `StaticCall` stores `null` and keeps the current unguarded strip. Its receiver is always a Model class routed through the pseudo-method path, never the non-builder shape, and `SafeSqlEloquentWhereArrayValues.phpt` depends on the whole-argument strip there (the element-wise path skips `StaticCall`).
- Membership check moved from `isset()` to `array_key_exists()`: `null`-valued `StaticCall` entries must survive the lookup.
- New type test `tests/Type/tests/TaintedSqlWhereWholeArgNonBuilderReceiver.phpt` (isolated `--threads=1` group): sealed-map call result into a non-builder `where()` reaching `DB::unprepared()`, verified red on master and green at HEAD.

Reviewer notes:

- Follow-up idea, pre-existing gap not worsened here: a positive regression phpt for the real-builder whole-argument strip (unsealed call-result flags, sealed map safe). The existing `SafeSql*WhereArrayVariable*` tests are self-labeled canaries that stay green even with the handler disabled.
- Accepted imprecision: a nullsafe `?->where()` receiver typed `Builder|null` resolves `isLaravelBuilder()` to false (identical to the pre-existing element-wise behaviour); `spl_object_id` per-file-flush lifecycle is pre-existing.

## Checklist

---
<a href="https://app.blacksmith.sh/psalm/codesmith/psalm-plugin-laravel/pr/1311"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787417274&installation_model_id=424990&pr_number=1311&repository=psalm%2Fpsalm-plugin-laravel&return_to=https%3A%2F%2Fgithub.com%2Fpsalm%2Fpsalm-plugin-laravel%2Fpull%2F1311&signature=73c293e4b483835ccde10d648b1c9b6da210722684dc6c13e51714ac650efe56"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/src/Handlers/Eloquent/WhereColumnTaintHandler.php
+++ b/src/Handlers/Eloquent/WhereColumnTaintHandler.php
@@ -72,7 +72,9 @@ use Psalm\Type\Union;
  *   value dies at its own `arrayvalue-assignment` edge and never reaches the argument node, while a
  *   tainted column position still flows through. This is what makes #1300's nested form precise.
  * - **Whole-argument** ({@see isBoundValueMap}) for everything else (a variable, a call result): a
- *   type-level check that only accepts the sealed all-string-key map, where every element is a value.
+ *   type-level check that only accepts the sealed all-string-key map, where every element is a value,
+ *   gated on the receiver via {@see isLaravelBuilder} same as the element-wise path — except a
+ *   StaticCall receiver, always a Model class, which stays unguarded (#1306).
  *
  * ### Why a dynamic key records nothing
  *
@@ -145,11 +147,15 @@ final class WhereColumnTaintHandler implements
 
     /**
      * `spl_object_id` of each recorded where-family first-argument node, consumed by
-     * {@see removeTaints}. Flushed per file at {@see beforeAnalyzeFile} (NOT per function-like — the
-     * record→read gap spans a closure-bearing call's whole analysis; see the class docblock). A stale
-     * cross-file id would wrongly STRIP taint, so the file flush is load-bearing.
+     * {@see removeTaints}. The value is the call's receiver node for a `MethodCall`/
+     * `NullsafeMethodCall` — checked against {@see isLaravelBuilder} at strip time, once its type
+     * exists, since it is unresolved at record time (see {@see beforeExpressionAnalysis}) — or `null`
+     * for a `StaticCall`, whose receiver is always a Model class and so stays unguarded (#1306).
+     * Flushed per file at {@see beforeAnalyzeFile} (NOT per function-like — the record→read gap spans
+     * a closure-bearing call's whole analysis; see the class docblock). A stale cross-file id would
+     * wrongly STRIP taint, so the file flush is load-bearing.
      *
-     * @var array<int, true>
+     * @var array<int, Expr|null>
      */
     private static array $whereColumnArgumentIds = [];
 
@@ -218,7 +224,11 @@ final class WhereColumnTaintHandler implements
 
         $argument = $args[0]->value;
 
-        self::$whereColumnArgumentIds[\spl_object_id($argument)] = true;
+        // The whole-argument strip ({@see isBoundValueMap} in removeTaints) needs the same receiver
+        // gate as the element-wise one below, for the same reason: a project's own `where(array
+        // $parts)` that happens to receive a sealed string-key map is not `addArrayOfWheres()`. A
+        // StaticCall stores null and stays unguarded — see the property docblock. #1306
+        self::$whereColumnArgumentIds[\spl_object_id($argument)] = $expr instanceof StaticCall ? null : $expr->var;
 
         // Element-wise recording additionally needs the receiver, because the method NAME alone does
         // not mean Laravel: a project's own `ReportQuery::where(array $parts)` that interpolates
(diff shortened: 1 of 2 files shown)


alies-dev:
Example for release notes:

```php
final class ReportQuery
{
    /** @param array{name: string} $conditions */
    public function where(array $conditions): void
    {
        DB::unprepared('select * from reports where name = ' . $conditions['name']);
    }
}

/** @return array{name: string} */
function buildConditions(Request $request): array
{
    return ['name' => (string) $request->input('name')];
}

(new ReportQuery())->where(buildConditions($request));
```

**Before:** no issues. The handler treated any `where()` call as `Builder::addArrayOfWheres()` and stripped the SQL taint from the whole argument, so user input flowing through a non-builder `where()` into a raw SQL sink was silently ignored.

**After:**

```
ERROR: TaintedSql - Detected tainted SQL
```

The whole-argument strip now checks the receiver with `isLaravelBuilder()` (as the element-wise strip already did), so it only applies to real Eloquent/Query builders and relations. Static calls like `Model::where([...])` keep the existing behaviour.

Issue #484, opened 2026-03-19
Detect LLM prompt-injection & unsafe AI output (`laravel/ai` taint analysis)


alies-dev:
> **Status (2026-06-16): not yet shipped on `master`; scope split.**
>
> Nothing merged yet. PR #937 (draft) implements the Phase-1 `laravel/ai` taint stubs; #938 is the follow-up to move them from class-redeclaration stubs to a handler-based table. The `stubs/integrations/laravel-ai/` + `src/Handlers/Ai/` referenced in design notes are **not** on `master` (only `stubs/integrations/carbon/` exists today).
>
> **Scope split** (per @danog, psalm MEMBER — [comment](https://github.com/psalm/psalm-plugin-laravel/issues/484#issuecomment-4089321438)): the generic LLM-prompt **taint-kind primitive** (`TaintKind::INPUT_LLM_PROMPT`) belongs in **psalm core 7.x**, not this plugin. The `laravel/ai`-specific source/sink wiring (Phases 1–3 below) stays here. This issue tracks the plugin integration; the core taint-kind is an upstream ask on vimeo/psalm.

---

## Summary

With the release of [laravel/ai](https://github.com/laravel/ai), Laravel now has first-party LLM integration. This introduces two new classes of vulnerability that the plugin should detect via taint analysis:

1. **Prompt injection** — user input flowing into LLM prompts without sanitization
2. **Output injection** — LLM-generated content used unsafely in SQL, HTML, shell, or file contexts

**OWASP:** LLM01:2025 — Prompt Injection (from [OWASP Top 10 for LLM Applications](https://genai.owasp.org/llm-top-10/))

## Analysis of laravel/ai

After analyzing the full source of `laravel/ai`, here are the security-relevant findings:

### Architecture

`laravel/ai` uses a gateway pattern through `PrismGateway` (backed by `prism-php/prism`) supporting OpenAI, Anthropic, Gemini, Groq, Mistral, Ollama, Azure, DeepSeek, Cohere, and more. The core data flow:

```
User Input ($prompt string)
  → Agent::prompt() / Agent::stream()
    → Middleware pipeline (including RememberConversation)
      → Provider::prompt() → GeneratesText trait
        → PrismGateway::generateText()
          → PrismMessages::fromLaravelMessages()
            → Prism library → LLM API
              → Response objects (TextResponse, AgentResponse)
                → $response->text (raw LLM output)
```

[...]

PR #1327, opened 2026-07-26
ci(taint): add laravel/ai drift canary guard

alies-dev:
## Summary

Phase 1 drift canary for the laravel/ai taint integration (#484). The support range in `Plugin::optionalIntegrationStubs()` has no upper ceiling below 1.0, so a future laravel/ai minor can drift the signatures our stubs redeclare, and that must show up as a clear CI failure rather than silently disabling coverage.

## What changed and why extend, not add a new workflow

`tests.yml` (`type_tests_laravel_ai`) and the PHP >= 8.3 cells of `test-laravel-app.yml` already install the newest laravel/ai release on every run (no version ceiling), already run weekly (Friday 06:00 UTC cron) independent of path filters, and are already `workflow_dispatch`-able. A brand new scheduled workflow would duplicate that scheduling and CI spend for no added coverage. The actual gaps were a silent-skip blind spot and an unclear alarm, so this PR extends the two existing legs instead of adding a third workflow:

- `type_tests_laravel_ai`: every PromptInjection phpt carries a SKIPIF gated on laravel/ai class presence (M5/M6). If the install ever fails, or a SKIPIF condition itself drifts, every phpt SKIPs and `composer test:type` still exits 0. Added a guard step that parses the PHPUnit JUnit log for the PromptInjection-filtered run and fails the job if zero tests were collected or every one of them skipped, with a warning on partial skips.
- Both legs get a `failure()` step that points straight at the psalm-tester diff or compact psalm output, which already names the exact drifted class/method/line, so a red run reads as "here's the signature that broke" rather than "tests failed."
- Documented both legs and the canary rationale in `docs/contributing/taint-analysis.md`, next to the version-gate section they extend.

## Verification

- `actionlint` on both modified workflow files: clean.
- Guard logic (JUnit parsing, all-skipped / zero-collected / partial-skip branches) exercised locally against real and synthetic JUnit fixtures.
- `composer.json` / `composer.lock`: untouched by this PR.
- Diff grepped for `CLAUDE`: no hits.
- Could not trigger the actual GitHub Actions run from here; static validation only for the workflow YAML itself.

---
[...]

PR #1328, opened 2026-07-26
Add `laravel/ai` drift canary guard

alies-dev:
## Summary

Phase 1 drift canary for the laravel/ai taint integration (#484). The support range in `Plugin::optionalIntegrationStubs()` has no upper ceiling below 1.0, so coverage is never silently dropped on a new laravel/ai minor. This PR strengthens the CI legs that compensate for that decision.

## What changed and why extend, not add a new workflow

`tests.yml` (`type_tests_laravel_ai`) and the PHP >= 8.3 cells of `test-laravel-app.yml` already install the newest laravel/ai release on every run (no version ceiling), already run weekly (Friday 06:00 UTC cron) independent of path filters, and are already `workflow_dispatch`-able. A brand new scheduled workflow would duplicate that scheduling and CI spend for no added coverage. This PR extends the two existing legs instead of adding a third workflow:

- `type_tests_laravel_ai`: every PromptInjection phpt carries a SKIPIF gated on laravel/ai class presence. If the install ever fails, or a SKIPIF condition itself drifts, every phpt SKIPs and `composer test:type` would otherwise still exit 0. Added a guard (`bin/ci/assert-tests-ran.php`) that reads the JUnit log the existing test run already produces and fails the job if zero PromptInjection tests were collected or every one of them skipped, with a warning on partial skips.
- `type_tests_laravel_ai` also runs `bin/ci/check-laravel-ai-stub-parity.php`: it parses every `stubs/integrations/laravel-ai/*.phpstub` with php-parser and diffs native parameter/return types against the installed vendor package via reflection. See the next section for why this exists.
- Both legs get a `failure()` step pointing at the psalm-tester diff or compact psalm output for the actual failure.
- Documented both legs and their scope in `docs/contributing/taint-analysis.md`, next to the version-gate section they extend.

## What these legs catch, and what they do not

They catch: laravel/ai failing to install (JUnit skip-guard above), a stubbed class being renamed/moved/removed (the same SKIPIF/all-skipped guard), the plugin's own regressions (the normal test suite), and, via `check-laravel-ai-stub-parity.php`, a stubbed method's native parameter/return type drifting on a laravel/ai release.

That last one needed its own check because nothing else can see it: for a method a stub redeclares, Psalm resolves the call against the stub's signature, never the vendor's, so a registered stub is never diffed against the real class it redeclares. Confirmed empirically during review: mutating an installed laravel/ai method's native signature while leaving its stub untouched left the PromptInjection suite green (12/12 passing, 0 skipped). Filed **#1331** with the reproduction. `check-laravel-ai-stub-parity.php` closes that gap and is itself verified against the same synthetic mutation.

A `KNOWN_GAPS` allowlist in that script covers pre-existing mismatches so the job fails only on genuinely new drift, not on gaps already tracked elsewhere. It currently allowlists one (`Contracts\CanActAsTool::description()`'s widened return type, fixed in #1330). Building the checker also surfaced a second, more serious gap not in the allowlist: `Responses\StreamableAgentResponse::__toString()` is declared in the stub as a taint source, but the installed class does not define that method at all, so `(string) $response` on that type is a fatal Error at runtime that the stub told Psalm was fine. That one is deliberately NOT allowlisted (also being fixed in #1330), so **this job is expected to be red until #1330 merges**. An allowlisted entry that stops reproducing prints a loud warning to remove it, rather than silently staying green or failing on a gap that no longer exists.

## Verification

- `actionlint` on both modified workflow files: clean.
- Guard logic (JUnit parsing, all-skipped / zero-collected / partial-skip branches): exercised locally against a real run and against synthetic all-skipped / zero-collected JUnit fixtures.
- Reproduced the signature-drift gap directly, confirmed `check-laravel-ai-stub-parity.php` catches it, restored the mutated vendor file afterward.
- Confirmed the stale-`KNOWN_GAPS` warning fires: simulated the `CanActAsTool` fix in a scratch copy of the stubs (not the real files) and got the "no longer reproduces" warning instead of a silent pass or a false failure.
- `composer test:type -- --no-progress`: 647 tests, 1 pre-existing unrelated skip, 0 failures.
- `composer cs`: clean. `composer psalm`: no errors, 100% type coverage. (`bin/ci/` is outside `psalm.xml`'s `projectFiles`, consistent with the other existing `bin/ci/*.php` scripts, so this self-analysis doesn't cover the new scripts.)
- `composer.json` / `composer.lock`: untouched by this PR.
- Diff grepped for `CLAUDE`: no hits.
- Not exercised: could not trigger `workflow_dispatch` on GitHub from here, so neither leg's actual Actions run was exercised end to end. Static validation only for the workflow YAML itself.

Related: #1331 (the reproduction and design rationale for the parity check), #1330 (fixes both `KNOWN_GAPS`-relevant stubs).


alies-dev:
Consolidating into #1327 (branch 484-phase1-gaps is now the canonical branch for this work per team-lead).

PR #1348, opened 2026-08-15
Fix: Recognize non-HTML response content

### LINKED ISSUE #1345: Taint Analysis: `TaintedHtml` on `response()->make()` when literal `$headers` declare a non-HTML `Content-Type`
**Describe the bug**

The `html` sink on `ResponseFactory::make()` is right by default: `make($content)` with no headers serves `text/html`. It is wrong when the same call passes a literal `$headers` array declaring a non-HTML `Content-Type`, or `Content-Disposition: attachment`. Every CSV, XML, iCal, and image export written in that form reports `TaintedHtml`.

This is not a "runtime information we cannot see" case. In the reported shape the headers are a literal array in the same call expression, statically visible at the sink.

Expected: no `TaintedHtml`. The response is never parsed as HTML by a browser. `Content-Disposition: attachment` settles it on its own (the response is downloaded, not rendered); the declared `text/csv` is the second, independent reason.

Actual: one `TaintedHtml` per tainted value reaching `$content`.

**Code to reproduce**

```php
public function __invoke(Request $request): Response
{
    $output = "Name\n" . $request->query('name') . "\n";

    return response()->make($output, 200, [
        'Content-Type' => 'text/csv',
        'Content-Disposition' => 'attachment; filename="members.csv"',
    ]);
    // TaintedHtml: Detected tainted HTML
}
```

The shape that surfaced this was a team-members CSV export concatenating member names and emails into `$output`, one `TaintedHtml` per tainted column. The `$request` source above is the same flow reduced to a documented source so the repro stands on a fresh app.

**Impacted Versions**

```text
laravel/framework       v13.24.0
psalm/plugin-laravel    dev-master (11131bd2c645)
vimeo/psalm             7.0.0-beta19
```

**Psalm config**

Default taint behaviour for Psalm 7 (`runTaintAnalysis` on, no taint-specific overrides); `errorLevel="1"`, `TaintedHtml` not suppressed.

**Additional context**

### Where the sink is declared

- `stubs/common/Routing/ResponseFactory.phpstub` — `@psalm-taint-sink html $content` on `make()`. Predates #1318.
- `stubs/common/Contracts/Routing/ResponseFactory.phpstub:23` — the mirrored contract annotation, added in #1318.

Zero-argument `response()` is typed as the contract, so the contract stub is the one that fires. Any change touches both files plus `ContractStubSinkParityTest`, which enforces the mirror in both directions.

### Precedent

#1318 dropped the `html` sink from `json()` for the same chain, and the concrete stub records the reasoning:

> `$data` is intentionally NOT an html sink: it is JSON-encoded and served as `application/json`, which browsers do not render as HTML. Reaching XSS from here needs legacy content-sniffing AND the app dropping nosniff, too weak a chain to justify a default-on sink.

`make()` with a literal `text/csv` plus `attachment` is the same chain and strictly weaker: it needs legacy content-sniffing, the app dropping `nosniff`, *and* the browser ignoring an explicit attachment disposition. If that chain does not justify a default-on sink for `json()`, it does not justify one here.

### Severity

Psalm 7 runs taint by default, so this now fires on every default-config run rather than only under `--taint-analysis`. On one Laravel 13 codebase I count a dozen-odd `response()->make()` call sites serving non-HTML content: four CSV exports (all `text/csv` + `attachment`), XML sitemaps, an iCal feed, and a set of certificate image responses. None are HTML, and each becomes a report as soon as its content carries user data.

### Options

1. **Strip `html` from `$content` when a literal sibling `$headers` declares a non-HTML response.** Gate on the receiver resolving to `ResponseFactory` (or its contract) and the third argument being an array literal whose `Content-Type` is a known non-HTML type or whose `Content-Disposition` begins with `attachment`. Absent, non-literal, or unrecognised headers keep the sink, so the default stays safe and every degradation lands on the status quo.

   `WhereColumnTaintHandler` is the working precedent for the mechanism and most of the cost is already paid there. `AddRemoveTaintsEvent` carries only the argument expression — no method id, no argument offset (`ArgumentAnalyzer::processTaintedness()`, `ArgumentAnalyzer.php:1759`) — so that handler pairs `BeforeExpressionAnalysisInterface`, which sees the whole call and records the `spl_object_id` of the argument node it wants stripped, with `removeTaints()`, which consumes the recorded id. The same shim applies here, with the same retirement path if `AddRemoveTaintsEvent` ever grows the method id and offset upstream.

   It also settles the house-rule question by example: that handler gates on the receiver (`isLaravelBuilder`) *and* on argument shape (a keyed-map literal, with `isScalarValued` guarding the case where the AST cannot see the runtime value). So "gate on the receiver, not the argument shape" is not absolute; it rules out gating on argument shape *alone*. This proposal keeps the receiver gate and adds a literal sibling argument as a second condition, which is the same combination.

2. **Do nothing and document it.** Record the gap with the existing `*KnownLimitation.phpt` plus caveat-docblock pattern, and add a line to `docs/security.md` telling users to suppress at the call site for non-HTML responses. Cheapest, but it pushes a suppression into every export controller, and the repo's own guidance is to fix rather than suppress.

3. **Drop the `make()` html sink entirely.** Rejected: `make($userInput)` with no headers really does serve `text/html`, which is the common and genuinely dangerous form.

Option 1. The precedent for both the reasoning (#1318's `json()` drop) and the mechanism (`WhereColumnTaintHandler`) already exists in the repo; option 2 is the fallback if the shim's cost is judged too high for the size of the false-positive class.

### Not in scope

CSV exports built this way are exposed to formula injection: a member name beginning with `=`, `+`, `-`, or `@` is evaluated when the file opens in Excel or Sheets. That is a real finding, but it is a different vulnerability class than the one an `html` sink models, and modelling it would need its own taint kind and its own escape (a leading-quote or apostrophe prefix). Mentioned only so that "the warning did catch something" does not get read as an argument for keeping the `html` sink here.

alies-dev:
## Issue to Solve

`ResponseFactory::make()` reports `TaintedHtml` for a response body that is explicitly served as a downloaded attachment.

## Related

Fixes #1345

Supersedes this PR's own earlier `FuncCall`/`StaticCall` recording gate. Upstream cause of that gate: vimeo/psalm#11924. Sibling issue #1415 is the intended second consumer of the mechanism landed here (not implemented in this PR).

## Solution Description

T3 — attachment-only response taint refinement; delta: yes.

Keeps the default HTML sink. The exemption is now decided at **issue-emission time** instead of by mutating the taint graph.

### Why the mechanism changed

The first implementation was a two-phase bridge: `BeforeExpressionAnalysisInterface` recorded the content AST node in a `WeakMap`, `RemoveTaintsInterface` answered `INPUT_HTML` for that node. `AddRemoveTaintsEvent` carries no dispatch context, so when the content is itself a call, the same node receives the event again during its own return-type fetch and the removal lands on the callee's **shared unspecialized `@psalm-flow` edge**, silencing unrelated flows project-wide (vimeo/psalm#11924). The defence was a syntactic gate that refused to record `FuncCall`/`StaticCall` content, which left a deliberate false positive on `make(decrypt(...), 200, [attachment])`.

### What ships instead

`ResponseFactoryTaintHandler` now implements only `BeforeAddIssueInterface` and never touches the taint graph. Per emitted issue, cheapest bail first:

1. `$issue instanceof TaintedHtml` — O(1), and the only work done on the ~600 non-matching dispatches of a full self-analysis.
2. The `$journey` tail label must be one of the three verified response-factory routes (concrete class, contract, `Response` facade).
3. That tail's `getSelectionBounds()` is exactly the content argument's span.
4. `Codebase::getStatementsForFile()` yields the sink file's AST; the `make()` call whose argument 0 occupies that span is located by offset, never by receiver shape (`parent::`, `$this->`, `static::` all reach this sink).
5. The literal proof re-runs on that node: three positional arguments, no unpack, no named argument, inline header array, literal string pairs, exactly one `Content-Disposition`, CR/LF/NUL guard on the raw value before the attachment match.
6. All satisfied → suppress. Anything unknown, unparsable, ambiguous, or mismatched → `null`, and the finding stands.

Properties this buys: sound by construction (the worst failure is a retained finding, never a lost one); no false-positive floor, so `make(decrypt(...), 200, [attachment])` is exempt too; stateless, since taint resolution runs in the main process after the worker pool and nothing a worker recorded still exists — which also means no static state and nothing to register in `Plugin::resetInvocationState()`.

### Behaviour changes versus the previous mechanism

- Content produced by a function or static call is now exempt (the recovered false positive).
- Both receiver gates are deleted. Receiver identity comes from Psalm's own resolution via the journey label, which is a stronger proof than an AST-side type check. Consequence: a receiver whose type **intersects** the factory or contract is now exempt as well, in both intersection orders. Pinned in `SafeResponseFactoryMakeLiteralNonHtmlHeaders.phpt` and documented under the conforming-receiver trust statement in `docs/security.md`.
- The root `\Response` alias is deliberately **not** in the label set and still reports.
- Non-matching paths return `null` rather than `true`, so a later plugin's `BeforeAddIssue` handler still runs.

## Accepted imprecision

- Content-type-only, dynamic, ambiguous, list-valued, underscore-named, malformed, and non-attachment headers retain `TaintedHtml`.
- Header names are folded the way `Symfony\Component\HttpFoundation\HeaderBag` folds them (`_` to `-`, case-insensitive) to decide which entries *count* as the disposition, but the canonical dashed spelling is still required to *prove* one. So `['Content-Disposition' => 'attachment', 'Content_Disposition' => 'inline']` is retained (Symfony's later `set()` wins, making the effective disposition `inline`), and an underscore-only spelling stays reported exactly as before rather than becoming a new exemption.
- The raw header value is rejected on any C0 control or DEL (horizontal tab excepted as legitimate OWS) before trimming. Unknown parameters after the token, `filename*=UTF-8''x.csv`, and `filename="a;b"` stay exempt — browsers treat all three as downloads.
- **Node-ID collision (not fixed, upstream).** `DataFlowNode` identifies a call-argument node by label plus *lower-cased* file name plus byte span, and equal IDs overwrite. On a case-sensitive filesystem, `src/Foo.php` and `src/foo.php` holding a factory call at an identical span collide, so a dangerous flow could inherit the exempt call's location and be suppressed. It needs a pathological repo layout, and the defect is Psalm's node ID rather than this handler.
- Positional arguments only. Named arguments and variable-held header arrays are explicit non-goals; the proof helper takes the whole call node so both extend naturally later.
- Conforming-receiver trust, as already stated in `docs/security.md`.
- **Shared-sink pruning (new, pinned by `TaintedHtmlResponseFactoryMakeSharedSinkKnownLimitation.phpt`).** An exempted flow stays in the graph and marks the project-wide `ResponseFactory::make#1` node visited for the html mask, so `TaintFlowGraph`'s `visited_source_ids` pruning can discard a *longer* dangerous flow into the same node. It requires the exempt call to be the shorter path; the reverse order still reports. There is **no regression against `master`**, which loses that flow identically — the previous mechanism only avoided it by way of the project-wide leak this PR removes. Same upstream family as vimeo/psalm#11924 and #1415.

## Verification

- Teeth check: registration commented out → the safe fixture reports all 10 exempt call sites, none swallowed, reproduced independently in the real 348-file taint batch as well as standalone.
- KnownLimitation non-vacuity proved by mutation in both directions: deleting the shallow exempt call makes the deep chain report; giving it non-attachment headers makes it report while the deep chain stays silent.
- Deleting the load-bearing named-argument check fails exactly the three new provider cases.
- Deleting the header-name folding fails the new Unproven phpt case; deleting the load-bearing named-argument check fails exactly the three new provider cases.
- `test:unit` 1,167 tests / 3,156 assertions; `test:type` 678 tests; `test:app`; PHPT conventions lint; `psalm` with `--find-dead-code --find-unused-psalm-suppress` and without `--no-suggestions`, 0 issues; rector and php-cs-fixer clean.
- Reviews: internal gate 95/100 with no must-fix, plus an adversarial soundness pass that produced the shared-sink limitation above. An external Codex (xhigh) round then found the two false negatives closed in the last commit; its third finding is the node-ID collision recorded above as upstream.
- Corpus delta versus `master`: inconclusive locally. Only one benchmark app's toolchain wiring was usable, and it showed no issue change, no coverage change, and no significant time change. The teeth for this PR are in the PHPTs, not in the corpus — the recovered shape (`make(fn(...), attachment)`) is rare in real applications.

## Reviewer notes (not blocking, no change made)

- PHPT `--ARGS--` batch isolation rests on each group's argument string being unique; nothing guards a future test from reusing one and re-sharing the batch.
- `%d`-only EXPECTF (the CI convention) pins counts and kind order but not line identity, so a suppress-here/report-there swap inside a multi-finding fixture would stay green.

## Backport

Not in this PR. `BeforeAddIssueInterface` is present on Psalm 6 with the same signature and the same pre-suppression dispatch point, and this design uses no taint-kind masks, so the Psalm 6/7 `list<string>` vs int divergence does not apply.

## Checklist
- [x] Documentation is updated (if needed, otherwise remove this item)

### DIFF
--- a/docs/contributing/README.md
+++ b/docs/contributing/README.md
@@ -207,8 +207,13 @@ Every class registration keeps the matching `require_once` beside
 `registerHooksFromClass()`. Call site taint handlers such as
 `ResponseFactoryTaintHandler` pair `BeforeExpressionAnalysisInterface` with
 `RemoveTaintsInterface` because Psalm's removal event has neither a method id nor
-an argument offset. Any node id recorded for that bridge is cleared before each
-file, so it cannot survive into a later AST.
+an argument offset. Key that bridge on the node object (a `WeakMap`), never on its
+`spl_object_id`. Psalm frees whole ASTs other than the analysed file's while that
+file is still being analysed (`ProjectAnalyzer::getMethodMutations()` on the
+constructor initialisation path, `ClassLikes::getTraitNode()`), and dispatches no
+`BeforeFileAnalysisEvent` for them, so a per file flush does not bound an id's
+lifetime. PHP then reissues the freed handle to an unrelated node, and a stale hit
+STRIPS taint, which is a missed finding rather than a false positive.
 
 See [Architecture Decisions](decisions.md) for design rationale, [Laravel Magic Call Patterns](laravel-magic-call-patterns.md) for how Laravel's __call/__callStatic chains work, [Psalm Type Annotations](types.md) for a quick reference of all supported types and annotations, and [Debugging with Xdebug](xdebug.md) for stepping through handler code.
 
--- a/docs/contributing/README.md
+++ b/docs/contributing/README.md
@@ -207,16 +207,29 @@ There are two ways to register:
 2. **Closure-level** (model property handlers): register via `$providers->property_type_provider->registerClosure(...)` — used by `ModelRegistrationHandler` to bind property handlers per-model after codebase is populated
 
 Every class registration keeps the matching `require_once` beside
-`registerHooksFromClass()`. Call site taint handlers such as
-`ResponseFactoryTaintHandler` pair `BeforeExpressionAnalysisInterface` with
-`RemoveTaintsInterface` because Psalm's removal event has neither a method id nor
-an argument offset. Key that bridge on the node object (a `WeakMap`), never on its
-`spl_object_id`. Psalm frees whole ASTs other than the analysed file's while that
-file is still being analysed (`ProjectAnalyzer::getMethodMutations()` on the
-constructor initialisation path, `ClassLikes::getTraitNode()`), and dispatches no
-`BeforeFileAnalysisEvent` for them, so a per file flush does not bound an id's
-lifetime. PHP then reissues the freed handle to an unrelated node, and a stale hit
-STRIPS taint, which is a missed finding rather than a false positive.
+`registerHooksFromClass()`.
+
+#### Exempting one call site from a stub's taint sink
+
+Narrow the finding at emission time, not in the taint graph.
+`ResponseFactoryTaintHandler` is the reference shape: it implements only
+`BeforeAddIssueInterface`, matches the issue's journey tail against the sink's
+labels, re-checks the call site, and returns `false` to drop the finding. Graph
+level removal (`RemoveTaintsInterface`) cannot express "this call site only":
+`AddRemoveTaintsEvent` carries neither a method id nor an argument offset, so the
+removal is keyed on an AST node, and Psalm dispatches the event for that node a
+second time while fetching its own callee's return type. The removal then lands on
+a shared, project wide `@psalm-flow` edge and silences unrelated flows. See
+[Architecture Decisions](decisions.md) for the full dead end record.
+
+An emission time handler must be stateless. Taint findings are resolved in the
+MAIN process after the worker pool exits (`Analyzer::analyzeFiles()`), while type
+issues are emitted inside the workers, so nothing a per expression or per file
+hook recorded survives to `beforeAddIssue()`. Re-derive every fact from the issue
+plus `Codebase::getStatementsForFile()`, which reads the parser cache. The hook
+also fires for every issue in the run, so bail on the issue class first and only
+then do anything expensive. Every uncertain path returns `null` and keeps the
+finding.
 
 See [Architecture Decisions](decisions.md) for design rationale, [Laravel Magic Call Patterns](laravel-magic-call-patterns.md) for how Laravel's __call/__callStatic chains work, [Psalm Type Annotations](types.md) for a quick reference of all supported types and annotations, and [Debugging with Xdebug](xdebug.md) for stepping through handler code.
 
--- a/docs/contributing/README.md
+++ b/docs/contributing/README.md
@@ -221,21 +221,16 @@ Narrow the finding at emission time, not in the taint graph.
 `ResponseFactoryTaintHandler` is the reference shape: it implements only
 `BeforeAddIssueInterface`, matches the issue's journey tail against the sink's
 labels, re-checks the call site, and returns `false` to drop the finding. Graph
-level removal (`RemoveTaintsInterface`) cannot express "this call site only":
-`AddRemoveTaintsEvent` carries neither a method id nor an argument offset, so the
-removal is keyed on an AST node, and Psalm dispatches the event for that node a
-second time while fetching its own callee's return type. The removal then lands on
-a shared, project wide `@psalm-flow` edge and silences unrelated flows. See
-[Architecture Decisions](decisions.md) for the full dead end record.
+level removal (`RemoveTaintsInterface`) cannot express "this call site only" and
+leaks onto shared flow edges; [Architecture Decisions](decisions.md) records why,
+along with the alternatives already ruled out.
 
-An emission time handler must be stateless. Taint findings are resolved in the
-MAIN process after the worker pool exits (`Analyzer::analyzeFiles()`), while type
-issues are emitted inside the workers, so nothing a per expression or per file
-hook recorded survives to `beforeAddIssue()`. Re-derive every fact from the issue
-plus `Codebase::getStatementsForFile()`, which reads the parser cache. The hook
-also fires for every issue in the run, so bail on the issue class first and only
-then do anything expensive. Every uncertain path returns `null` and keeps the
-finding.
+Such a handler has to be stateless, because the analysis hooks run in the worker
+processes and taint issues are emitted only after those exit. Re-derive every fact
+from the issue plus `Codebase::getStatementsForFile()`, which reads the parser
+cache. The hook fires for every issue in the run, so bail on the issue class first
+and only then do anything expensive. Every uncertain path returns `null` and keeps
+the finding.
 
 See [Architecture Decisions](decisions.md) for design rationale, [Laravel Magic Call Patterns](laravel-magic-call-patterns.md) for how Laravel's __call/__callStatic chains work, [Psalm Type Annotations](types.md) for a quick reference of all supported types and annotations, and [Debugging with Xdebug](xdebug.md) for stepping through handler code.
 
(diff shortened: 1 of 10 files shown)


alies-dev:
Benchmarked this branch (`f727f9e9`) against the real-world app corpus and it produces **zero delta** versus `master` (`7e7f7d77`): the taint finding lists are identical, app for app. The PR body says "delta: yes", but on this corpus there is none.

Cause: no corpus app writes the inline-literal header form the handler requires.

```php
|| !$args[2]->value instanceof Array_
|| !self::provesAttachment($args[2]->value)
```

The two shapes that actually occur both bail out here:

1. **Variable holding the literal**, assigned a few lines above the call:

```php
$headers = [
    'Content-Type' => 'text/csv',
    'Content-Disposition' => "attachment; filename=\"report.csv\"",
];

return response()->make($output, 200, $headers);
```

2. **Helper method returning the array** — BookStack's `DownloadResponseFactory::directly()`:

```php
return response()->make($content, 200, $this->getHeaders($fileName, strlen($content)));
```

where `getHeaders()` defaults `$mime` to `application/octet-stream` and therefore emits `Content-Disposition: attachment`.

Measured: BookStack reports 12 taint findings on both refs, including the two `TaintedHtml` on `app/Http/DownloadResponseFactory.php:21` this PR is meant to silence. A second app (private, not in the public list) reports 31 on both, including 4 `TaintedHtml` on CSV-export endpoints using shape 1.

Context: these findings are new in v4.15.4 via #1318, which added the `html` sink to the `Illuminate\Contracts\Routing\ResponseFactory` stub. `response()` returns the contract, so the sink became reachable for the first time; the concrete-class sink predates it and rarely fired. Across the corpus, 6 of the 15 new taint findings in v4.15.4 are this exact false positive, and all 6 survive this PR.

Suggestion for a follow-up (or for extending this PR): handle shape 1 by resolving the argument to its literal when the variable has a single unambiguous assignment in the same scope. That stays purely syntactic, needs no type information, and covers 4 of the 6. Shape 2 needs the header-building method's literal returns, which is a larger step with more false-negative risk.

Not a blocker for the change as scoped, just worth correcting the delta claim before merge.

PR #1349, opened 2026-08-16
Fix: guard instance where taint stripping

### LINKED ISSUE #1337: Instance-path where() array strip trusts inheritance: a Builder subclass overriding where() with raw SQL is silenced
- `WhereColumnTaintHandler` matches the instance-call receiver by inheritance against `BUILDER_CLASSES`, so a userland `class ReportQuery extends Builder` that declares its OWN `where()` (raw SQL concatenation) still gets the #734 value strip: a real sink is silenced (false negative).
- Probed during #1336 review: the shape reproduces today; pre-existing behavior, unchanged by the #1336 fix.
- The static path already solved this in #1300: `isStaticReceiverLaravelBound()` rejects a receiver whose called method has a concrete declaration outside `Illuminate\`. The instance path can reuse the same `getDeclaringMethodId()` gate once the receiver type is resolved.
- Low prevalence: requires a builder subclass that both overrides `where()` and interpolates raw SQL in it. Such a class should also carry its own `@psalm-taint-sink`, which would keep flagging regardless.
- Family: #734, #1300, #1306, #1336.

alies-dev:
## Issue to Solve

A userland `Query\Builder` subclass that overrides `where()` inherited Laravel's taint value strip, which could suppress a real `TaintedSql` report from its own raw SQL construction.

## Related

Fixes #1337

## Solution Description

T2 — dispatch-gated taint-suppression fix; 97 changed source lines.

- Records the called method name for every queued instance `where` argument position.
- Rejects the strip when the concrete declaration belongs outside `Illuminate\`.
- Preserves native Laravel behavior and leaves unresolved, widened, or delegated userland receivers tainted.
- Adds a taint PHPT for a concrete `Query\Builder::where()` override and confirms the native control remains safe.
- Accepted imprecision: `FN-over-FP`; uncertain dispatch retains taint.
- Reviewer gate: 96/100, SHIP. No must-fix findings.
- Verification: targeted taint pin and controls, type suite, unit suite, Psalm CI invocation with `--no-cache`, php-cs-fixer dry run, and PHPT conventions passed.

## Checklist

### DIFF
--- a/src/Handlers/Eloquent/WhereColumnTaintHandler.php
+++ b/src/Handlers/Eloquent/WhereColumnTaintHandler.php
@@ -180,7 +180,7 @@ final class WhereColumnTaintHandler implements
      * the class docblock). A stale cross-file id would wrongly STRIP taint, so the file flush is
      * load-bearing.
      *
-     * @var array<int, Expr|null>
+     * @var array<int, array{method: lowercase-string, receiver: Expr}|null>
      */
     private static array $whereColumnArgumentIds = [];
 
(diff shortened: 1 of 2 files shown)


alies-dev:
Closing because the rare userland Builder override case does not justify the additional dispatch guard complexity.

PR #1382, opened 2026-08-20
Support `$schema = Schema::connection(...)` in migrations

### LINKED ISSUE #1377: Migration schema aggregator misses a schema builder assigned to a variable
The migration schema aggregator does not recognise a schema builder that is first assigned to a local variable, so every column of such a migration is invisible to `ModelMetadataRegistry`. Models backed by those tables lose their column properties, which surfaces as false `UndefinedMagicPropertyFetch` / `UnknownModelAttribute` reports.

Laravel Telescope's published migration uses exactly this shape:

```php
public function up(): void
{
    $schema = Schema::connection($this->getConnection());

    $schema->create('telescope_entries', function (Blueprint $table): void {
        $table->bigIncrements('sequence');
        $table->uuid('uuid');
        $table->longText('content');
        // ...
    });
}
```

`SchemaAggregator::extractSchemaCall()` handles only two forms:

- direct: `Schema::create('users', ...)`
- connection-chained: `Schema::connection('mysql')->create('users', ...)`

The `Assign` statement is skipped, and the later `$schema->create(...)` is a `MethodCall` on a `Variable`, which matches neither branch.

**Prevalence.** 5 of the 19 apps in the local benchmark corpus (vito, coolify, pixelfed, solidtime, plus the Telescope-published migration itself) carry this shape, for both `create()` and `dropIfExists()`.

**Scope.**

- Track a local variable assigned from `<Schema facade>::connection(...)`, per method, and route the existing `create` / `table` / `drop` / `dropIfExists` / `dropColumns` / `rename` switch through it.
- Untrack the variable when it is reassigned to anything else (including `foreach` rebinds), so a non-builder variable never attributes columns to a table.
- `Illuminate\Database\Schema\Builder` has no `setConnection()` in the supported Laravel versions, so only `connection()` is tracked.

Upstream reference: larastan/larastan#1884 (open) proposes the same idea for Larastan, without the untracking.

alies-dev:
## Issue to Solve

The migration schema aggregator saw only two call forms:

- direct: `Schema::create('users', ...)`
- connection-chained: `Schema::connection('mysql')->create('users', ...)`

A migration that assigns the builder to a variable first matched neither, so every column it declared was invisible to `ModelMetadataRegistry`. Models backed by those tables lost their column properties, surfacing as false `UndefinedMagicPropertyFetch` / `UnknownModelAttribute` reports.

Laravel Telescope's published migration is exactly this shape:

```php
$schema = Schema::connection($this->getConnection());

$schema->create('telescope_entries', function (Blueprint $table): void {
    $table->bigIncrements('sequence');
    $table->longText('content');
    // ...
});
```

5 of the 19 apps in the local benchmark corpus carry it, for both `create()` and `dropIfExists()`.

## Related

Fixes #1377.

Upstream reference: larastan/larastan#1884 proposes the same idea for Larastan (still open). This PR diverges from it in two ways, both deliberate:

- it untracks the variable on rebinding (larastan's version only ever tracks);
- it omits `setConnection()`, which `Illuminate\Database\Schema\Builder` does not have in any supported Laravel version.

## Solution Description

`SchemaAggregator::addMethodStatements()` keeps a local `array<string, true>` of variable names holding a schema builder.

- A name is tracked only on `<SchemaClass>::connection(...)`, gated by the existing `isSchemaClass()` (so custom Schema facades and root aliases keep working).
- The name is untracked on any other `Assign`, on `AssignRef` / `AssignOp`, and on a `foreach` value var.
- `extractSchemaCall()` gained a third branch: a `MethodCall` on a tracked variable. The existing `create` / `table` / `drop` / `dropIfExists` / `dropColumns` / `rename` switch and the four per-method handlers are unchanged.

The tracking set is local to the method walk rather than a property, because one aggregator instance spans every migration file and `addClassStatements()` walks every non-`down()` method. Keeping it local stops a name from leaking between helper methods.

Narrow gating matters here: `$schema = Schema::make(...)` appears 170 times in the benchmark corpus and is Filament's unrelated `Filament\Schemas\Schema`. The `connection()` name check and `isSchemaClass()` each exclude it independently.

Verification: 12 unit tests in `tests/Unit/Handlers/Eloquent/Schema/SchemaVariableTest.php` (4 positive, 8 negative). Every accept branch and every untrack branch is pinned individually: each clause was deleted in turn, the corresponding test confirmed red, and the file restored byte-identical (md5 checked both ways). Full unit suite, type suite, and CI-exact Psalm self-analysis (no `--no-suggestions`, `--no-cache`) are green.

## Reviewer notes

- `MigrationCache` payload shape is unchanged (still `SchemaAggregator::$tables`), so no cache-format bump is needed.
- Accepted imprecision, per the FN-over-FP and cheap-heuristic-first conventions: builders reached by a route the statement walker cannot see stay untracked, and are therefore silently ignored rather than mis-attributed. That covers `Schema::getFacadeRoot()`, chained assignment (`$a = $b = Schema::connection(...)`), property-held and DI-injected builders, and a parameter named `$schema`.
- Shapes that stay tracked through a rebinding the walker cannot follow (`[$a, $schema] = f()`, `list()`, `foreach ($m as $schema => $v)`, `catch (X $schema)`, by-ref `use (&$schema)`) would require a real `Schema::connection()` assignment followed by a rebind followed by a `create()` on the rebound name, which fatals at runtime. Zero corpus prevalence, no baseline delta.
- Optional follow-up, not taken here to keep the diff inside budget: `extractSchemaCall()`'s connection-chained branch could reuse `isSchemaConnectionCall()` and drop ~5 duplicated lines.

## Checklist
- [x] Tests cover the change (unit tests in `tests/Unit/Handlers/Eloquent/Schema/SchemaVariableTest.php`)

### DIFF
--- a/src/Handlers/Eloquent/Schema/SchemaAggregator.php
+++ b/src/Handlers/Eloquent/Schema/SchemaAggregator.php
@@ -117,14 +117,32 @@ final class SchemaAggregator
      */
     private function addMethodStatements(array $stmts): void
     {
+        // Names of local variables currently holding a Schema builder instance
+        // (e.g. `$schema = Schema::connection(...)`), scoped to this method only.
+        /** @var array<string, true> $schema_vars */
+        $schema_vars = [];
+
         // Flatten nested block structures (if/else, try/catch, foreach, etc.)
         // so Schema calls inside conditionals are not missed.
         foreach (self::flattenStatements($stmts) as $stmt) {
+            // A foreach loop variable rebinds the name each iteration
+            // (e.g. `foreach ($tables as $schema)`), so any tracked builder
+            // under that name no longer holds a schema builder afterward.
+            if ($stmt instanceof PhpParser\Node\Stmt\Foreach_) {
+                if ($stmt->valueVar instanceof PhpParser\Node\Expr\Variable && \is_string($stmt->valueVar->name)) {
+                    unset($schema_vars[$stmt->valueVar->name]);
+                }
+
+                continue;
+            }
+
             if (!$stmt instanceof PhpParser\Node\Stmt\Expression) {
                 continue;
             }
 
-            $schema_call = $this->extractSchemaCall($stmt->expr);
+            $this->trackSchemaVariable($stmt->expr, $schema_vars);
+
+            $schema_call = $this->extractSchemaCall($stmt->expr, $schema_vars);
             if ($schema_call === null || !$schema_call->name instanceof PhpParser\Node\Identifier) {
                 continue;
             }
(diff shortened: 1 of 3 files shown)


alies-dev:
Reopening to re-run CI: the style auto-fix commit was pushed by Actions, which does not trigger workflow runs.

PR #1449, opened 2026-09-02
Fix: custom Eloquent builder methods returning self flagged as unused return value

### LINKED ISSUE #1448: Custom Eloquent builder methods returning self are reported as PossiblyUnusedReturnValue when call sites discard the return
**Describe the bug**

A method on a custom Eloquent builder that returns `self`/`static` is reported as `PossiblyUnusedReturnValue` whenever every call site discards the return value.

Discarding it is correct Laravel. `Illuminate\Database\Eloquent\Builder` is mutable: `where()`, `whereNull()` and friends mutate `$this->query` and return `$this` only as a convenience. So `$query->published();` and `$query = $query->published();` are equivalent, and the statement form is the idiomatic one when a builder is handed to you to modify in place — a `Closure` filter callback, a policy scope, a `Scope` object, a shared query-modifier helper.

Psalm has an exemption for this shape (`$method_storage->probably_fluent`, checked in `Psalm\Internal\Codebase\ClassLikes::checkMethodReferences()`), but its detection is purely syntactic (`Psalm\Internal\PhpVisitor\Reflector\FunctionLikeNodeScanner`):

```php
$last_stmt = end($stmt->stmts);

if ($last_stmt instanceof PhpParser\Node\Stmt\Return_
    && $last_stmt->expr instanceof PhpParser\Node\Expr\Variable
    && $last_stmt->expr->name === 'this'
) {
    $storage->probably_fluent = true;
}
```

It requires a literal `return $this;`. A builder method is almost never written that way — it is written `return $this->whereNull(...);`, which is the same thing through Eloquent's own fluent API. The declared return type is not consulted at all, so `self` vs `static` makes no difference.

Psalm cannot fix this generically: for an immutable fluent API, discarding the return really is a bug, and Psalm has no way to know which is which. This plugin does know. It already owns custom-builder semantics (`CustomBuilderMethodHandler`, registered per builder class by `ModelRegistrationHandler`), so it can tell that a method declared on an `Illuminate\Database\Eloquent\Builder` subclass and returning `self`/`static`/the builder type is mutating-fluent by construction.

The failure is also unstable in a way that makes it read as noise rather than a finding. On a real codebase, a builder with ten such methods reported exactly two: the two whose *only* call site was a statement. The eight others had at least one chained call site somewhere (`->published()->get()`), which registers a return reference and clears the issue for every other call site too. Same code shape, same correctness, opposite verdict, decided by an unrelated file.

**Code to reproduce**

```php
// app/PsalmRepro/Post.php
namespace App\PsalmRepro;

use Illuminate\Database\Query\Builder as QueryBuilder;

final class Post extends Model
{
    /** @param QueryBuilder $query */
    #[\Override]
    public function newEloquentBuilder($query): PostBuilder
    {
        return new PostBuilder($query);
    }
}
```

```php
// app/PsalmRepro/PostBuilder.php
namespace App\PsalmRepro;

/** @extends Builder<Post> */
final class PostBuilder extends Builder
{
    public function published(): self
    {
        return $this->whereNotNull('published_at');
    }

    public function draft(): self
    {
        return $this->whereNull('published_at');
    }
}
```

```php
// app/PsalmRepro/PostFilter.php
namespace App\PsalmRepro;

final class PostFilter
{
    public function apply(PostBuilder $query, bool $onlyPublished): void
    {
        if ($onlyPublished) {
            $query->published();
        } else {
            $query->draft();
        }
    }
}
```

```php
// app/PsalmRepro/registration.php
// Class references only; without one Psalm reports UnusedClass and never inspects members.
return [
    \App\PsalmRepro\Post::class,
    \App\PsalmRepro\PostBuilder::class,
    \App\PsalmRepro\PostFilter::class,
];
```

Full-project run, `findUnusedCode="true"`:

```text
PossiblyUnusedReturnValue  app/PsalmRepro/PostBuilder.php:12  The return value for this method is never used
PossiblyUnusedReturnValue  app/PsalmRepro/PostBuilder.php:17  The return value for this method is never used
```

Expected: no findings. Add `->get()` after either call and that method's finding disappears while the other remains, which shows the verdict is driven by an unrelated call site rather than by the method.

**Impacted Versions**

```text
laravel/framework           v13.29.0
psalm/plugin-laravel        dev-master
psalm/psalm-plugin-api      0.1.0
vimeo/psalm                 7.0.0-beta19
```

PHP 8.5.

**Psalm config**

```xml
<?xml version="1.0"?>
<psalm
        errorLevel="1"
        resolveFromConfigFile="true"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns="https://getpsalm.org/schema/config"
        xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
        findUnusedCode="true"
>
    <projectFiles>
        <directory name="app"/>
        <ignoreFiles>
            <directory name="vendor"/>
        </ignoreFiles>
    </projectFiles>
    <issueHandlers>
        <UnusedClass errorLevel="info"/>
        <PossiblyUnusedMethod errorLevel="info"/>
    </issueHandlers>
    <plugins>
            <experimental value="true"/>
            <failOnInternalError value="true"/>
        </pluginClass>
    </plugins>
</psalm>
```

**Additional context**

Possible fix: in an `AfterClassLikeVisit` handler, for a class extending `Illuminate\Database\Eloquent\Builder`, set `$methodStorage->probably_fluent = true` on public methods whose declared return type is `self`, `static`, `$this` or the builder class itself. `checkMethodReferences()` reads that flag directly (`$method_storage->is_static || !$method_storage->probably_fluent`), so nothing else needs to change, and methods returning a model, a collection, a count or `void` keep reporting normally.

Scope question worth deciding: whether the same treatment should extend to legacy `scopeXxx()` and `#[Scope]` methods that return a builder, and to `Illuminate\Database\Query\Builder` subclasses. Both are the same mutable-fluent contract, though scopes are reached through `__call` and may already be covered by other reference edges.

A narrower alternative is to only exempt methods whose body ends in `return $this->...;` on the builder, which is closer to Psalm's existing heuristic, but it would still miss a method ending in a conditional return and does not add much over the return-type check.

alies-dev:
## Issue to Solve

A method declared on a custom Eloquent builder (`extends Illuminate\Database\Eloquent\Builder`) that returns `self`/`static`/its own class name gets flagged `PossiblyUnusedReturnValue` whenever every call site discards the return, e.g. `$query->published();`. That's correct Laravel: the builder is mutable, `where()`/`whereNull()` and friends return `$this` only as a convenience, so the statement form is idiomatic when a builder is handed to you to modify in place (a filter callback, a policy scope, a shared query-modifier helper).

Psalm's own `probably_fluent` detection is purely syntactic — it only recognizes a literal `return $this;` body. A builder method almost always reads `return $this->whereNotNull(...);` instead, so it never trips the exemption, and the declared return type is never consulted. The failure is also order-dependent: on a real codebase a builder with ten such methods reported findings only for the two whose sole call site was a statement — any other call site chaining the same method (`->published()->get()`) cleared the issue everywhere.

## Related

Fixes #1448

## Solution Description

`BuilderFluentReturnHandler` (new, `AfterCodebasePopulatedInterface`) sets `MethodStorage::$probably_fluent = true` for public, non-static, declared methods on an `Illuminate\Database\Eloquent\Builder` subclass whose declared return type is the builder. Psalm's gate reads that flag directly (`ClassLikes.php`: `$method_storage->is_static || !$method_storage->probably_fluent`), so nothing else changes.

Return-type rule, which is asymmetric on purpose:

- **Exempted:** `self`, `static` (native and docblock-only — they parse to different `TNamedObject` shapes and both need matching), `$this`, the builder's own class name, and intersections where **any** member is the builder (`self&Contract`, `Contract&self`; `TypeParser::getTypeFromIntersectionTree()` demotes every member after the first into `extra_types`).
- **Declined:** a union unless **every** arm is the builder. `probably_fluent` suppresses the unused-return check for the whole method, so exempting `self|Collection` or `self|null` on the strength of one arm would hide a genuinely discarded result. Intersection members all describe the same returned object; union arms are different possible objects.

Methods returning a model, a collection, a count, or `void` keep reporting normally. Static methods are skipped because `is_static` short-circuits Psalm's gate, so setting the flag on one cannot change any outcome.

Scoped narrowly to the issue's repro — legacy `scopeXxx()`/`#[Scope]` methods and `Illuminate\Database\Query\Builder` subclasses are intentionally out of scope (same mutable-fluent contract, no confirmed FP evidence yet).

## Verification

`findUnusedCode` needs a whole-project run, so the regression is a subprocess-fixture PHPUnit test (a phpt only analyzes the file it is passed), matching the existing `IndirectMethodReferencesTest` pattern.

The fixture needs a reachable entry point (`app/entry.php`): without one, `PostQueries` is `UnusedClass`, every builder method is `PossiblyUnusedMethod`, and Psalm never reaches the unused-return-value check — the fixture then produces **zero** findings and the regression passes vacuously.

Findings with the handler enabled: exactly two, `PostBuilder.php:67` (the non-fluent `Collection` control) and `:78` (the `self|Collection` union control). Note `PossiblyUnusedReturnValue` is reported at the method's `return_type_location`, which for a docblock-only `@return` is the annotation line, not the declaration line.

RED proofs, both against the final fixture:

- handler unregistered -> 7 findings; the six fluent methods false-positive at lines 15, 27, 39, 46, 52, 59. This is #1448.
- union rule reverted to "any arm matches" -> `maybeCollection()` wrongly exempted, 1 finding, test fails on the union control.

Gates: `composer test:unit` (1272 tests), `composer test:type` (758 tests), `composer psalm` (no errors, 100% inferred coverage, no baseline and no suppressions), rector + cs.

## Checklist
- [x] Tests cover the change (unit test in `tests/Unit/`)
- [x] Documentation is updated (`docs/contributing/README.md` handler-registration diagram)

### DIFF
--- a/docs/contributing/README.md
+++ b/docs/contributing/README.md
@@ -34,7 +34,7 @@ flowchart TD
     "]
 
     I["Psalm scans all project files"] -.->|afterCodebasePopulated| J["ModelRegistrationHandler"]
-    I -.->|afterCodebasePopulated| K["Eloquent Builder subclass fix-ups:\nBuilderSubclassQueryMixinHandler (restores dropped Query Builder @mixin)\nBuilderNativeStaticReturnTypeHandler (native ': static' return becomes docblock 'static')"]
+    I -.->|afterCodebasePopulated| K["Eloquent Builder subclass fix-ups:\nBuilderSubclassQueryMixinHandler (restores dropped Query Builder @mixin)\nBuilderNativeStaticReturnTypeHandler (native ': static' return becomes docblock 'static')\nBuilderFluentReturnHandler (sets probably_fluent on self/static-returning builder methods, #1448)"]
     I -.->|afterCodebasePopulated| FMB["FactoryModelBindingHandler (injects @extends Factory&lt;TModel&gt; on bare factory subclasses, #780)"]
     I -.->|afterCodebasePopulated| FSP["FacadeStubPrecedenceHandler (drops conflicting facade @method pseudos when the plugin ships a real stubbed static method)"]
     I -.->|afterCodebasePopulated| FTF["FacadeTaintForwardingHandler (copies taint sinks from a facade's forwarding target onto its @method pseudo-methods)"]
--- a/src/Handlers/Eloquent/BuilderFluentReturnHandler.php
+++ b/src/Handlers/Eloquent/BuilderFluentReturnHandler.php
@@ -38,8 +38,11 @@ final class BuilderFluentReturnHandler implements AfterCodebasePopulatedInterfac
             $ownNameLower = \strtolower($storage->name);
 
             foreach ($storage->methods as $method_storage) {
-                // is_static is already exempt in ClassLikes::checkMethodReferences(), and an
-                // already-fluent method (Psalm's own `return $this;` detection) needs no help.
+                // ClassLikes::checkMethodReferences() gates on `is_static || !probably_fluent`:
+                // a static method is ALWAYS subject to the unused-return check regardless of
+                // probably_fluent (is_static short-circuits the OR to true), so setting the flag
+                // on one would be a pure no-op — skip them. An already-fluent method (Psalm's own
+                // `return $this;` detection) needs no help either.
                 if ($method_storage->is_static || $method_storage->probably_fluent) {
                     continue;
                 }
@@ -58,21 +61,41 @@ final class BuilderFluentReturnHandler implements AfterCodebasePopulatedInterfac
                         continue;
                     }
 
-                    // is_static covers native `: static` (reflected with the declaring class's
-                    // own FQN and is_static=true); the literal value check covers `self`,
-                    // docblock-only `@return static` (parses to value='static', is_static=false),
-                    // and the builder's own class name spelled out explicitly.
-                    $value_lower = \strtolower($atomic->value);
-                    if ($atomic->is_static
-                        || $value_lower === 'self'
-                        || $value_lower === 'static'
-                        || $value_lower === $ownNameLower
-                    ) {
+                    // An intersection type (`Foo&Bar`) stores only its first-listed member as
+                    // the top-level atomic; the rest live in that atomic's own `extra_types`
+                    // (TypeParser::getTypeFromIntersectionTree()), so `Contract&self` needs the
+                    // same match applied there too — the builder type can be either side.
+                    if (self::matchesFluentReturn($atomic, $ownNameLower)) {
                         $method_storage->probably_fluent = true;
                         break;
+                    }
+
+                    foreach ($atomic->extra_types as $extra_type) {
+                        if ($extra_type instanceof TNamedObject && self::matchesFluentReturn($extra_type, $ownNameLower)) {
+                            $method_storage->probably_fluent = true;
+                            break 2;
+                        }
                     }
                 }
             }
         }
+    }
+
+    /**
+     * is_static covers native `: static` (reflected with the declaring class's own FQN and
+     * is_static=true); the literal value check covers `self`, docblock-only `@return static`
+     * (parses to value='static', is_static=false), and the builder's own class name spelled out
+     * explicitly.
+     *
+     * @psalm-pure
+     */
+    private static function matchesFluentReturn(TNamedObject $atomic, string $ownNameLower): bool
+    {
+        $value_lower = \strtolower($atomic->value);
+
+        return $atomic->is_static
+            || $value_lower === 'self'
+            || $value_lower === 'static'
+            || $value_lower === $ownNameLower;
     }
 }
--- a/src/Handlers/Eloquent/BuilderFluentReturnHandler.php
+++ b/src/Handlers/Eloquent/BuilderFluentReturnHandler.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psalm\LaravelPlugin\Handlers\Eloquent;
+
+use Illuminate\Database\Eloquent\Builder;
+use Psalm\Internal\Analyzer\ClassLikeAnalyzer;
+use Psalm\Plugin\EventHandler\AfterCodebasePopulatedInterface;
+use Psalm\Plugin\EventHandler\Event\AfterCodebasePopulatedEvent;
+use Psalm\Type\Atomic\TNamedObject;
+
+/**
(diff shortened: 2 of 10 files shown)


alies-dev:
Addressed the review. Two of the four points were real; two were not, and I verified each against a real whole-project run (Psalm `7.0.0-beta21`, `laravel/framework` pinned below 13.32 as CI does).

**Real, fixed**

- **Fixture reachability (the actual cause of red CI).** `PostQueries` was referenced from nowhere, so with `findUnusedCode` Psalm reported `UnusedClass` for it and `PossiblyUnusedMethod` for every builder method, and produced **zero** unused-return findings. The regression was vacuous: all seven `assertNull`s passed for the wrong reason and the control assertion read `null`. Added `app/entry.php` as a reachable reference into `PostQueries::run()`.
- **Union arms.** Matching a single arm exempted `self|Collection` and `self|null`, hiding a genuinely discarded result. Unions now require **every** arm to be the builder; intersections still match on any member, because every member describes the same returned object. Added a `self|Collection` control that must stay reportable.

**Not reproducible, unchanged**

- **Control line 67.** `PossiblyUnusedReturnValue` is reported at the method's **`return_type_location`**, not the declaration line — for a docblock-only `@return` that is the `@return` line. Verified: the finding lands on `PostBuilder.php:67`, the `@return Collection<int, Post>` line, not 69.
- **Static `forGuest()` should be reported.** It is not. `ClassLikes.php:2097` does keep statics eligible (`is_static || !probably_fluent`), but no static method in this shape produces a discarded-return finding at all — a probe static returning `Collection` with its return discarded produced none either. So it stays in the not-reported set; I corrected its docblock, which previously claimed statics are "always checked". It pins that the handler leaves statics alone, nothing more.

**Verification**

Fixture findings, handler enabled: exactly two, `PostBuilder.php:67` (`Collection` control) and `:78` (union control).

RED proofs, both with the new fixture:

- handler unregistered -> 7 findings; the six fluent methods false-positive at lines 15, 27, 39, 46, 52, 59. This is #1448.
- previous handler restored -> `maybeCollection()` wrongly exempted, 1 finding, test fails on the union control.

Gates: `composer test:unit` 1272 tests green, `composer test:type` 758 tests green, `composer psalm` no errors at 100% inferred coverage, rector + cs clean.

PR #1457, opened 2026-09-15
Drop the Psalm `7.0.0-beta21` workarounds

alies-dev:
## Issue to Solve

#1456 shipped the `DataFlowNode::getForTaint()` rename that Psalm `7.0.0-beta21` requires, and carried a set of temporary guards so CI could run at all on that Psalm. Both guards exist only because of upstream defects. This PR removes them.

- vimeo/psalm#11958: an unparseable `@self-out` / `@this-out` type escapes as an uncaught `TypeParseTreeException` from the scan phase, before any issue is reported. `laravel/framework` v13.32.0 ships the first docblock of that shape (`Collection::prepend()`), so every job resolving it analysed nothing at all. Reproduces on Psalm 6.17.2 too, and on a project that never mentions Laravel, because composer's autoload `files` pull `Collection.php` into the scan.
- vimeo/psalm#11959: one of two taint flows reaching the same sink method from two different files is dropped when both are analysed together. Thirteen fixtures lose a finding that way (eleven taint, plus two PromptInjection once laravel/ai shifts the batch), and one reports its finding twice. Each still passes analysed on its own.

## Related

Depends on vimeo/psalm#11958 and vimeo/psalm#11959. Follows #1456.

## Solution Description

Straight removal, nothing replaced:

- the `laravel/framework` upper bound in `psalm.yml`, the `tests.yml` type-test matrix (back to `^13.3`), the `tests.yml` unit lockfile cells, and both laravel/ai legs
- the `LARAVEL_FRAMEWORK_CONSTRAINT` hook in `laravel-test.sh` and its use in `test-laravel-app.yml`
- `PsalmVersion::skipOnRange()` and the thirteen fixture gates built on it

One thing is deliberately kept: the explicit `return` in the two PromptInjection `--SKIPIF--` guards. It stops a second skip reason being printed when laravel/ai is absent and has nothing to do with either upstream defect.

Verification that the removal is complete rather than approximate: diffed against `ece3ad4c`, the commit preceding the workarounds. The only differences this branch carries over that tree are #1456's actual fix (`composer.json`, `docs/upgrade-v4.md`, `TimingUnsafeComparisonHandler`) and those two `return` lines.

## Merge preconditions

Draft on purpose. **CI on this branch is expected to be red until upstream ships both fixes** — that red is the signal, not a problem to work around.

Before merging:

1. vimeo/psalm#11958 fixed and released.
2. vimeo/psalm#11959 fixed and released.
3. A Psalm carrying both is what `composer.json`'s `^7.0.0-beta21 || dev-master` actually resolves to. If either fix lands behind a constraint bump, raise the floor in the same PR.
4. Re-run `composer test:type`, `composer test:unit`, `composer psalm` and the fresh-app leg green, with `laravel/framework` left unpinned.

If only one of the two is fixed, split this PR rather than merging half of it.

## Checklist
- [x] Tests cover the change (restores thirteen fixtures to unconditional execution, which is the test coverage in question)
- [x] Documentation is updated (the guards carried their own inline rationale and `@see`; all of it goes with them)


alies-dev:
Superseded by #1458 (vimeo/psalm#11958, the laravel/framework 13.32 pins) and #1459 (vimeo/psalm#11959, the taint fixture gates). Split so that whichever upstream fix lands first can be merged on its own.

